PR #13, opened 2019-09-30
Emphasize whole line deletions and insertions

nkouevda:
For paired edits, `minus_color` and `plus_color` denote unchanged parts
of the line, while the `emph_*` colors denote changes.

For unpaired edits, since the whole line is a change, it should be
colored with `emph_*`.

### DIFF
--- a/src/edits.rs
+++ b/src/edits.rs
@@ -47,7 +47,7 @@ where
 
                 // Emit as unpaired the plus lines already considered and rejected
                 for plus_line in &plus_lines[emitted..(emitted + considered)] {
-                    annotated_plus_lines.push(vec![(noop_insertion, plus_line)]);
+                    annotated_plus_lines.push(vec![(insertion, plus_line)]);
                 }
                 emitted += considered;
                 annotated_minus_lines.push(annotated_minus_line);
@@ -61,11 +61,11 @@ where
             }
         }
         // No homolog was found for minus i; emit as unpaired.
-        annotated_minus_lines.push(vec![(noop_deletion, minus_line)]);
+        annotated_minus_lines.push(vec![(deletion, minus_line)]);
     }
     // Emit any remaining plus lines
     for plus_line in &plus_lines[emitted..] {
-        annotated_plus_lines.push(vec![(noop_insertion, plus_line)]);
+        annotated_plus_lines.push(vec![(insertion, plus_line)]);
     }
 
     (annotated_minus_lines, annotated_plus_lines)
@@ -366,7 +366,7 @@ mod tests {
             vec!["á á b á á c á á b"],
             (
                 vec![
-                    vec![(MinusNoop, "á a a á a a á a a")],
+                    vec![(Deletion, "á a a á a a á a a")],
                     vec![
                         (MinusNoop, "á á b á á "),
                         (Deletion, "b"),
@@ -390,13 +390,13 @@ mod tests {
             vec!["bbbb ! bbb", "dddd d ddd", "cccc ! ccc"],
             (
                 vec![
-                    vec![(MinusNoop, "aaaa a aaa")],
+                    vec![(Deletion, "aaaa a aaa")],
                     vec![(MinusNoop, "bbbb "), (Deletion, "b"), (MinusNoop, " bbb")],
                     vec![(MinusNoop, "cccc "), (Deletion, "c"), (MinusNoop, " ccc")],
                 ],
                 vec![
                     vec![(PlusNoop, "bbbb "), (Insertion, "!"), (PlusNoop, " bbb")],
-                    vec![(PlusNoop, "dddd d ddd")],
+                    vec![(Insertion, "dddd d ddd")],
                     vec![(PlusNoop, "cccc "), (Insertion, "!"), (PlusNoop, " ccc")],
                 ],
             ),
@@ -601,8 +601,8 @@ mod tests {
     // Assert that no edits are inferred for the supplied minus and plus lines.
     fn assert_no_edits(minus_lines: Vec<&str>, plus_lines: Vec<&str>, max_line_distance: f64) {
         let expected_edits = (
-            minus_lines.iter().map(|s| vec![(MinusNoop, *s)]).collect(),
-            plus_lines.iter().map(|s| vec![(PlusNoop, *s)]).collect(),
+            minus_lines.iter().map(|s| vec![(Deletion, *s)]).collect(),
+            plus_lines.iter().map(|s| vec![(Insertion, *s)]).collect(),
         );
         assert_edits(minus_lines, plus_lines, expected_edits, max_line_distance)
     }


dandavison:
Hi, thanks for this proposal! I've been thinking about it while dealing with other open issues. The main thing that's on my mind is -- this makes delta behave differently from both the Github and GitLab UIs, whereas currently it behaves the same as them. Now, I agree that there is a logical argument for what you're proposing. I.e. one could view it as establishing a consistent mapping:

- emph color : added or removed content
- non-emph color: unchanged content in an affected line

Whereas with the GitHub / GitLab scheme it's perhaps not really consistent:

- emph color:  added or removed content in a line with unchanged content
- non-emph color: either unchanged content in a changed line, or changed content in a line which lacks any unchanged content.

However, an alternate view could be that the color semantics under GitHub/GitLab are

- non-emph color: this line has been changed
- emph color: here's a bit of extra help in understanding the change

I'm not saying that it's a premise of Delta that it must replicate GitHub/GitLab UIs, but departing from them is something I want to do only with some thought. (I wonder whether this issue has been discussed in the course of GitLab development).

I suspect that what I'm going to end up concluding is that this makes sense as a command line non-default option, but that for now I'd like delta to remain consistent with the popular GitHub/GitLab diff views by default.

As an example, let's look at https://github.com/davidhalter/jedi/commit/258d5aee4a8733b8aedcc12e6acafa1104b2796e.

(There's a secondary question: what was your intention for the "two-tone" effect on unpaired lines? Should the leading and trailing whitespace be emph or non-emph? Currently it seems that leading is emph and trailing is non-emph.)

#### Here's a comparison of delta on your branch, versus GitHub and GitLab.

<table>
  <tr>
    <td>
      delta (proposed) vs. github
    </td>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
  <tr>
    <td>
      delta (proposed) vs. gitlab
    </td>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
</table>

#### In contrast, here's how that comparison looks with delta currently:

 
<table>
  <tr>
    <td>
      delta (current) vs. github
    </td>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
  <tr>
    <td>
      delta (current) vs. gitlab
    </td>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
</table>

nkouevda:
Thank you for the detailed response; this made me realize a particular strange aspect of how GitHub/GitLab highlight, and why I find the current delta behavior confusing:

In your screenshots, GitHub/GitLab have emph highlighting in the first hunk (+1/-1) but not in the second hunk (+5/-1), whereas [current] delta has emph highlighting in both. That's the issue I'm trying to address, and where delta currently differs from GitHub/GitLab: for hunks where there's a mix of changed and inserted/deleted lines, delta still does emph for the changed lines, making such "mixed" hunks more difficult to understand.

So to clarify your summary of GitHub/GitLab behavior, I believe this is accurate:

- non-emph color: this line has been changed
- emph color: here's a bit of extra help in understanding the change **(only if there are no whole-line deletions or insertions in this hunk)**

+1 to having delta behave the same way by default, i.e. to not use emph at all for mixed hunks.

And then what I'm proposing here, to use emph for all "added or removed content" and non-emph for "unchanged content in an affected line", can indeed be opt-in. Btw this is how Gerrit and Phabricator highlight.

nkouevda:
Forgot to address the secondary question: I don't have a strong preference here, but for consistency with Gerrit and Phabricator, it would be emph for the entire line, i.e. not how this PR currently behaves.

dandavison:
> In your screenshots, GitHub/GitLab have emph highlighting in the first hunk (+1/-1) but not in the second hunk (+5/-1), whereas [current] delta has emph highlighting in both. That's the issue I'm trying to address, and where delta currently differs from GitHub/GitLab: for hunks where there's a mix of changed and inserted/deleted lines, delta still does emph for the changed lines, making such "mixed" hunks more difficult to understand.

OK, this is very helpful. I didn't realize that delta behaved differently from GitHub in this regard. Here's an illustration of what you're pointing out:

<table>
  <tr>
    <td>
      delta vs. github
    </td>
    <td>
      image
    </td>
    <td>
      image
    </td>
  </tr>
</table>

[test-multiline-changes-base...test-multiline-changes](https://github.com/dandavison/delta/compare/test-multiline-changes-base...test-multiline-changes?expand=1).

I think there are two things to say about this comparison:

1. **Regarding a.txt**: As you point out, GitHub uses the emph color for entirely changed lines in this context, i.e. when they are mixed with partially inserted/deleted lines. That's despite the fact that otherwise, Github uses the non-emph color for entirely changed lines. I didn't realize that.

2. **Regarding b.txt**: GitHub only attempts to pair up lines for comparison when there is an equal number of red and green lines. I had noticed that, and decided to try to "outdo" GitHub by pairing up lines even when the number of red and green lines are unequal. (Delta does this by considering two lines to be a "homologous pair" if they exceed a configurable similarity threshold: `--max-line-distance`. It only considers solutions in which the order of the lines is the same before and after.)

So, do you think it makes sense to

1. Switch delta to use emph color for entirely changed lines in mixed blocks.

2. Allow delta to continue to do its thing where it attempts to identify homologous lines even when there are unequal numbers of removed and added lines.

?

Btw, it looks like GitLab behaves the same as GitHub:

image

nkouevda:
Oh interesting, I also didn't know that GitHub's behavior in a.txt was a thing.

So for mixed hunks, options are:

1. [current delta behavior] emph for partially changed lines, and no emph for whole inserted/deleted lines
2. no emph at all
3. [consistent with Phab/Gerrit] emph for both partially changed and whole inserted/deleted lines
4. [consistent with GitHub/GitLab] mix of (2) and (3): if same number of + and - lines, emph for both partially changed and whole inserted/deleted lines; else no emph at all

IMO (1) and (2) aren't great, and then it's up to you if you want default delta behavior to be consistent with GitHub/(4), but either way I think having (3) would be great, at least opt-in.

dandavison:
Right, agreed, thanks for the analysis, I'll think about it!

> Btw this is how Gerrit and Phabricator highlight.

This was really helpful info also, thanks.

waldyrious:
@dandavison sorry if this has been discussed elsewhere, but looking at the screenshots above, I thought I'd mention that the behavior that would seem intuitive to me is for wholly added or wholly removed lines to be wholly emphasized. Something like this (done with some crude image editing):

<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

dandavison:
Hi @waldyrious and @nkouevda, I'm happy you commented on this. Since @nkouevda and I first discussed this last year, Delta has gained a lot of flexibility for specifying colors, so I was intending to revisit this work and see if it would make sense to implement something using the new options framework, even if as an option rather than a breaking change.

dandavison:
> For unpaired edits, since the whole line is a change, it should be colored with `emph_*`.

> the behavior that would seem intuitive to me is for wholly added or wholly removed lines to be wholly emphasized.

This is now possible on `master`, see https://github.com/dandavison/delta/pull/785#issuecomment-981179697 (copied below)

Thanks very much @nkouevda for opening this a long time ago!

--------------------------------

Basically all you have to do is

```gitconfig
    minus-style = minus-emph-style
    plus-style = plus-emph-style
```

Below is an example theme that does it (I would really like to have examples of themes that work well on dark terminal backgrounds; please let me know if you have some!)

```gitconfig
[delta]
    features = hoopoe2 consistent-color-semantics

[delta "consistent-color-semantics"]
    # All added/removed content is styled the same, regardless of
    # whether there is unchanged content on the same line.
    minus-style = minus-emph-style
    plus-style = plus-emph-style

[delta "hoopoe2"]
    light = true

    pink = "#ffe0e0"
    dark-pink = "#ffc0c0"
    green = "#d0ffd0"
    dark-green = "#a0efa0"

    minus-emph-style = normal hoopoe2.dark-pink
    minus-non-emph-style = normal hoopoe2.pink
    plus-emph-style = syntax hoopoe2.dark-green
    plus-non-emph-style = syntax hoopoe2.green
```

PR #17, opened 2019-10-02
Replace tabs with spaces

### LINKED ISSUE #16: Add option to set the width of a tab character
Delta currently cannot calculate the correct line width when there are tabs present because it doesn't know how much width the tab takes up when it is rendered.

_Originally posted by @dandavison in https://github.com/dandavison/delta/issues/14#issuecomment-537264100_

This can be resolved by adding a flag which sets the display width of a tab character, defaulting to something sensible (2, 4, or 8).

dandavison:
Replace (leading) tabs with spaces, and add a `--tabs` option to control this. Fixes #16


iomintz:
Works great, tyvm!

PR #20, opened 2019-10-13
Add dyn to suppress warning

lzybkr:


dandavison:
Thanks! I'm pretty new to the Rust language and hadn't caught up with the `dyn` keyword, but this LGTM. (I see the warnings in the Travis build output; perhaps I need to update my local build environment.)

PR #28, opened 2019-10-16
Extended separators to better match github diff

ath3:
Left is with the original separators, right is with the extended:

delta_sep


dandavison:
Excellent, thanks very much @ath3. I can see this is strictly an improvement.

(By the way, do let me know if you ever experiment with `--max-line-distance` and form any opinions. I'm conscious that the default there may not be optimal currently.)

PR #30, opened 2019-10-16
Use $BAT_PAGER || $PAGER instead of hardcoded less

myfreeweb:
Among other things, this allows passing custom flags to less, which disables the default flags, which include --no-init (-X) which removes Page Down / Page Up handling.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -58,7 +58,7 @@ fn main() -> std::io::Result<()> {
     }
 
     let mut output_type =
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks for the contribution! Sorry for being slow here. I've been wanting this change for a while also.

I'll remove the `pager` entry from the `config` struct after merging this.

Maybe worth mentioning here for others that custom options can also be passed to less using the `LESS` environment variable.

PR #33, opened 2019-10-17
Replace all tabs in line with spaces

### LINKED ISSUE #27: Unusual context box
The context box here is not sized correctly. It's also an odd attribution, but maybe that's `git` and not `delta`.

image

To repro:

```
git clone clone https://github.com/lzybkr/less
cd less
git show 163de2d
```

dandavison:
@lzybkr Would you be able to confirm that this fixes #27 for you?

- Previously, delta was replacing tabs with spaces when the tabs occurred at the beginning of the line only.
- This PR changes that so that tabs are replaced throughout the line.

The number of spaces to use as a tab replacement is configurable (`--tabs`), defaulting to 4.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -266,7 +266,6 @@ fn prepare(line: &str, tab_width: usize, append_newline: bool) -> String {
     let terminator = if append_newline { "\n" } else { "" };
     if !line.is_empty() {
         let mut line = line.graphemes(true);
-        let mut output_line;
 
         // The first column contains a -/+/space character, added by git. We skip it here and insert
         // a replacement space when formatting the line below.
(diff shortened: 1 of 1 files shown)


src/delta.rs
```diff
@@ -171,10 +170,11 @@ fn handle_hunk_meta_line(
         cli::SectionStyle::Plain => panic!(),
     };
     let ansi_style = Blue.normal();
-    let (code_fragment, line_number) = parse::parse_hunk_metadata(&line);
+    let (raw_code_fragment, line_number) = parse::parse_hunk_metadata(&line);
+    let code_fragment = prepare(raw_code_fragment, config.tab_width, false);
```
dandavison: Previously we were only calling the `prepare` function on code lines within diff hunks; this change calls `prepare` on the code fragment line captured by git to act as a hunk context marker.

src/delta.rs
```diff
@@ -253,30 +257,34 @@ fn handle_hunk_line(painter: &mut Painter, line: &str, state: State, config: &Co
...
 // highlight correctly.
 // See https://docs.rs/syntect/3.2.0/syntect/parsing/struct.SyntaxSetBuilder.html#method.add_from_folder
-fn prepare(line: &str, tab_width: usize) -> String {
+fn prepare(line: &str, tab_width: usize, append_newline: bool) -> String {
+    let terminator = if append_newline { "\n" } else { "" };
     if !line.is_empty() {
-        let mut line = line.graphemes(true).peekable();
-        let mut left_fill = "".to_string();
+        let mut line = line.graphemes(true);
+        let mut output_line;
```
lzybkr: I get a warning here:

```
warning: variable does not need to be mutable
   --> src\delta.rs:269:13
    |
269 |         let mut output_line;
    |             ----^^^^^^^^^^^
    |             |
    |             help: remove this `mut`
    |
```

If an older version or `rustc` requires the `mut`, you could change the assignment to:

```
let output_line = if tab_width > 0 { ... } else { ... }
```

lzybkr:
Fix is confirmed.

dandavison:
Thanks for alerting me to the compiler warnings; I've run `rustup update` and see them locally now.

> you could change the assignment to:
> `let output_line = if tab_width > 0 { ... } else { ... }`

Thanks, I took that suggestion.

PR #36, opened 2019-10-18
Handle unrecognised file types

### LINKED ISSUE #34: Handle unrecognised file types
Treat as .txt is treated currently; no syntax highlighting, but otherwise as normal.

dandavison:
Fixes #34

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -279,6 +279,9 @@ fn prepare(line: &str, tab_width: usize) -> String {
 mod tests {
     use super::*;
     use console::strip_ansi_codes;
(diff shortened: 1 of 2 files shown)


dandavison:
cc @lzybkr This branch changes delta so that it applies background diff highlighting to all files, regardless of extension. (For example, I believe it is an improvement in a repo like `less`, which has some files with file extensions currently unsupported by delta/bat, as well as some with no extension.)

dandavison:
Thanks! I'm going to try to add some unit tests to this branch before merging.

PR #55, opened 2019-11-14
Properly calculate string widths with unicode_width

xen0n:
Fixes diffs rendering with lines containing CJK characters.

P.S. this software is neat, really like it ;-)


dandavison:
Thanks! Could you explain the bug that you're fixing, maybe post a screenshot, and show me how to reproduce? Adding a unit test would be great too!

xen0n:
Hi @dandavison,

The previous width calculation implied that 1 grapheme = 1 char, which is false for at least CJK characters. Example diff input:

```diff
diff --git a/test.txt b/test.txt
index 66a5cb0..0e8c1b9 100644
--- a/test.txt
+++ b/test.txt
@@ -1,3 +1,10 @@
-This is an English sentence. This is an English sentence. This is an English sentence. This is an English sentence. This is an English sentence.
+This is an English sentence. This is an English sentence. This is an English sentence. This is an English sentence. This is an English sentence. some more words
 
-这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。
+这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。这是一个汉语句子。多点字
+
+Another line of English
+更多一行汉语
+Another line of English
+更多一行汉语
+Another line of English
+更多一行汉语
```

Before the change:

before

Note the excess colored cells at the end of lines containing Chinese. After switching to `unicode-width` things are back to normal:

after

I'll see how to add test for this later today.

dandavison:
@xen0n thanks for explaining and thanks again for doing this, this is really helpful. I'm not knowledgable about this area -- I had naively thought I was doing it correctly with `unicode-segmentation`. If you are able to add a unit test demonstrating the previous failure to correctly handle CJK characters in diffs that would be fantastic.

dandavison:
@xen0n thanks very much for this fix. Sorry for leaving it so long, merging now. I haven't added a unit test yet but I've added your test diff to the `testing/` dir of the `repo.

Issue #53, opened 2019-11-12
Support diff -u ... | delta


fdcds:
It would be nice if `diff -u ... | delta` was supported. Currently `delta` does not appear to parse the header correctly and instead passes on some garbage:
```
❯ diff -u a b | delta
diff -u a/file b/file
--- a/file      2019-11-12 19:33:40.199834121 +0100

file    2019-11-12
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
+++ b/file      2019-11-12 19:33:54.767769647 +0100

1
 a                                                                                                                                                                                                                                                                                                                                                                                                                                       
 b                                                                                                                                                                                                                                                                                                                                                                                                                                       
 c                                                                                                                                                                                                                                                                                                                                                                                                                                       
 x                                                                                                                                                                                                                                                                                                                                                                                                                                       
   a                                                                                                                                                                                                                                                                                                                                                                                                                                     
   b                                                                                                                                                                                                                                                                                                                                                                                                                                     
   c                                                                                                                                                                                                                                                                                                                                                                                                                                     
 y                                                                                                                                                                                                                                                                                                                                                                                                                                       
 iff -u a/file2 b/file2                                                                                                                                                                                                                                                                                                                                                                                                                  
--- a/file2     2019-11-12 19:40:49.344885714 +0100

file2   2019-11-12
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
+++ b/file2     2019-11-12 19:40:54.800860547 +0100

1 
 1                                                                                                                                                                                                                                                                                                                                                                                                                                       
 2   
```

Note how the `diff` command for `file2` even lost its first character. (It actually displays like that and is not a copy & paste mistake by me.)

Using delta version 0.0.14.

dandavison:
Thanks, absolutely agree this makes sense. (Indeed it makes me uncomfortable that the package name is `git-delta` seeing as I would like it to support standard unified diff when not embedded in git output.)

Incidentally, this would make a good issue for anyone looking for a smallish Rust exercise.

m-lima:
[...]

PR #56, opened 2019-11-17
Support unified diff format

m-lima:
As discussed on https://github.com/dandavison/delta/issues/53, I wasn't able to see the behavior reported. However, there is an assumption that the file name will be prepended by `a/` and `b/` as it comes from `git diff`.

I ended up generating a new state to represent that the output is coming from `git diff`, since I didn't want to assume that the user wans't comparing files inside directories `a` and `b`, respectively

Some `cargo fmt` came in as well.. I can undo these, if you wish

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -7,6 +7,7 @@ use unicode_segmentation::UnicodeSegmentation;
 use crate::bat::assets::HighlightingAssets;
 use crate::cli;
 use crate::config::Config;
+use crate::delta::State::GitFileMeta;
 use crate::draw;
 use crate::paint::Painter;
 use crate::parse;
@@ -14,12 +15,13 @@ use crate::style;
 
 #[derive(Debug, PartialEq)]
 pub enum State {
-    CommitMeta, // In commit metadata section
-    FileMeta,   // In diff metadata section, between commit metadata and first hunk
-    HunkMeta,   // In hunk metadata line
-    HunkZero,   // In hunk; unchanged line
-    HunkMinus,  // In hunk; removed line
-    HunkPlus,   // In hunk; added line
+    CommitMeta,   // In commit metadata section
+    GitFileMeta,  // In git diff metadata section, between commit metadata and first hunk
+    DiffFileMeta, // In diff unified metadata section, before first hunk
+    HunkMeta,     // In hunk metadata line
+    HunkZero,     // In hunk; unchanged line
+    HunkMinus,    // In hunk; removed line
+    HunkPlus,     // In hunk; added line
     Unknown,
 }
 
@@ -35,14 +37,15 @@ impl State {
 // Possible transitions, with actions on entry:
 //
 //
-// | from \ to  | CommitMeta  | FileMeta    | HunkMeta    | HunkZero    | HunkMinus   | HunkPlus |
-// |------------+-------------+-------------+-------------+-------------+-------------+----------|
-// | CommitMeta | emit        | emit        |             |             |             |          |
-// | FileMeta   |             | emit        | emit        |             |             |          |
-// | HunkMeta   |             |             |             | emit        | push        | push     |
-// | HunkZero   | emit        | emit        | emit        | emit        | push        | push     |
-// | HunkMinus  | flush, emit | flush, emit | flush, emit | flush, emit | push        | push     |
-// | HunkPlus   | flush, emit | flush, emit | flush, emit | flush, emit | flush, push | push     |
+// | from \ to   | CommitMeta  | FileMeta    | HunkMeta    | HunkZero    | HunkMinus   | HunkPlus |
+// |-------------+-------------+-------------+-------------+-------------+-------------+----------|
+// | CommitMeta  | emit        | emit        |             |             |             |          |
+// | GitFileMeta |             | emit        | emit        |             |             |          |
+// | DiffFileMeta|             | emit        | emit        |             |             |          |
+// | HunkMeta    |             |             |             | emit        | push        | push     |
+// | HunkZero    | emit        | emit        | emit        | emit        | push        | push     |
+// | HunkMinus   | flush, emit | flush, emit | flush, emit | flush, emit | push        | push     |
+// | HunkPlus    | flush, emit | flush, emit | flush, emit | flush, emit | flush, push | push     |
 
 pub fn delta<I>(
     lines: I,
@@ -57,7 +60,6 @@ where
     let mut minus_file = "".to_string();
     let mut plus_file;
     let mut state = State::Unknown;
-    let mut git_diff = false;
 
     for raw_line in lines {
         let line = strip_ansi_codes(&raw_line).to_string();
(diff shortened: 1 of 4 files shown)


src/delta.rs
```diff
@@ -70,16 +73,20 @@ where
             }
         } else if line.starts_with("diff --git ") {
             painter.paint_buffered_lines();
-            state = State::FileMeta;
+            state = State::GitFileMeta;
+            painter.set_syntax(parse::get_file_extension_from_diff_line(&line));
+        } else if line.starts_with("diff -u ") || line.starts_with("diff -U") {
```
dandavison: Shall we include the trailing space after both?

m-lima:
Hi Dan, thanks for the review. I totally agree that some tests are needed. Just found a couple more new corner cases. I'll go ahead and plop in this different scenarios to be tested

As for the "renamed/modified/added", what about something like:
`comparing: /tmp/was.rs 2019-11-19 ⟶ ./src/delta.rs 2019-11-18`
To give the context of what "added to" and "removed from" lines mean?

src/delta.rs
m-lima: The `-U` parameter takes an argument (the number of context lines)
So something like `diff -U3 ...` is a valid command

I avoided a regex matcher because `-U` is the only parameter that begins with a capital U, so it seemed unnecessary overhead.

Let me know what you think. I'm happy to either extend the check and look for spaces or digits, or make a regex pattern

dandavison: > The `-U` parameter takes an argument

Oh, I see. Makes sense then, I like your simple and efficient `starts_with` implementation.

But, sorry if I'm being stupid, I don't see a string like `diff -u` in unified diff output, do you?

```
$ diff -u <(echo a) <(echo b)
--- /dev/fd/63  2019-11-20 10:08:51.973225512 -0500
+++ /dev/fd/62  2019-11-20 10:08:51.973585959 -0500
@@ -1 +1 @@
-a
+b
```

```
$ diff --version
diff (GNU diffutils) 3.6
```

However, the output pasted by @fdcds in the intro of #53 does show a string like that, so I'm a bit confused. Is there a different version of `diff` that outputs such a string? (On a Mac the system diff and the one from the coreutils brew package seem to be the same.) Seems the same on linux:

```
$ docker run ubuntu bash -c 'diff -u <(echo a) <(echo b); uname -a; diff --version'
--- /dev/fd/63  2019-11-20 15:17:14.345904000 +0000
+++ /dev/fd/62  2019-11-20 15:17:14.345904000 +0000
@@ -1 +1 @@
-a
+b
Linux 7c3c98a876c6 4.9.184-linuxkit #1 SMP Tue Jul 2 22:58:16 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
diff (GNU diffutils) 3.6
```

m-lima: Yeah.. This is what I commented on https://github.com/dandavison/delta/issues/53, it wasn't very clear what was happening and I couldn't replicate it. I was very confused at first. But 
I did manage to find out what he meant, though. You need to `diff -u <directory1> <directory2>`, this will output `diff -u ...` before each file, or `Only in <directory1>` for files that do not have a name match

dandavison: Right, I see, thanks for explaining.

src/delta.rs
```diff
@@ -7,21 +7,26 @@ use unicode_segmentation::UnicodeSegmentation;
...
+    HunkZero,   // In hunk; unchanged line
+    HunkMinus,  // In hunk; removed line
+    HunkPlus,   // In hunk; added line
+    Unknown,
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Source {
+    GitDiff,     // Coming from a `git diff` command
+    DiffUnified, // Coming from a `git -u` command
```
dandavison: Little typo here: s/git/diff/

src/delta.rs
```diff
@@ -56,60 +60,80 @@ pub fn delta<I>(
...
 
-    for raw_line in lines {
-        let line = strip_ansi_codes(&raw_line).to_string();
-        if line.starts_with("commit ") {
-            painter.paint_buffered_lines();
-            state = State::CommitMeta;
-            if config.opt.commit_style != cli::SectionStyle::Plain {
+    for raw_line in lines_peekable {
+        if source == Source::Unknown {
+            writeln!(painter.writer, "{}", raw_line)?;
```
dandavison: Do you think a `continue` statement here would make the code more readable?

src/delta.rs
```diff
@@ -252,6 +313,28 @@ fn handle_hunk_line(painter: &mut Painter, line: &str, state: State, config: &Co
     }
 }
 
+fn handle_file_uniqueness(
```
dandavison: When I read the code I struggled a bit to understand this function name. The other analogous functions are called `handle_commit_meta_header_line`, `handle_file_meta_header_line`, `handle_hunk_meta_line`, `handle_hunk_line`.

WDYT about `handle_unified_diff_unique_file_name`? And would you mind adding a brief docstring explaining what it does?

src/delta.rs
```diff
@@ -539,6 +622,54 @@ mod tests {
...
+    #[test]
+    fn test_diff_unified_two_files() {
+        let options = get_command_line_options();
+        let output = strip_ansi_codes(&run_delta(DIFF_UNIFIED_TWO_FILES, &options)).to_string();
+        let mut lines = output.split('\n');
+
+        // Header
+        assert_eq!(
+            lines.nth(1).unwrap(),
+            "comparing: one.rs\t2019-11-20 ⟶   src/two.rs\t2019-11-18"
```
dandavison: I'm hesitant about including these dates with the file names. What's your thinking on this? I'm kind of thinking that timestamps are orthogonal to the question of how two files differ and we should use 

```comparing: one.rs ⟶   src/two.rs```.

File: src/parse.rs [hunk 6]
```diff
@@ -25,19 +35,33 @@ pub fn get_file_path_from_file_meta_line(line: &str) -> String {
...
-        (minus_file, "/dev/null") => format!("deleted: {}", minus_file),
-        ("/dev/null", plus_file) => format!("added: {}", plus_file),
-        (minus_file, plus_file) => format!("renamed: {} ⟶   {}", minus_file, plus_file),
+pub fn get_file_change_description_from_file_paths(
+    minus_file: &str,
+    plus_file: &str,
+    comparing: bool,
+) -> String {
+    if comparing {
+        format!("comparing: {} ⟶   {}", minus_file, plus_file)
```
dandavison: Thanks, I like your style decision here to make this harmonious with the git diff "renamed" output. (I just have that concern above about not including the date timestamps).

src/delta.rs
dandavison: Or `handle_directory_diff_unique_file_name`?

dandavison:
Here's an example and a question about the output. This creates two directories that have two comparable files, and also one unique file:

```sh
mkdir /tmp/z1 /tmp/z2
git show HEAD~10:./src/delta.rs > /tmp/z1/a.rs
git show HEAD:./src/delta.rs > /tmp/z2/a.rs
touch /tmp/z2/b.rs
git show HEAD~10:./src/paint.rs > /tmp/z1/c.rs
git show HEAD:./src/paint.rs > /tmp/z2/c.rs
diff -u /tmp/z1 /tmp/z2 | delta
```

At the end of the diff we get

image

Now, did we expect the `Only in` text to be blue? And would it be appropriate for it to have the same underline styling as the other file headers sections?

m-lima:
Thanks for the input, Dan. This is great feedback :)

To address some of your points:
- Yes! It should be blue! It was suppose to have the same markers as a new file block. I'll look into it
- I kind of liked to have the date, because it allows the user to know which file is more recent. But I see your point and I also know that delta is about a clean and simple output. I'll fix this
- Because of your later comment, I suppose you thought more about it. So I'll go for your last suggestion: handle_directory_diff_unique_file_name

src/delta.rs
```diff
@@ -90,11 +114,20 @@ where
                 handle_hunk_meta_line(&mut painter, &line, config)?;
                 continue;
             }
+        } else if source == Source::DiffUnified && line.starts_with("Only in ") {
+            state = State::FileMeta;
```
m-lima: I want to avoid generating a new state, and since this section should behave like a new file block, I used the same state
Let me know if you'd rather have a new state for it

dandavison: Reusing the same state here makes sense to me.

dandavison:
Thanks very much @m-lima, this is a great contribution.

Incidentally, I wonder about the package name `git-delta` that I've been using (due to `delta` being unavailable on the standard package repositories), since you've now made `delta` genuinely support a non-git diff format.

m-lima:
It was a pleasure picking up an issue with this project and a great first experience contributing to OSS :)

PR #57, opened 2019-11-18
Make sure rename only shows up once

m-lima:
If files are renamed and modified, `delta` will output the renaming twice.
For instance, given the following `git diff` output:

```
diff --git a/one.rs b/two.rs
similarity index 99%
rename from one.rs
rename to two.rs
index 50fc1ba..37f64c4 100644
--- a/one.rs
+++ b/two.rs
@@ -19,7 +19,6 @@
 ### Examples

 These are changes:
-one:
+two:
```

then `renamed: one.rs ⟶   two.rs` will show up twice

This PR assess that by resetting the file names on each `FileMeta` state entry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,9 @@ test:
                                           --hunk-style plain \
                                   | ansifilter | cut -c 2-)"
 
+release:
+	cargo publish
+
 hash:
 	@version=$$(grep version Cargo.toml | head -n1 | sed -E 's,.*version = "([^"]+)",\1,') && \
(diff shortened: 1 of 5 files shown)


dandavison:
Thanks for this and #56! I'll try to review them soon.

dandavison:
@m-lima could you merge, or rebase on, master here?

waldyrious:
This has now been fixed in c68bb01.

dandavison:
Thanks @waldyrious and @m-lima. I'm not quite sure why I let this PR stagnate but as @waldyrious says we should have the functionality now.

PR #62, opened 2019-11-26
Handle git diff submodule change lines

### LINKED ISSUE #60: 🐛Handle diff.submodule log
It appears that delta has some trouble finding the end of a hunk:
```diff
--- a   2019-11-27 12:06:18.472973416 +0100
+++ b   2019-11-27 12:06:23.992967043 +0100
@@ -1,7 +1,7 @@
 x
 y
 z
-a
+b
 z
 y
 x
Submodule x/y/z contains untracked content
```

Is displayed (roughly) as:
```
a
---------------------------------------
-----|
 ... |
-----|
1
 x
 y
 z
-a
+b
 z
 y
 x
 ubmodule x/y/z contains untracked content
```

Note how "Submodule"'s first character gets removed by delta.

Using delta 0.0.14.

dandavison:
Fixes #60. @m-lima I'm going to leave this open for a bit, just in case you're interested in reviewing. cc @fdcds

## How this was tested
New unit test: `test_submodule_contains_untracked_content`

Also

```bash
bash testing/56-unified-directory-diff.sh | delta
```

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -115,11 +115,17 @@ where
                 continue;
             }
         } else if source == Source::DiffUnified && line.starts_with("Only in ") {
+            // Additional FileMeta cases:
+            //
+            // 1. When comparing directories with diff -u, if filenames match between the
+            //    directories, the files themselves will be compared. However, if an equivalent
+            //    filename is not present, diff outputs a single line (Only in...) starting
+            //    indicating that the file is present in only one of the directories.
             state = State::FileMeta;
             painter.paint_buffered_lines();
             if config.opt.file_style != cli::SectionStyle::Plain {
                 painter.emit()?;
-                handle_directory_diff_unique_file_name(&mut painter, &raw_line, config)?;
+                handle_generic_file_meta_header_line(&mut painter, &raw_line, config)?;
                 continue;
             }
         } else if state.is_in_hunk() {
@@ -190,12 +196,23 @@ fn handle_commit_meta_header_line(
     Ok(())
 }
 
+/// Construct file change line from minus and plus file and write with FileMeta styling.
 fn handle_file_meta_header_line(
     painter: &mut Painter,
     minus_file: &str,
     plus_file: &str,
     config: &Config,
     comparing: bool,
+) -> std::io::Result<()> {
+    let line = parse::get_file_change_description_from_file_paths(minus_file, plus_file, comparing);
+    handle_generic_file_meta_header_line(painter, &line, config)
+}
+
+/// Write `line` with FileMeta styling.
+fn handle_generic_file_meta_header_line(
+    painter: &mut Painter,
+    line: &str,
+    config: &Config,
 ) -> std::io::Result<()> {
     let draw_fn = match config.opt.file_style {
         cli::SectionStyle::Box => draw::write_boxed_with_line,
@@ -206,9 +223,7 @@ fn handle_file_meta_header_line(
     writeln!(painter.writer)?;
     draw_fn(
         painter.writer,
-        &ansi_style.paint(parse::get_file_change_description_from_file_paths(
-            minus_file, plus_file, comparing,
-        )),
+        &ansi_style.paint(line),
         config.terminal_width,
         ansi_style,
         false,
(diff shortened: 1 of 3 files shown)


testing/60-submodule.sh
```diff
@@ -0,0 +1,13 @@
...
+repo=$(mktemp -d)
+cd $submodule
+git init
+git commit --allow-empty -m "Initial commit"
+cd $repo
+git init
+git submodule add $submodule submodule
+git commit --allow-empty -m "Initial commit"
+touch submodule/a
+git diff | cat
```
dandavison: @fdcds Do you know how to reproduce the "Submodule XXX contains untracked content" message? This script was my first attempt, but it yields

```diff
diff --git a/submodule b/submodule
--- a/submodule
+++ b/submodule
@@ -1 +1 @@
-Subproject commit 4def2debb586b2ae007ca8589d52efeb9a827432
+Subproject commit 4def2debb586b2ae007ca8589d52efeb9a827432-dirty
```

src/delta.rs
```diff
@@ -114,12 +114,27 @@ where
...
+            //
+            // See https://github.com/dandavison/delta/issues/60#issuecomment-557485242 for a
+            // proposal for more robust parsing logic.
+
             state = State::FileMeta;
             painter.paint_buffered_lines();
             if config.opt.file_style != cli::SectionStyle::Plain {
                 painter.emit()?;
-                handle_directory_diff_unique_file_name(&mut painter, &raw_line, config)?;
+                handle_generic_file_meta_header_line(&mut painter, &raw_line, config)?;
```
dandavison: @m-lima I thought that your function `handle_directory_diff_unique_file_name` could actually be used more broadly to implement all `FileMeta` output, so I repurposed it and moved your explanatory comments to this call site.

src/delta.rs
```diff
@@ -114,12 +114,27 @@ where
...
+            // 1. When comparing directories with diff -u, if filenames match between the
+            //    directories, the files themselves will be compared. However, if an equivalent
+            //    filename is not present, diff outputs a single line (Only in...) starting
+            //    indicating that the file is present in only one of the directories.
+            //
+            // 2. Git diff emits lines describing submodule state such as "Submodule x/y/z contains
+            //    untracked content"
+            //
+            // See https://github.com/dandavison/delta/issues/60#issuecomment-557485242 for a
+            // proposal for more robust parsing logic.
```
dandavison: @fdcds your implementation proposal in #60 sounds much more robust than what I've done here. I didn't feel quite ready to go ahead with it yet so have gone with this "easy option". This state machine part of the delta codebase is really hard to understand and fragile, as I expect @m-lima will agree with. I think it would make sense to investigate better abstractions for implementing the parser here.

src/delta.rs
mlimat: Nice! Makes a lot of sense.
Especially if you decide in the future change the appearance of some of the `FileMeta` lines, you can easily plug them in 👍

m-lima:
Makes sense. Simple and effective for this case.

As to your comment, I'll add my two cents :)
I think it is good that the state machine never goes into `State::Unknown` from a known state. That means, however, that once in the `State::Hunk*`, the parser gets "stuck" there until recognizing a known pattern (thus the clipping of the lines, for example)

A possible way to get out of the `State::Hunk*` without going through the trouble of parsing the `^@@` context is at `delta.rs:34`, where `is_in_hunk` could check for `^[ +-]`. But leaving the `State::Hunk*` would put the parser in an unknown state.
On the other hand, this might be preferable to having `delta.rs:117` grow larger and larger with the checks for new patterns as new cases appear.

src/delta.rs
m-lima: Nice! Makes a lot of sense.
Especially if you decide in the future change the appearance of some of the `FileMeta` lines, you can easily plug them in 👍

testing/60-submodule.sh
fdcds: This is the (at least remotely) related part of my `~/.gitconfig`:
```ini
[rebase]
#       autostash = true    
#       autosquash = true    
        stat = true
[fetch]
        prune = true
[merge]
        defaultToUpstream = yes
        log = yes
[color]
        ui = auto
[log]
        decorate = true
[diff]
        submodule = log
[status]
        submodulesummary = true
[core]
        pager = delta --dark
```

I suspect `diff.submodule = log` to be the part that causes git to output the message you are missing. Can you reproduce with that?

src/delta.rs
fdcds: /cc @devurandom

testing/60-submodule.sh
dandavison: Thanks, yes that's it. 9736ddb1c368e834d3f68845025c840a450b9d68

dandavison:
Thanks @m-lima and @fdcds!

PR #67, opened 2019-12-04
README: Document env. var to fix mouse scrolling (fix #58)

### LINKED ISSUE #58: Support mouse scroll
In the default git pager (`less` on my system) I can use the mouse scroll wheel to scroll through the pager output. When `delta` is used, the mouse scroll does not work, and I'm forced to use the navigation shortcuts. It would be really nice to be able to scroll with the mouse wheel.

System: `Linux pop-os 5.3.0-22-generic #24+system76~1573659475~19.10~26b2022-Ubuntu SMP Wed Nov 13 20:0 x86_64 x86_64 x86_64 GNU/Linux`

ronjouch:


dandavison:
Excellent, thanks very much! LGTM.

Issue #19, opened 2019-10-11
Config file and custom themes similar to bat


clarfon:
Would be very nice to have.

dandavison:
Hi, thanks for the feedback!

Regarding a config file, `delta` is nearly always used in conjunction with `git`, so the `~/.gitconfig` file plays the role of `bat`'s config file. I.e. both files specify the desired command-line invocation. Would you agree that we don't need an additional config file beyond entering  the command-line invocation in `~/.gitconfig`?

For custom themes, could you expand on what you'd like `delta` to add? Currently we have the themes below available. In order to add a new theme, the process is actually to clone `bat` and add the theme in the `bat` repo following the [bat instructions](https://github.com/sharkdp/bat#adding-new-themes)! Then, copy the `assets/themes.bin` from the `bat` repo to the `delta` repo. This is referenced [here](https://github.com/dandavison/delta#supported-languages-and-themes) but perhaps I could spell it out a bit more.

```
$ delta --list-themes 
Light themes:
    GitHub
    Monokai Extended Light
    OneHalfLight
    ansi-light
Dark themes:
    1337
    DarkNeon
    Monokai Extended
    Monokai Extended Bright
    Monokai Extended Origin
    OneHalfDark
    Sublime Snazzy
    TwoDark
    ansi-dark
    base16
    zenburn
```

clarfon:
Certainly! I accidentally pressed enter when composing this and submitted it prematulrely.

In terms of config… you're right that gitconfig takes that role, and I think we'd be fine without a separate config file. It is nice that `bat` offers a bit more control and allows a config file which lets you specify command-line options ahead of time, and being able to do that might be a bit more user-friendly than gitconfig.

In terms of custom themes, the main issue is having to recompile the binary every time. It'd be nice if they could be loaded dynamically from e.g. `~/.config` somewhere. Honestly, if we're going to just pull from bat's config, I'd be fine with just using `~/.config/bat/themes`.

mgred:
@dandavison 

> In order to add a new theme, the process is actually to clone `bat` and add the theme in the `bat` repo following the [bat instructions](https://github.com/sharkdp/bat#adding-new-themes)! Then, copy the `assets/themes.bin` from the `bat` repo to the `delta` repo.

[...]

PR #68, opened 2019-12-06
feat: Solarized themes

hakamadare:
adding Solarized (light/dark) themes from
https://github.com/paulcpederson/solarized-sublime


dandavison:
Awesome, thanks! I believe we want to add `Solarized (light)` to the list of light themes, so that delta uses the background colors for a light terminal background when we do `delta --theme='Solarized (light)'`. Could you consider adding the following:

```diff
diff --git a/src/style.rs b/src/style.rs
index 4868d12..bbb05a2 100644
--- a/src/style.rs
+++ b/src/style.rs
@@ -1,10 +1,11 @@
 use syntect::highlighting::{Color, FontStyle, Style, StyleModifier};
 
-pub const LIGHT_THEMES: [&str; 4] = [
+pub const LIGHT_THEMES: [&str; 5] = [
     "GitHub",
     "Monokai Extended Light",
     "OneHalfLight",
     "ansi-light",
+    "Solarized (light)",
 ];
 
 pub const DEFAULT_LIGHT_THEME: &str = "GitHub";
```

(or you can add me as a collaborator on your fork/PR; I'm not quite sure what the workflow is supposed to be in this situation.)

dandavison:
I don't think I'm seeing any difference between the foreground colors used by light vs. dark. WDYT, should these look different?

image

My commands there are
- `git show 07f9421 | delta --theme='Solarized (dark)'`
- `git show 07f9421 | delta --theme='Solarized (light)'`

dandavison:
Do you know whether it would make sense to add these themes to [bat](https://github.com/sharkdp/bat) also, or have a suspicion why it hasn't been done already? I think there are some considerations regarding terminal themes that I haven't understood yet (see e.g. https://github.com/sharkdp/bat/issues/490).

hakamadare:
> I don't think I'm seeing any difference between the foreground colors used by light vs. dark. WDYT, should these look different?

the color schemes should not look dramatically different; one of the features of Solarized is that its overall look and feel does not change when switching from light to dark mode (https://ethanschoonover.com/solarized/#features)

solarized-dark-vs-light

see the difference between the two different versions of the same content?  in the dark mode (upper image) the base text is Solarized color "base 0", in the light mode the base text is Solarized color "base 00"

hakamadare:
> Do you know whether it would make sense to add these themes to [bat](https://github.com/sharkdp/bat) also, or have a suspicion why it hasn't been done already? I think there are some considerations regarding color themes for terminal emulators that I haven't understood yet (see e.g. [sharkdp/bat#490](https://github.com/sharkdp/bat/issues/490)).

i'm not opposed to opening a similar PR against bat; the message i took from the bat documentation was that the author wants people to compile their own theme/style bundles and load them at runtime, whereas with delta the theme and style bundles are compiled in.

dandavison:
Thanks!

> the message i took from the bat documentation was that the author wants people to compile their own theme/style bundles and load them at runtime

 There are definitely [accepted PRs](https://github.com/sharkdp/bat/pulls?utf8=%E2%9C%93&q=is%3Apr+theme+) adding themes to bat. cc @sharkdp would adding solarized themes to bat be a helpful contribution?

> whereas with delta the theme and style bundles are compiled in.

Right, currently. But I intend to (a) support all themes that bat supports by default (i.e. update delta periodically to match bat) and (b) support reading custom themes and syntaxes from the same directory that bat uses (`~/.cache/bat`), see #19 and https://github.com/dandavison/delta/issues/58#issuecomment-561769422.

hakamadare:
done! sharkdp/bat#768

PR #76, opened 2019-12-24
read themes from bat's cache

data-pup:
This does not _entirely_ address #19, but does address [this comment](https://github.com/dandavison/delta/issues/19#issuecomment-541437420).

This pulls in some more code from `bat` to read themes from `~/.cache/bat/`. Tested locally on my machine (with an existing `bat` install) and this does work properly. That said, I know this does not include tests: I would be happy to add some, just unsure of where those would go.


dandavison:
@data-pup Thanks very much for this! It looks great; very clean. I've tried it out, rebuilding the bat cache with `bat cache --build` and can see that delta is picking up new theme and syntax files. I guess I have no excuse not to work on an emacs-lisp sublime syntax definition now...

So the one blocker is we need to figure out what's up with the Windows build: https://travis-ci.com/dandavison/delta/jobs/269918710

If you would like to add some tests, src/delta.rs would be a reasonable place.  (It already has some tests related to theme).

data-pup:
@dandavison Hmmm, so I tried running `cargo build --target x86_64-pc-windows-msvc` on `master` at `13e39b1de9ef0a8324b882fade04eeff65241f40`, and `1c3e7baf0f96bc0ff07da8d0aab9b1f2d15b02d2`, and ended up seeing the same build errors. I am _suspicious_ that something elsewhere may have broken the build? I'm not quite sure how to debug this tbh :confounded:

dandavison:
Just FYI I'm going to be offline for the next 10 days so will go quiet, but I'm keen to get this PR merged. Thanks for looking at the build failure.

dandavison:
You're right, I see now that Windows builds have started to fail on master also, so unrelated to this PR. Thanks very much then! I'll merge this now.

PR #77, opened 2019-12-28
Fix panic when showing a diff which contains a diff

Ma27:
When committing e.g. a patch-file in a git repository that has a line
which contains `--- ` only, `delta` fails with the following error (as it
expects such a line to contain keywords for the diff):

```
thread 'delta::tests::test_diff_in_diff' panicked at 'byte index 2 is out of bounds of ``', src/libcore/str/mod.rs:2017:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

This patch changes the diffing behavior to ensure that lines containing
`--- ` only won't be parsed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.com/dandavison/delta.svg?branch=master)](https://travis-ci.com/dandavison/delta)
+[![Gitter](https://badges.gitter.im/dandavison-delta/community.svg)](https://gitter.im/dandavison-delta/community?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
 
 ## A syntax-highlighter for git and diff output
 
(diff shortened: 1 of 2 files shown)


dandavison:
Hi @Ma27, thanks very much for this! (I'm going to be offline for the next ten days so will be slow to review and merge this, but will do ASAP.)

Ma27:
@dandavison would you mind taking another look at this? :)

dandavison:
Hi @Ma27, thanks for the ping and sorry for forgetting about this!

I've been trying to understand some history here... I think what's happened is that, in the interim, #90 merged a similar change to yours to master, so that your test now passes, without any change to the non-test code. Accordingly, I've pushed some commits to this branch, leaving the new unit test as the change that this branch will introduce when merged.

Do you buy all that and does it SGTY?

Ma27:
> Do you buy all that and does it SGTY?

LGTM, thanks! :+1:

dandavison:
Great, thanks very much!

PR #94, opened 2020-02-12
Show binary file diff line

### LINKED ISSUE #93: Show binary files changes in delta output
When I use `GIT_PAGER=less git show` I see the following section for each binary file:

```diff
diff --git a/foo b/foo
new file mode 100644
index 0000000..4088911
Binary files /dev/null and b/foo differ
```

When I use delta with `GIT_PAGER="delta --theme=TwoDark" git show`, I don't see any output for `foo`, which is a binary file.

It would be great to see something that shows that there was a binary file and that it has changed, what less shows would be fine.

Quick repro:

```sh
mkdir -p tst && cd tst
git init
echo -n -e \\x48\\x00\\x49\\x00 > foo
git add foo
GIT_PAGER="delta --theme=TwoDark" git diff --staged
GIT_PAGER=less git diff --staged
```

Output:

```console
$ mkdir -p tst && cd tst
$ git init
Initialized empty Git repository in ~/tmp/tst/.git/
$ echo -n -e \\x48\\x00\\x49\\x00 > foo
$ git add foo
$ GIT_PAGER="delta --theme=TwoDark" git diff --staged
$ GIT_PAGER=less git diff --staged
  diff --git a/foo b/foo
  new file mode 100644
  index 0000000..7a42ff1
  Binary files /dev/null and b/foo differ
```

Thanks for making such a great diff viewer by the way, best one I've found hands down.

dandavison:
Fixes #93 cc @gibfahn

This fixes the issue in my local testing. However, I expected the test added before the fix at af84d5c7019f4481a6d64112d74e4eae727c4bd1 to fail, but it passes. I don't know why yet.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -704,6 +704,13 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_binary_files_differ() {
+        let options = get_command_line_options();
+        let output = strip_ansi_codes(&run_delta(BINARY_FILES_DIFFER, &options)).to_string();
+        assert!(output.contains("Binary files /dev/null and b/foo differ\n"));
+    }
+
     const ADDED_FILE_INPUT: &str = "\
 commit d28dc1ac57e53432567ec5bf19ad49ff90f0f7a5
 Author: Dan Davison <dandavison7@gmail.com>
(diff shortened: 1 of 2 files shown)


dandavison:
Thanks @gibfahn and @Kr1ss-XD!

> No idea why the test passes without the change having been made, but I reproduced that too.

Indeed. Well, progress... merging this. Opened #95 to establish what's going on with this test, and also #96 as I think the `Binary files /dev/null and b/foo differ` message could perhaps be improved as `Binary file foo added`.

PR #98, opened 2020-02-14
Fix missing newline in --commit-style=box

guyboltonking:
When using `--commit-style=box` there's a missing newline; this PR attempts to fix that.

Before:

image

After:

image

### DIFF
--- a/src/draw.rs
+++ b/src/draw.rs
@@ -46,6 +46,7 @@ pub fn write_boxed_with_line(
         line_style,
         heavy,
     )?;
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks! Tested locally, looks great. (A similar newline character [is written](https://github.com/dandavison/delta/blob/cd3226a8f6449ab76bb772d30d74f009a28e1cb0/src/draw.rs#L61) in `write_underlined` so makes sense wee'd ned it here.)

Issue #108, opened 2020-03-01
Add flag to disable 'less' behaviour


jesseduffield:
It looks like delta pipes itself to less when there's more than a page of content. I'm hoping to make it possible for users to use delta in lazygit, but currently it's not possible unless I can get delta to simply output all the content and then exit.

Is this something that already exists? If not, I'm happy to put up a PR if it's simple enough.

PR #109, opened 2020-03-01
add paging mode option

jesseduffield:
fixes https://github.com/dandavison/delta/issues/108

This is my first ever PR in rust, hopefully the first of many!

The use case here is to allow me to use delta in lazygit without having to deal with `less` getting in the way

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -245,7 +245,14 @@ pub fn process_command_line_arguments<'a>(
     let paging_mode = match opt.paging_mode.as_ref() {
         "always" => PagingMode::Always,
         "never" => PagingMode::Never,
-        _ => PagingMode::QuitIfOneScreen,
+        "auto" => PagingMode::QuitIfOneScreen,
+        _ => {
+            eprintln!(
+                "Invalid paging value: {} (valid values are \"always\", \"never\", and \"auto\")",
+                opt.paging_mode
+            );
+            process::exit(1);
+        }
     };
 
     config::get_config(
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -155,7 +155,9 @@ pub struct Opt {
     #[structopt(long = "max-line-distance", default_value = "0.3")]
     pub max_line_distance: f64,
 
-    /// The paging mode. Options are: always, never and auto.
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -153,6 +154,12 @@ pub struct Opt {
     /// one into the other.
     #[structopt(long = "max-line-distance", default_value = "0.3")]
     pub max_line_distance: f64,
+
+    /// Whether to use a pager when displaying output. Options are: auto, always, and never. The
+    /// default pager is `less`: this can be altered by setting the environment variables BAT_PAGER
+    /// or PAGER (BAT_PAGER has priority).
+    #[structopt(long = "paging", default_value = "auto")]
+    pub paging_mode: String,
```
dandavison: Since `delta` is in some ways similar to [bat](https://github.com/sharkdp/bat), I've made it a design guideline to be exactly the same as bat wherever it makes sense. bat has basically exactly the option that you're adding, so I've pushed commits to your PR making the interface identical. Here's a section from `bat --help` for reference.

image

src/cli.rs
```diff
@@ -237,11 +244,25 @@ pub fn process_command_line_arguments<'a>(
...
+        "auto" => PagingMode::QuitIfOneScreen,
+        _ => {
+            eprintln!(
+                "Invalid paging value: {} (valid values are \"always\", \"never\", and \"auto\")",
+                opt.paging_mode
+            );
+            process::exit(1);
+        }
+    };
+
```
dandavison: I made a small change here to demand that if the user supplies a value, it is valid.

src/delta.rs
```diff
@@ -624,6 +624,7 @@ mod tests {
             hunk_style: cli::SectionStyle::Box,
             hunk_color: "blue".to_string(),
             width: Some("variable".to_string()),
+            paging_mode: "auto".to_string(),
```
dandavison: I had to add an entry here to make delta's tests pass.

src/cli.rs
dandavison: Sorry I should be totally explicit here, since lazy git is going to pass this option: I have changed the name of the option in your PR so that it will now be `--paging=never`.

jesseduffield:
Thanks for the quick response :)

Issue #100, opened 2020-02-18
Support `git stash show --stat --patch`


gibfahn:
Looks like `git stash show` doesn't get formatted by delta:

<img width="341" alt="image" src="https://user-images.githubusercontent.com/15943089/74733529-6c830100-5244-11ea-9e6e-8bed110154ff.png">

Using `delta 0.0.16` on macOS.

### Script to reproduce

```shell
export GIT_PAGER='delta --theme=TwoDark'
mkdir tst
cd tst
git init
printf "bar\nbaz\n" >foo_diff
printf "bar\nbaz\n" >foo_stash

git add foo*
git commit -m "Initial Commit"

printf "bat\n" >> foo_stash
git stash
printf "bat\n" >> foo_diff

git diff # Works with delta
git stash show # Doesn't.
```

dandavison:
Hi @gibfahn I believe you have  something like this in your git configuration, right?

```
[stash]
    showStat = true
    showPatch = true
```

I say that because, with out-of-the-box git, I'm seeing reasonable behaviour from delta with git stash:

![image](https://user-images.githubusercontent.com/52205/75650701-5732ad00-5c1c-11ea-8f79-7fa7a14c7263.png)

However, I definitely agree that what you point out is a bug: i.e. that when the `stat` and `patch` outputs are combined, delta fails to recognise the patch:

![image](https://user-images.githubusercontent.com/52205/75650760-847f5b00-5c1c-11ea-827d-832a479970a6.png)

It's fairly clear what the issue is. We [peek at the first line](https://github.com/dandavison/delta/blob/master/src/delta.rs#L186) to see if we recognise the type of input, see `foo_stash | 1 +`, don't recognise that, and thereafter never attempt to parse anything. So it seems that one solution would be to keep trying to recognise sections, even after we initially haven't recognised the type of input.

There is an underlying problem here: while I'm OK with most of delta's code, the [state machine parser](https://github.com/dandavison/delta/blob/master/src/delta.rs#L53-L189) needs to be rewritten!

gibfahn:
> Hi @gibfahn I believe you have something like this in your git configuration, right?
> 
> [stash]
>     showStat = true
>     showPatch = true

Ahh yeah, I should have thought of that, I tend to forget how much I have in my git config...


>So it seems that one solution would be to keep trying to recognise sections, even after we initially haven't recognised the type of input.

Seems reasonable

[...]

PR #123, opened 2020-04-12
Don't give up on parsing

### LINKED ISSUE #121: Highlight diffs when there is unrecognised initial content
cc @m-lima @gibfahn Currently, if delta doesn't recognize the type of diff from the first line, it [gives up](https://github.com/dandavison/delta/blob/master/src/delta.rs#L72) on parsing. I'm aware of two situations where this is too strict:

1. [Magit](https://github.com/magit/magit) makes calls like `git show -p --cc --format=  --numstat --stat`. This results in the `numstat` and `stat` output appearing before the diff, and the diff is not highlighted as a result. (I'm making it possible to use `delta` as a diff rendering plugin for magit.)

2. In #100 delta does not render `git stash show --stat --patch` correctly for the same reason.

dandavison:
cc @m-lima @gibfahn fixes #121 #100 

This is a change to the parsing logic: instead of requiring that we detect the source diff type (git vs unified diff) from the first line, we keep trying to detect the source until we succeed. This causes diffs to be highlighted even when there is unrecognized initial material. Two tests are added for this sort of behavior, but one test is sacrificed -- involving non-diff input that looks diff-like.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -59,20 +59,17 @@ pub fn delta<I>(
 where
     I: Iterator<Item = String>,
 {
-    let mut lines_peekable = lines.peekable();
     let mut painter = Painter::new(writer, config, assets);
     let mut minus_file = "".to_string();
     let mut plus_file;
     let mut state = State::Unknown;
-    let source = detect_source(&mut lines_peekable);
+    let mut source = Source::Unknown;
 
-    for raw_line in lines_peekable {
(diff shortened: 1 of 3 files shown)


dandavison:
Only if you have time, but any thoughts/opinions/sanity-checking etc here would be appreciated @m-lima!

dandavison:
Merging this, comments still welcome at any time.

PR #132, opened 2020-04-27
cli: remove panic on invalid width

wchargin:
Test Plan:
Pass `--width foobar`, and note that the binary still exits with failure
and an appropriate message, but no longer includes a “thread panicked;
run with backtrace” annotation. Note that there are no other explicit
panics in `cli.rs`.

wchargin-branch: cli-no-panic-width


dandavison:
Thanks very much @wchargin!

PR #134, opened 2020-04-27
Fix Arch Linux spelling

svenstaro:
Arch Linux is written either like that or archlinux in lowercase.


dandavison:
Thanks @svenstaro!

PR #143, opened 2020-04-28
chore: Update console to 0.11

ignatenkobrain:
Signed-off-by: Igor Raits <i.gnatenko.brain@gmail.com>


dandavison:
Thanks @ignatenkobrain! So I understand, was the previous version of console causing a problem?

ignatenkobrain:
@dandavison depends what you mean by `causing a problem` :)

We want to get git-delta packaged in Fedora and we try to keep only latest versions of crates there. So for us it is a problem with packaging git-delta..

For others this is probably not a problem :)

dandavison:
OK thanks! Is there a recommended procedure I should follow for keeping crates up-to-date in the way that is convenient for fedora? (I take it this was the only one that needed updating?)

ignatenkobrain:
@dandavison people usually set up @dependabot which automatically sends PR to for version updates.

ignatenkobrain:
So yes, keeping dependencies update would be very helpful, but don't set things like `foo = "^1.2.3"` unless needed, just keep `^1` or `^1.2`.

dandavison:
Thanks, I've registered delta with dependabot.

ignatenkobrain:
@dandavison thanks! would it be possible to release new version on crates.io?

PR #164, opened 2020-04-30
Preserve ANSI sequences if delta is not applying colors

### LINKED ISSUE #163: 🐛 Custom git log coloring is overwritten
When specifying a custom git log format which includes colors (e.g. via `git log --patch --pretty=format:'%C(red)I SHOULD BE RED'`) delta overwrites these custom colors starting from the seconds log message.

image

dandavison:
Fixes #163 cc @0xC0FFEE .

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -138,6 +138,8 @@ where
                 continue;
             }
         } else if state.is_in_hunk() {
+            // A true hunk line should start with one of: '+', '-', ' '. However, handle_hunk_line
+            // handles all lines until the state machine transitions away from the hunk states.
             state = handle_hunk_line(&mut painter, &line, state, config);
             painter.emit()?;
             continue;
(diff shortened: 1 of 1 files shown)


src/delta.rs
```diff
@@ -308,9 +316,7 @@ fn handle_hunk_line(painter: &mut Painter, line: &str, state: State, config: &Co
             painter.plus_lines.push(prepare(&line, true, config));
             State::HunkPlus
         }
-        _ => {
-            // First character at this point is typically a space, but could also be e.g. '\'
-            // from '\ No newline at end of file'.
+        Some(' ') => {
             let prefix = if line.is_empty() { "" } else { &line[..1] };
```
wchargin, line 320: This also fixed a panic on some lines starting with a multi-byte UTF-8
code unit, as in:

``````markdown
Note that if you cap access token lifetime to 30 seconds—

```diff
diff --git a/tensorboard/data/server/gcs/auth.rs b/tensorboard/data/server/gcs/auth.rs
--- a/tensorboard/data/server/gcs/auth.rs
+++ b/tensorboard/data/server/gcs/auth.rs
@@ -296,3 +296,3 @@ impl RefreshTokenCreds {
             access_token: res.access_token,
-            expires: Instant::now() + Duration::from_secs(res.expires_in),
+            expires: Instant::now() + Duration::from_secs(res.expires_in.min(30)),
         })
```

—then access tokens are still only refreshed *once* every 30 seconds,
even among concurrent loads.
``````

…which, before this commit, panicked as:

```
thread 'main' panicked at 'byte index 1 is not a char boundary; it is inside '—' (bytes 0..3) of `—then access tokens are still only refreshed *once* every 30 seconds,`', src/delta.rs:314:60
```

Since `line.chars().next() == Some(' ')` implies that `&line[..1]` is a
valid slice, the new code looks safer to me. Thanks!

PR #171, opened 2020-05-02
chore: Exclude unneeded files from crates.io

ignatenkobrain:
Nothing uses those files when using releases on crates.io, so no need to ship those.


dandavison:
Sorry this got ignored / never merged!

PR #181, opened 2020-05-06
Import `less --no-init` avoidance from `bat`

### LINKED ISSUE #88: Don't pass `--no-init` to less when `BAT_PAGER` is explicitly set
Hello. I'm running into some issues where `git log --graph` is no longer clearing the screen after I exit. I believe this is because `delta` is passing the `--no-init` argument (along with a few others) to less explicitly. See the screenshot below:

image

I already have my own set of less configuration options in `LESS='-R -F --mouse --wheel-lines=10'`, and it would be ideal if the tool could avoid overriding the user's explicit settings.

I noticed that a similar problem was recently addressed in bat: https://github.com/sharkdp/bat/issues/749, where their solution was to stop passing `--no-init` to recent enough versions of `less`; is it possible that `delta` is just not depending on a recent enough version of `bat`?

da-x:
This fixes #88.

### DIFF
--- a/src/bat/output.rs
+++ b/src/bat/output.rs
@@ -9,6 +9,7 @@ use std::process::{Child, Command, Stdio};
 use shell_words;
 
 use crate::env;
+use super::less::retrieve_less_version;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[allow(dead_code)]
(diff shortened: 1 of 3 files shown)


src/bat/output.rs
```diff
@@ -79,7 +80,27 @@ impl OutputType {
...
+                            None => {
+                                p.arg("--no-init");
+                            }
+                            Some(version)
+                                if (version < 530 || (cfg!(windows) && version < 558)) =>
+                            {
+                                p.arg("--no-init");
+                            }
+                            _ => {}
+                        }
```
dandavison, line 102: @lzybkr Would you be able to review this change from the Windows point-of-view? The code here is copied from bat. If Delta simply does not work correctly with anything other than your version of less then perhaps this version-checking doesn't make sense on Windows?

dandavison:
Thanks again @da-x. This is good to have, as `--no-init` (`-X`) is really something that should be up to the user without delta mysteriously overriding it. #185 is an issue raised by a Delta user who encountered the old behavior.

I think this is the right thing to do on Windows since bat and delta users are advised to use recent versions of less on Windows which have @lzybkr's patches.

cohml:
Sorry to bring my personal problems here, but I simply cannot crack this.

When using `less` with `delta`, no matter how I configure it (e.g., via `.gitconfig`, via `DELTA_PAGER`), `less` does not respect `--no-init`. Meanwhile, `--no-init` works perfectly when using `less` with `bat`. I also see that when I add arguments for `less` via `.gitconfig` or `DELTA_PAGER`, they are respected (e.g., `-N`); only `--no-init` doesn't seem to work.

As a potentially relevant datapoint, when I do `LESS= PAGER= \less some_long_file.txt`, the screen is cleared upon exiting `less`, despite that I'm not obviously passing that in anywhere.

Anyway, can you provide any thoughts as to why `--no-init` might not be working with `delta`?

Issue #130, opened 2020-04-27
Would it be possible to add a line number sidebar?


eth-p:
Thanks for the amazing work on delta!

Do you think it would be possible to add an optional line number sidebar similar to the one that `bat` has? If `delta` had something like that, I think it could really help when there are large multi-line hunks where it's impractical to count past the first twenty or so lines.

dandavison:
Hi @eth-p, thanks for your work on bat! Yes, this makes sense and is doable and there have been a few requests for it.

There are a couple of design decisions I see: in delta's case (unlike bat) we have line numbers in the previous version and the new version. So, do we emulate (e.g.) GitLab and GitHub's two-column line number display, and is there an argument for a single-column line number display showing line numbers in the new version only, and what is the default going to be.

(When implementing, we need to be careful with the terminal-width accounting; people's customized `$LESS` settings easily cause the right-padding to overflow.)

eth-p:
> There are a couple of design decisions I see: in delta's case (unlike bat) we have line numbers in the previous version and the new version. So, do we emulate (e.g.) GitLab and GitHub's two-column line number display, and is there an argument for a single-column line number display showing line numbers in the new version only, and what is the default going to be.

That's definitely a pretty tough decision to make. Personally, I would be find a two column display like this to work for me (and line numbers could be dimmed where aren't any changes):

```
120 ⋮ 120 │ This line is the same.
    ⋮ 121 │ This line was added.
121 ⋮ 122 │ This line is the same.
122 ⋮ 123 │ This line was modified.
123 ⋮     │ This line was removed.
```

I'm not sure how you have printing implemented with `delta`, but that approach might be a bit difficult or space-consuming unless you knew the maximum line number before printing the sidebar (for example, in `bat` we use a fixed width and let the formatting break between line 9999 and 10000).

> (When implementing, we need to be careful with the terminal-width accounting; people's customized `$LESS` settings easily cause the right-padding to overflow.)

If you don't mind leaving it up to the user, you could change `--width=[number]` to accept offsets in the form of `+[number]`/`-[number]`, which would be used to set the width to `width + [offset]`. It would require people with customized `less` settings to do a little bit more work, but it also provides a way to let them to keep using their custom settings without breaking `delta`'s output.

dandavison:
> Personally, I would find a two column display like this to work for me

Thanks, that looks like a very reasonable design.

> that approach might be a bit difficult or space-consuming unless you knew the maximum line number before printing the sidebar 

[...]

Issue #189, opened 2020-05-13
🚀 Highlight merge diffs and conflict markers


rygwdn:
When showing a patch for a merge with `--cc` or `-c`, git will generate patch prefixes that are two characters instead of just one, (e.g. `++` or ` -`). Delta will highlight any lines that have a leading `+` or `-` but will not highlight lines with a leading space. I believe Delta should be highlighting the lines the same way git does (e.g. by ignoring the leading space so a ` -` line is red and a ` +` line is green).

Sample input from `git show --cc 27d2427484c6efe283671300b48b5aadc7549c8a` in this repo:
```
commit 27d2427484c6efe283671300b48b5aadc7549c8a
Merge: 55703ed 268887c
Author: Dan Davison <dandavison7@gmail.com>
Date:   Thu Apr 16 12:11:21 2020 -0400

    Merge branch 'master' into width-calc-fix

diff --cc Cargo.lock
index 9bd779d,5be85a6..cee10bb
--- a/Cargo.lock
+++ b/Cargo.lock
@@@ -293,11 -358,15 +358,16 @@@ dependencies =
   "regex 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
   "shell-words 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
   "structopt 0.2.18 (registry+https://github.com/rust-lang/crates.io-index)",
-  "syntect 3.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+  "syntect 4.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
   "unicode-segmentation 1.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
 + "unicode-width 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
  ]

+ [[package]]
+ name = "glob"
+ version = "0.3.0"
+ source = "registry+https://github.com/rust-lang/crates.io-index"
+
  [[package]]
  name = "heck"
  version = "0.3.1"
diff --cc Cargo.toml
index 914197f,cde6620..8e33592
--- a/Cargo.toml
+++ b/Cargo.toml
@@@ -27,9 -27,8 +27,9 @@@ lazy_static = "1.4
  regex = "1.2.1"
  shell-words = "0.1.0"
  structopt = "0.2.18"
- syntect = "3.2.0"
+ syntect = "4.1.0"
  unicode-segmentation = "1.3.0"
 +unicode-width = "0.1.6"

[...]

PR #190, opened 2020-05-14
Add option to show line numbers

clnoll:
Adds a CLI option, `--number`, that shows line numbers for the input and output diffs. The first column shows the line number for the original file, and the second column shows line number for the output file. A blank cell indicates that the line does not exist in one of the files (was removed or added, respectively).

Also provides the ability to customize the appearance of line numbers. Users can pass in a format string showing the position of the line numbers within the column, and can apply style elements to the numbers as well as the format string text, using new command line options:
- `--number-minus-style`
- `--number-plus-style`
- `--number-minus-format`
- `--number-plus-format`
- `--number-minus-format-style`
- `--number-plus-format-style`

image

Addresses https://github.com/dandavison/delta/issues/130.

TODO: tests.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -127,6 +127,24 @@ within a style string):
    Specifying colors like this is useful if your terminal only supports 256 colors (i.e. doesn\'t
    support 24-bit color).
 
+LINE NUMBERS
+------------
+
+Options that have a name like --*-format allow you to specify a string to display for the line
+number columns. The string should specify the location of the line number using the placeholder
+%ln.
+
+For example, to display the line numbers divided by specific characters:
+
+    8 ⋮   9 │ Here is an output line
+    9 ⋮  10 │ Here is another output line
+   10 ⋮  11 │ Here is the line number
+
+you would use the following input:
+
+--number-minus-format '%ln ⋮'
+--number-plus-format '%ln │'
+
 If something isn't working correctly, or you have a feature request, please open an issue at
 https://github.com/dandavison/delta/issues.
 "
@@ -244,6 +262,52 @@ pub struct Opt {
     /// section. One of the special attributes 'box', 'ul', 'overline', or 'underoverline' must be
     /// given.
     pub hunk_header_decoration_style: String,
+
+    /// Display line numbers next to the diff. The first column contains line
+    /// numbers in the previous version of the file, and the second column contains
+    /// line number in the new version of the file. A blank cell in the first or
+    /// second column indicates that the line does not exist in that file (it was
+    /// added or removed, respectively).
+    #[structopt(short = "n", long = "number")]
+    pub show_line_numbers: bool,
+
+    /// Style (foreground, background, attributes) for the left (minus) column of line numbers
+    /// (--number), if --number is set. See STYLES section.
+    /// Defaults to --hunk-style.
+    #[structopt(long = "number-minus-style", default_value = "auto")]
+    pub number_minus_style: String,
+
+    /// Style (foreground, background, attributes) for the right (plus) column of line numbers
+    /// (--number), if --number is set. See STYLES section.
+    /// Defaults to --hunk-style.
+    #[structopt(long = "number-plus-style", default_value = "auto")]
+    pub number_plus_style: String,
+
+    /// Format string for the left (minus) column of line numbers (--number), if --number is set.
+    /// Should include the placeholder %ln to indicate the position of the line number.
+    /// See the LINE NUMBERS section.
+    /// Defaults to '%ln⋮'
+    #[structopt(long = "number-minus-format", default_value = "%ln⋮")]
+    pub number_minus_format: String,
+
+    /// Format string for the right (plus) column of line numbers (--number), if --number is set.
+    /// Should include the placeholder %ln to indicate the position of the line number.
+    /// See the LINE NUMBERS section.
+    /// Defaults to '%ln│ '
+    #[structopt(long = "number-plus-format", default_value = "%ln│ ")]
+    pub number_plus_format: String,
+
+    /// Style (foreground, background, attributes) for the left (minus) line number format string
+    /// (--number), if --number is set. See STYLES section.
+    /// Defaults to --hunk-style.
+    #[structopt(long = "number-minus-format-style", default_value = "auto")]
+    pub number_minus_format_style: String,
+
+    /// Style (foreground, background, attributes) for the right (plus) line number format string
+    /// (--number), if --number is set. See STYLES section.
+    /// Defaults to --hunk-style.
+    #[structopt(long = "number-plus-format-style", default_value = "auto")]
+    pub number_plus_format_style: String,
 
     #[structopt(long = "color-only")]
     /// Do not alter the input in any way other than applying colors. Equivalent to
--- a/src/config.rs
+++ b/src/config.rs
@@ -41,12 +41,19 @@ pub struct Config<'a> {
     pub navigate: bool,
     pub null_style: Style,
     pub null_syntect_style: SyntectStyle,
+    pub number_minus_format: String,
+    pub number_minus_format_style: Style,
+    pub number_minus_style: Style,
+    pub number_plus_format: String,
+    pub number_plus_format_style: Style,
+    pub number_plus_style: Style,
     pub paging_mode: PagingMode,
     pub plus_emph_style: Style,
     pub plus_file: Option<PathBuf>,
     pub plus_line_marker: &'a str,
     pub plus_non_emph_style: Style,
     pub plus_style: Style,
+    pub show_line_numbers: bool,
     pub syntax_dummy_theme: SyntaxTheme,
     pub syntax_set: SyntaxSet,
     pub syntax_theme: Option<SyntaxTheme>,
@@ -120,6 +127,17 @@ pub fn get_config<'a>(
     let (commit_style, file_style, hunk_header_style) =
         make_commit_file_hunk_header_styles(&opt, true_color);
 
+    let (
+        number_minus_format_style,
+        number_minus_style,
+        number_plus_format_style,
+        number_plus_style,
+    ) = make_line_number_styles(
+        &opt,
+        hunk_header_style.decoration_ansi_term_style(),
+        true_color,
+    );
+
     let syntax_theme = if syntax_theme::is_no_syntax_highlighting_theme_name(&syntax_theme_name) {
         None
     } else {
@@ -164,12 +182,19 @@ pub fn get_config<'a>(
         navigate: opt.navigate,
         null_style: Style::new(),
         null_syntect_style: SyntectStyle::default(),
+        number_minus_format: opt.number_minus_format,
+        number_minus_format_style: number_minus_format_style,
+        number_minus_style: number_minus_style,
+        number_plus_format: opt.number_plus_format,
+        number_plus_format_style: number_plus_format_style,
+        number_plus_style: number_plus_style,
         paging_mode,
         plus_emph_style,
         plus_file: opt.plus_file.map(|s| s.clone()),
         plus_line_marker,
         plus_non_emph_style,
         plus_style,
+        show_line_numbers: opt.show_line_numbers,
         syntax_dummy_theme,
         syntax_set,
         syntax_theme,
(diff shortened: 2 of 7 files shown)


dandavison:
Thanks very much for this @clnoll! This is a substantial contribution and it's a feature that's been requested multiple times. I'll start reviewing.

src/paint.rs
```diff
@@ -144,6 +174,10 @@ impl<'a> Painter<'a> {
                 {
                     line.truncate(line.len() - ANSI_SGR_RESET.len());
                 }
+
+                if config.show_line_numbers {
+                    output_buffer.push_str(&line_numbers);
+                }
```
dandavison, line 180: I have a refactoring suggestion. What do you think about deferring the formatting of the line numbers until here? You could move `prepare_line_numbers` into this module (could we call it `format_line_numbers`?), and change the code in `delta.rs` so that it accumulates the line numbers as (optional) ints rather than as formatted text.

src/delta.rs
```diff
@@ -339,6 +365,7 @@ fn handle_hunk_line(
             Painter::paint_lines(
                 vec![syntax_style_sections],
                 vec![vec![(style::NO_BACKGROUND_COLOR_STYLE_MODIFIER, &line)]],
+                line_numbers,
```
dandavison, line 368: How about constructing this argument using the `vec!` macro rather than creating the Vec and then pushing to it?

src/delta.rs
```diff
@@ -314,22 +322,40 @@ fn handle_hunk_line(
...
+            ));
             State::HunkMinus
         }
         Some('+') => {
             painter.plus_lines.push(prepare(&line, true, config));
+            line_numbers.push(prepare_line_numbers(
+                None,
+                Some(painter.line_number_out),
+                config,
+            ));
```
dandavison, line 345: Are these ones (in the `-` and `+` branches) necessary?

src/paint.rs
```diff
@@ -24,6 +25,8 @@ pub struct Painter<'a> {
     pub highlighter: HighlightLines<'a>,
     pub config: &'a config::Config<'a>,
     pub output_buffer: String,
+    pub line_number_in: usize,
+    pub line_number_out: usize,
```
dandavison, line 29: To be consistent with other variables names, would it make sense for these to be named `minus_line_number` and `plus_line_number`?

src/delta.rs
```diff
@@ -382,6 +409,28 @@ fn prepare(line: &str, append_newline: bool, config: &Config) -> String {
...
+    let formatted = format!(
+        "{}⋮{}",
+        _line_number_style(line_number_in),
+        _line_number_style(line_number_out),
+    );
+    let numbers = paint::paint_text_foreground(&formatted, config.hunk_color, config.true_color);
+    let divider = paint::paint_text_foreground("│", config.hunk_color, config.true_color);
+    format!("{}{}", numbers, divider)
+}
+
```
dandavison, line 433: Playing around with the vertical dividing line character. E.g. https://jrgraphix.net/r/Unicode/2500-257F https://unicode.org/charts/nameslist/n_2500.html. Any thoughts? (I haven't found one yet which yields a vertical dashed line without "breaks" between lines.)

**Current:**

**Alternatives:**

`U2506`

`U250A`

src/delta.rs
```diff
@@ -382,6 +409,28 @@ fn prepare(line: &str, append_newline: bool, config: &Config) -> String {
     }
 }
 
+fn _line_number_style(line_number: Option<usize>) -> String {
+    match line_number {
+        Some(x) => format!("{:^4}", x),
+        None => format!("    "),
```
rygwdn, line 415: does this handle line numbers over 4 digits without breaking alignment? Maybe if it were deferred to `paint_lines` as suggested below, it would be easier to ensure that longer numbers are handled since you'd have access to all of them.

clnoll:
Thanks @dandavison & @rygwdn! Great suggestions, making these changes.

clnoll:
@dandavison @rygwdn I took all of the suggestions mentioned above, and added some additional command line options (colors for the line numbers and the dividers, and the ability to configure a string for the dividers).

FYI I'm seeing that it's a bit slower than `master` when benchmarking, so will take a closer look to see what I can do to speed it up.

Master:
```
Time (mean ± σ):      1.496 s ±  0.060 s    [User: 1.494 s, System: 0.042 s]
Range (min … max):    1.460 s …  1.661 s    10 runs
```

Feature branch, with `--numbers`
```
Time (mean ± σ):      1.807 s ±  0.078 s    [User: 1.786 s, System: 0.044 s]
Range (min … max):    1.757 s …  2.019 s    10 runs
```

Feature branch, without `--numbers`
```
Time (mean ± σ):      1.722 s ±  0.017 s    [User: 1.720 s, System: 0.043 s]
Range (min … max):    1.696 s …  1.748 s    10 runs
```

src/cli.rs
```diff
@@ -210,6 +210,42 @@ pub struct Opt {
...
+    #[structopt(long = "number-divider-center-foreground-color")]
+    pub number_divider_center_foreground_color: Option<String>,
+
+    /// String to use as the center divider of the line numbers section, if --number is set.
+    #[structopt(long = "number-divider-center-string", default_value = "⋮")]
+    pub number_divider_center_string: String,
+
+    /// String to use as the right divider of the line numbers section, if --number is set.
+    #[structopt(long = "number-divider-right-string", default_value = "│")]
+    pub number_divider_right_string: String,
```
dandavison, line 248: WDYT about changing these to `--number-right-divider-*` and `--number-center-divider-*`? (My brain parses "number-divider-right-foreground-color" as implying there's a right foreground.)

README.md
```diff
@@ -219,6 +219,7 @@ FLAGS:
                                      The default behavior is to output a space character in place of these markers.
         --light                      Use default colors appropriate for a light terminal background. For more control,
                                      see the other color options.
+        --line-numbers               Show line numbers for each line of the diff hunk.
```
dandavison, line 222: This is out of date now. (I do this step with `delta --help >> README.md` and then moving the appended text into position and fiddling around a bit.)

src/config.rs
```diff
@@ -153,6 +160,33 @@ pub fn get_config<'a>(
...
+        number_divider_center_string: opt.number_divider_center_string.as_ref(),
+        number_divider_right_string: opt.number_divider_right_string.as_ref(),
+    }
+}
+
+fn get_color_option_or_default(color_option: Option<&String>, default: &String) -> Color {
+    let _number_color = color_option.clone();
+    match color_option {
+        Some(x) => return color_from_rgb_or_ansi_code(&x),
+        None => return color_from_rgb_or_ansi_code(default),
```
dandavison, line 189: Can we use `color_from_rgb_or_ansi_code_with_default()` here?

src/cli.rs
```diff
@@ -210,6 +210,42 @@ pub struct Opt {
...
+    /// Display line numbers next to the diff. The first column contains line
+    /// numbers in the previous version of the file, and the second column contains
+    /// line number in the new version of the file. A blank cell in the first or
+    /// second column indicates that the line does not exist in that file (it was
+    /// added or removed, respectively).
+    #[structopt(short = "n", long = "number")]
+    pub show_line_numbers: bool,
+
+    /// Color for the left (minus) column of line numbers (--number), if --number is set.
+    /// Defaults to the configured hunk_color.
```
dandavison, line 223: Let's phrase this as
> Defaults to --hunk-color.

src/cli.rs
clnoll: Np, done! Also updated variable names to match.

README.md
clnoll: 👍 done.

src/config.rs
clnoll: Yep!! Updated.

src/cli.rs
clnoll: Updated.

src/delta.rs
```diff
@@ -246,7 +246,10 @@ fn handle_hunk_meta_line(
         cli::SectionStyle::Plain => panic!(),
         cli::SectionStyle::Omit => return Ok(()),
     };
-    let (raw_code_fragment, line_number) = parse::parse_hunk_metadata(&line);
+    let (raw_code_fragment, line_number_in, line_number_out) = parse::parse_hunk_metadata(&line);
+    painter.line_number_in = line_number_in.parse::<usize>().unwrap_or(0);
+    painter.line_number_out = line_number_out.parse::<usize>().unwrap_or(0);
```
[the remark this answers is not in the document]
clnoll: It was failing on new files (new line number would be 0). But at any rate as you've seen, I updated `parse_hunk_metadata` to handle this case.

src/delta.rs
clnoll: Done

src/delta.rs
clnoll: Nope! Removed.

src/paint.rs
clnoll: Yep.

src/delta.rs
clnoll: Added an option to input a divider (`--number-center-divider-string`).

src/delta.rs
clnoll: 👍 deferred to `paint_lines`. Need to take a look at how it handles >4 digits.

src/parse.rs
```diff
@@ -61,17 +62,31 @@ pub fn get_file_change_description_from_file_paths(
...
-    let code_fragment = iter.next().unwrap_or("");
-    (code_fragment, line_number)
+/// Return " pub fn delta(" and a vector of line numbers
+pub fn parse_hunk_metadata(line: &str) -> (&str, Vec<usize>) {
+    let pattern = r"@+ (?P<lns>([-+]\d+(?:,\d+)? ){2,4})@+(?P<cf>.*\s?)";
+    let re = Regex::new(&pattern).unwrap();
+    let caps = re.captures(line).unwrap();
+    let line_numbers = _make_line_number_vec(caps.name("lns").unwrap().as_str());
+    let code_fragment = caps.name("cf").unwrap().as_str();
+    return (code_fragment, line_numbers);
```
dandavison, line 89: I just learned that building regular expressions like this inside a function called many times is extremely inefficient. This commit on master speeds delta up by almost 2x 2cb3e403e802250308ef05e37e33105ad68eaef6! I suspect if you do the same thing here it will get rid of the performance difference between this branch and master.

In addition to `make benchmark` I've also added `make flamegraph`, for profiling on MacOS, which is sort of fun. (The Makefile assumes that you've cloned https://github.com/brendangregg/FlameGraph and symlinked two of its perl scripts  into your $PATH with the names `flamegraph` and `stackcollapse-sample`.)

src/parse.rs
```diff
@@ -61,17 +62,31 @@ pub fn get_file_change_description_from_file_paths(
...
+    for s in re.split(line) {
+        let number = s.split(',').nth(0).unwrap().split_whitespace().nth(0);
+        match number {
+            Some(number) => numbers.push(number.parse::<usize>().unwrap()),
+            None => continue,
+        }
+    }
+    return numbers;
+}
+
```
dandavison, line 79: Do you think it's possible to do it all with a regular expression, i.e. capture the numbers as well rather than using this function? While we're at it, we could also capture and expose the first set of `@` markers seeing as they will be needed for determining the number of merge parents for #189 cc @rygwdn

dandavison:
I've been doing a lot of refactoring in parallel and so if it's OK with you I'd like to treat `style-strings` as the base branch for this branch to target. Seeing as you've already had to rebase once, I rebased this branch on top of `style-strings` as `line-numbers-rebased`. If that diff LGTY then please feel free to take it (the commit has your attribution as author email).

You'll note that it...doesn't actually compile :) But it's fairly superficial -- we need to re-express the color options in this branch as what are referred to as "styles" in `style-strings`. The function `paint_text_foreground` has gone away and one just calls `style.paint(...)` instead.

src/cli.rs
```diff
@@ -210,6 +210,42 @@ pub struct Opt {
...
+
+    /// Color for the left (minus) column of line numbers (--number), if --number is set.
+    /// Defaults to --hunk-color.
+    #[structopt(long = "number-minus-foreground-color")]
+    pub number_minus_foreground_color: Option<String>,
+
+    /// Color for the right (plus) column of line numbers (--number), if --number is set.
+    /// Defaults to --hunk-color.
+    #[structopt(long = "number-plus-foreground-color")]
+    pub number_plus_foreground_color: Option<String>,
```
dandavison, line 230: In the new world of the `style-strings` branch I think we're going to want these options to be named `--number-minus-style` and `--number-plus-style`. (It might help to take a look at the new style documentation in `delta --help`.)

src/cli.rs
```diff
@@ -210,6 +210,42 @@ pub struct Opt {
...
+    #[structopt(long = "number-center-divider-foreground-color")]
+    pub number_center_divider_foreground_color: Option<String>,
+
+    /// String to use as the center divider of the line numbers section, if --number is set.
+    #[structopt(long = "number-center-divider-string", default_value = "⋮")]
+    pub number_center_divider_string: String,
+
+    /// String to use as the right divider of the line numbers section, if --number is set.
+    #[structopt(long = "number-right-divider-string", default_value = "│")]
+    pub number_right_divider_string: String,
```
dandavison, line 248: I think I'd like to investigate offering users a format string, rather than specifying the divider characters individually. So basically something with placeholder slots for the two line numbers. It's hard to make future-proof decisions about the CLI: I wonder whether that should be called something like `--hunk-line-format` in order to retain the possibility that in the future it might have other placeholder slots?

dandavison:
> I've been doing a lot of refactoring in parallel and so if it's OK with you I'd like to treat `style-strings` as the base branch for this branch to target

The style-string refactoring is all in `master` now.

src/parse.rs
```diff
@@ -249,9 +270,42 @@ mod tests {
...
+    #[test]
+    fn test_parse_hunk_metadata_merge() {
+        let parsed = parse_hunk_metadata("@@@ -293,11 -358,15 +358,16 @@@ dependencies =");
+        let code_fragment = parsed.0;
+        let line_numbers = parsed.1;
+        assert_eq!(code_fragment, " dependencies =",);
+        assert_eq!(line_numbers[0], 293,);
+        assert_eq!(line_numbers[1], 358,);
+        assert_eq!(line_numbers[2], 358,);
```
dandavison, line 309: Nice! Looks like this will help with highlighting merge diffs in #189.

PR #208, opened 2020-05-29
README: make comparison table more compact

waldyrious:
Leverage some markdown features to make the comparison table less unwieldy :)


dandavison:
Awesome, TIL! Thanks very much @waldyrious.

PR #225, opened 2020-06-17
Drop leading space, highlight empty lines, highlight whitespace errors

### LINKED ISSUE #179: 🚀 empty line handling and getting rid of the space leftover from -/+
Hi, I use delta as part of git in the terminal. I find it really useful to be able to copy chunks of deleted or added code and paste them in my IDE. Problem is there is always one space before each row. So I have to copy, paste and then remove that space for each line. Is it possible to get rid of this one-character column altogether in order to streamline the copy&paste? As default behavior or as an option?

dandavison:
Fixes #179 cc @phillipwood

- Unless `--keep-plus-minus-markers` is in effect, drop the first column so that the real code starts in the first column.

- Highlight addition of trailing whitespace as a whitespace error (i.e. git's `blank-at-eol`). Style determined by new option `--whitespace-error-style` which defaults to git's `color.diff.whitespace`.

- Highlight removals and additions of empty lines with background color in the first column if these would not otherwise be visible (i.e. if the applicable style lacks a background color).

- Additions of empty lines at end-of-file are _not_ currently highlighted as whitespace errors, but rather as normal addition-of-empty line. This is contra git's `blank-at-eof` and is a bug to be addressed in subsequent work.

See git documentation: https://git-scm.com/docs/git-config#Documentation/git-config.txt-corewhitespace

### DIFF
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -14,6 +14,8 @@ use crate::paint::superimpose_style_sections::superimpose_style_sections;
 use crate::style::Style;
 
 pub const ANSI_CSI_CLEAR_TO_EOL: &str = "\x1b[0K";
+pub const ANSI_CSI_CLEAR_TO_BOL: &str = "\x1b[1K";
+pub const ANSI_CSI_CURSOR_BACK_1: &str = "\x1b[1D";
 pub const ANSI_SGR_RESET: &str = "\x1b[0m";
 
 pub struct Painter<'a> {
@@ -103,6 +105,7 @@ impl<'a> Painter<'a> {
                 },
                 self.config.minus_style,
                 self.config.minus_non_emph_style,
+                Some(self.config.minus_empty_line_marker_style),
                 None,
             );
         }
@@ -120,6 +123,7 @@ impl<'a> Painter<'a> {
                 },
                 self.config.plus_style,
                 self.config.plus_non_emph_style,
+                Some(self.config.plus_empty_line_marker_style),
                 None,
             );
         }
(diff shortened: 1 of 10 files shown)


src/paint.rs
```diff
@@ -227,6 +235,18 @@ impl<'a> Painter<'a> {
...
+            } else if line.is_empty() {
+                if let Some(empty_line_style) = empty_line_style {
+                    output_buffer.push_str(
+                        &empty_line_style
+                            .ansi_term_style
+                            .paint(format!(
+                                "{}{}",
+                                ANSI_CSI_CLEAR_TO_BOL, ANSI_CSI_CURSOR_BACK_1
+                            ))
+                            .to_string(),
```
dandavison, line 247: @phillipwood I've rebased over the commit on which you commented earlier but the call to `ansi_term_style.paint()` here wraps `\033[1K\033[1D` with a prefix which sets the background color and a `SGR RESET` suffix. Therefore I believe the end result is to emit the pattern of ANSI sequences that you suggested (modulo the details of the background color):

```
'\033[41m\033[1K\033[1D\033[m\n'
```

But do let me know if something doesn't look right!

phillipwood: Ah that explains the reset. However I don't think you need `ANSI_CSI_CURSOR_BACK_1` though. Looking back at my comments on the issue I must have copied and pasted the wrong line from where I had been playing around in the terminal. It should have been `<set background color><ANSI_CLEAR_TO_BOL><ANSI_SGR_RESET><\n>` The `ANSI_CSI_CUROR_BACK_1` was left over from something else.  Sorry for the confusion

dandavison: Right, thanks for that. I have removed the `ANSI_CSI_CURSOR_BACK_1`.

PR #235, opened 2020-07-04
Update README.md with Windows Scoop installation

daanzu:


dandavison:
Thanks very much @daanzu!

PR #241, opened 2020-07-08
Readme: do not use quotes in .gitconfig

yparitcher:
quoting arguments in the [delta] section (aside from color hex codes which need to be quoted) will pass the quote to delta which will not match the argument correctly (for example a theme with spaces)


dandavison:
Thank you @yparitcher!

dandavison:
Thank you @yparitcher!

PR #242, opened 2020-07-09
Use `syntax-theme` not `theme` in gitconfig example

benjamineskola:
`theme` is deprecated so it doesn't make sense to recommend it. (Actually when I tried it, copying from the example, it didn't work at all.)


dandavison:
Thanks  very much @benjamineskola.

> when I tried it, copying from the example, it didn't work at all.

Thanks, I can reproduce that.

dandavison:
> when I tried it, copying from the example, it didn't work at all.

Fixed in c0bcc9d641de2b0fba9ae1857ea7c142128e5a3d

PR #249, opened 2020-07-13
In README.md, fix incorrect theme listing flag

aeriksson:
So the flag `--list-theme-names` doesn't seem to exist, but `--list-syntax-themes` does (and seems to do what the readme says `--list-theme-names` should).


dandavison:
Thanks @aeriksson!

PR #254, opened 2020-07-15
Fix #138: Add option to hide hunk headers

### LINKED ISSUE #138: Add option to not show hunk headers
I think it would be a good idea to be able to disable the hunk highlighting in certain cases.
In javascript for example, some code is outside of the hunk that's highlighted, and this can get a bit confusing.

For example with the following code
```js
function test() {
    console.log("go go go");
}


test();
console.log("this isn't in the hunk!")
```

when adding the last line, the git diff detects the hunk for `function test()` and this can be a bit misleading as all changes are outside of hunks.

git diff result:
ss

I think it would be best if i could disable hunks for a specific language (in this case anything that can have code outside of a hunk.

ryuta69:
Fix #138

## What this PR does

if you set option `hunk-header-hide`, delta won't output hunk header.

ss 8

## Why make this PR

There is some needs to hide hunk header. It is because hunk header is not perfect due to git problem as https://github.com/dandavison/delta/issues/138#issuecomment-620591554 mentioned.

## Remarks

First of all I thought I should make option like `hunk-header-style=none`, however, hunk-header-style is all about theme. So it shouldn't control whether displays hunk headers or not.


dandavison:
Hi @ryuta69, this is great, however, this can already be done with `hunk-header-style = omit`. Since delta has many options already, I do like expressing this as part of the style string language. It works for `commit-style`, `file-style`, and `hunk-header-style`. I'm sorry that wasn't documented, I've fixed that now.

ryuta69:
@dandavison

Wow, sorry I didn't notice that. Thank you for updating document, I'm gonna close this PR.

By the way, shouldn't issue #138 be closed as well?

dandavison:
> By the way, shouldn't issue #138 be closed as well?

Thanks! You're right, I'd forgotten about that one.

PR #261, opened 2020-07-22
Add OSC hyperlinks

### LINKED ISSUE #257: 🚀 Support displaying files and commits as OSC 8 hyperlinks
[This proposal](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) introduces a spec for displaying hyperlinks in terminal emulators. Support for it among terminal emulators is already good, with more pending. Unfortunately neither [less -R](https://github.com/gwsw/less/issues/54) nor [tmux](https://github.com/tmux/tmux/issues/911) support it yet, so for now Delta users would only be able to benefit by (a) using a fork of `less` (and tmux), or (b) not using `less` (or tmux) with delta.

In any case, two ways in which delta could make use of this feature are

1. Format file names as clickable `file://` links.
2. Format commit hashes as links to an appropriate page in a git web-hosting site, if that is discernible from the git remote in the repo in which delta was invoked.

Both these would be optimistic: delta just receives text on stdin and there's no guarantee that the files in that text have ever existed anywhere! Also delta doesn't need to be invoked inside a git repo, and I think perhaps neither type of link-formatting  should be done when delta is not invoked inside a git repo.

dandavison:
Closes #257.

```
--hyperlinks   Render commit hashes, file names, and line numbers as hyperlinks, according to the
               hyperlink spec for terminal emulators:
               https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda. By default,
               file names and line numbers link to the local file using a file URL, whereas commit
               hashes link to the commit in GitHub, if the remote repository is hosted by GitHub.
               See --hyperlinks-file-link-format for full control over the file URLs emitted.
               Hyperlinks are supported by several common terminal emulators. However, they are
               not yet supported by less, so they will not work in delta unless you install a
               patched fork of less (see https://github.com/dandavison/less). If you use tmux,
               then you will also need a patched fork of tmux (see
               https://github.com/dandavison/tmux)
```

```
--hyperlinks-file-link-format <hyperlinks-file-link-format>
    Format string for file hyperlinks. The placeholders "{path}" and "{line}" will be replaced by the absolute
    file path and the line number, respectively. The default value of this option creates hyperlinks using
    standard file URLs; your operating system should open these in the application registered for that file
    type. However, these do not make use of the line number. In order for the link to open the file at the
    correct line number, you could use a custom URL format such as "file-line://{path}:{line_number}" and
    register an application to handle the custom "file-line" URL scheme by opening the file in your editor/IDE
    at the indicated line number. See https://github.com/dandavison/open-in-editor for an example [default:
    file://{path}]
```


dandavison:
cc @Kr1ss-XD, @navarroaxel new feature in master in case it's of interest. Locally, I have this so that command-click on any line number in delta opens my text editor on the file at that line number, but as you can see from the help text it takes a bit of set-up (OSC hyperlink terminal emulator codes not yet supported in less or tmux for starters.)

Kr1ss-XD:
That sounds ace !
I'll give it a shot this weekend, won't be able to get into it sooner, but thank you for the cc.

dandavison:
This is what it looks like in iTerm2 (iTerm2 renders the hyperlinks with a dashed underline):

PR #265, opened 2020-07-29
Fix width option in git_config honored

### LINKED ISSUE #264: 🐛 width option is not honored in git config sections
When side-by-side is on, it gets really wide.
It would be great if I could set max-column( sorry if there is already. I look for document and issues, but couldn't find.)

screenshot 0002-07-28 at 2 37 28 PM

ryuta69:
Fix #264

## What this PR does

This PR enables width option in git-config.
If width option in git-config is set, and width cli option is given, cli option value have priority.

ss 3

## Unittest

```
❯ cargo test test_width_in_git_config_is_honored

    Finished test [unoptimized + debuginfo] target(s) in 0.16s
     Running target/debug/deps/delta-356ae7bb950c7ce7

running 1 test
test options::set::tests::test_width_in_git_config_is_honored ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 165 filtered out
```


dandavison:
Thanks @ryuta69!

This looks like a great start. The main problem I'm seeing currently is that we need to handle `width` in a custom feature. So, this unit test should pass, but it currently fails:

```rust
    #[test]
    fn test_width_in_git_config_custom_feature_is_honored() {
        let git_config_contents = b"
[delta]
    features = my-width-feature

[delta \"my_width_feature\"]
    width = variable
";
        let git_config_path = "delta__test_width_in_git_config_custom_feature_is_honored.gitconfig";

        let opt = integration_test_utils::make_options_from_args_and_git_config(
            &[],
            Some(git_config_contents),
            Some(git_config_path),
        );

        assert_eq!(opt.computed.decorations_width, cli::Width::Variable);

        remove_file(git_config_path).unwrap();
    }
``` 

I _think_ what you'll need to do is allow the logic in `set_options!` to determine the user-supplied width value first, and then interpret that user-supplied width value afterwards. The `set_options!` macro is a bit complicated: it exists to do the logic of looking up values from the command line, from the `[delta]` section, and from custom and built-in features, in exactly the same way for all delta options.

src/options/set.rs
```diff
@@ -492,10 +492,13 @@ fn set_widths(opt: &mut cli::Opt) {
             });
             (cli::Width::Fixed(width), true)
         }
-        None => (
-            cli::Width::Fixed(opt.computed.available_terminal_width),
-            true,
-        ),
+        None => {
+            match git_config.as_ref().unwrap().get::<usize>("delta.width")
```
dandavison, line 496: I believe this `unwrap()` is causing lots of tests to fail.

src/options/set.rs
```diff
@@ -656,4 +660,23 @@ pub mod tests {
 
         remove_file(git_config_path).unwrap();
     }
+
+    #[test]
+    fn test_width_in_git_config_is_honored() {
+        let git_config_contents = b"
+[delta]
+    width = 7
```
dandavison, line 668: I think we should make this test value "variable". Otherwise there is a possibility that the test will be running in a terminal which really does have the width value 7!

src/config.rs
```diff
@@ -62,6 +62,7 @@ pub struct Config {
     pub syntax_dummy_theme: SyntaxTheme,
     pub syntax_set: SyntaxSet,
     pub syntax_theme: Option<SyntaxTheme>,
+    pub width: Option<String>,
```
dandavison, line 65: The `config` struct already has `available_terminal_width` and `decorations_width`. I feel like it should not need another field holding a width value.

ryuta69:
Oh, sorry for my lack of consideration. I'm gonna fix.

src/options/set.rs
ryuta69: Oh, sorry I hadn't run `make test`, and yes as you said, I find a lot of tests failed now. I'm totally sorry.

dandavison: No problem! When there's that many failures it's usually something easy to fix.

ryuta69:
@dandavison 

Hi, thank you very much for your review again.
I fixed those of your reviews.

### Test
```
❯ make test
cargo test                                                                              Compiling git-delta v0.3.1 (/Users/ryuta69/.ghq/github.com/ryuta69/delta)
    Finished test [unoptimized + debuginfo] target(s) in 4.73s
     Running target/debug/deps/delta-356ae7bb950c7ce7

----
# many tests
----

test result: ok. 161 passed; 0 failed; 5 ignored; 0 measured; 0 filtered out
```

It used to have failed many tests at first PR, sorry.

### Behaviour

I checked these cases working as intended.

1 - Without any config and options => opt.computed.available_terminal_width
2 - With config and options => option value
3 - With config without options => config value
4 - With config defined in custom features => config value inside custom features

src/config.rs
ryuta69: fixed here
313378a

src/options/set.rs
ryuta69: thanks to your comment https://github.com/dandavison/delta/pull/265#issuecomment-665825388 as well, fixed here
00f3899

src/main.rs
```diff
@@ -207,10 +206,10 @@ fn show_config(config: &config::Config) {
...
             .map(|t| t.name.unwrap_or("none".to_string()))
             .unwrap_or("none".to_string()),
-        width = config
-            .width
-            .clone()
-            .unwrap_or((Term::stdout().size().1 - 1).to_string()),
+        width = match config.decorations_width {
+            cli::Width::Fixed(width) => width.to_string(),
+            _ => config.available_terminal_width.to_string()
+        },
```
ryuta69, line 212: I thought `fn show_config` should show `available_terminal_width` if user doesn't assign any number to width, but if you think it is not as intended, please tell me.

dandavison: Right, I do think that if the user does e.g. `delta --width variable --show-config` then we need it to print out `width = variable` here, so that the output faithfully represents the active config values.

src/options/set.rs
```diff
@@ -479,9 +479,28 @@ fn parse_paging_mode(paging_mode_string: &str) -> PagingMode {
...
+    opt: &mut cli::Opt,
+    git_config: &mut Option<git_config::GitConfig>,
+    arg_matches: &clap::ArgMatches,
+    option_names: &HashMap<&str, &str>,
+) {
     // Allow one character in case e.g. `less --status-column` is in effect. See #41 and #10.
     opt.computed.available_terminal_width = (Term::stdout().size().1 - 1) as usize;
+
+    let empty_builtin_features = HashMap::new();
+    if opt.width.as_deref() == None {
```
dandavison, line 492: I believe this is better written as `opt.width.is_none()`

ryuta69: fixed here
904b59e

src/main.rs
ryuta69: fixed here
4284d6e
ss 5

ryuta69:
@dandavison 
Hi, thank you for pointing out, I now fixed two!

dandavison:
Perfect. Thanks very much for this work @ryuta69.

ryuta69:
I'm glad to contribute my favorite OSS!
Thank you for your reviews again!

Issue #258, opened 2020-07-19
🐛  Mismatch in line numbers between git diff and git add -p


davidmatas:
Hi,

I'm getting differences in lines number between `git diff` and `git add -p` execution. See the following captures:

`git add -p`
![image](https://user-images.githubusercontent.com/271760/87882015-69f82380-c9fd-11ea-9602-500aec51437e.png)

`git diff`
![image](https://user-images.githubusercontent.com/271760/87882024-7da38a00-c9fd-11ea-8d0e-7cbdc4a8f724.png)

The correct one is the one I'm getting in `git diff`.

Versions:

delta: `0.3.0`
git: `2.27.0`

yoichi:
The problem occurs when we have following setting in gitconfig:

```
[interactive]
	diffFilter = delta --color-only
[delta]
	line-numbers = true
```

More precisely, it is caused since handle_header_line is skipped at
https://github.com/dandavison/delta/blob/master/src/delta.rs#L116

On the other hand, if we call handle_header_line there even when raw/color-only option is set by changing code as follows,
then `git add -p` cause error.

![スクリーンショット 2020-07-22 5 31 40](https://user-images.githubusercontent.com/107537/88103954-b3eb3000-cbdc-11ea-8945-80bff90193da.png)

The error is caused since an empty line is added in its output for a hunk.

dandavison:
Hi @davidmatas, @yoichi, thanks very much for opening this, and for the investigation & analysis, and now a PR! (I'm sorry, I thought I'd commented already to acknowledge this helpful issue submission, but apparently my comment didn't get submitted and I didn't realize.)

PR #272, opened 2020-08-05
Calculate line numbers correctly in raw/color-only mode

yoichi:
fix https://github.com/dandavison/delta/issues/258

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -131,7 +131,7 @@ where
             painter.paint_buffered_minus_and_plus_lines();
             state = State::HunkHeader;
             painter.set_highlighter();
-            if config.line_numbers || should_handle(&state, config) {
+            if should_handle(&state, config) {
                 painter.emit()?;
                 handle_hunk_header_line(&mut painter, &line, &raw_line, &plus_file, config)?;
                 continue;
@@ -188,6 +188,9 @@ where
 
 /// Should a handle_* function be called on this element?
 fn should_handle(state: &State, config: &Config) -> bool {
+    if *state == State::HunkHeader && config.line_numbers {
+        return true;
+    }
     let style = config.get_style(state);
     !(style.is_raw && style.decoration_style == DecorationStyle::NoDecoration)
 }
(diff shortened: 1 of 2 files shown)


src/delta.rs
```diff
@@ -131,7 +131,7 @@ where
             painter.paint_buffered_minus_and_plus_lines();
             state = State::HunkHeader;
             painter.set_highlighter();
-            if should_handle(&state, config) {
+            if config.line_numbers || should_handle(&state, config) {
```
dandavison, line 134: Do you think it would make sense to incorporate the `if config.line_numbers` condition into `should_handle` so that all the logic is in one place?

dandavison:
Thanks very much for this PR and test @yoichi, and the initial analysis in #258. I just have that question about how you think the code should ideally be organized.

dandavison:
Thanks very much for this work @yoichi. And @alfredbez for testing and reviewing.

I added one failing test in master that will pass when this branch is merged: https://github.com/dandavison/delta/commit/0e980778538437968e242044fd125d58d672b50e (asserting that line numbers start at 1, not 0, under color-only)

dandavison:
Now let's figure out what to do about side-by-side and then I'll get these fixes released, since `add -p` is important.

src/delta.rs
```diff
@@ -413,7 +416,9 @@ fn handle_hunk_header_line(
     let (raw_code_fragment, line_numbers) = parse::parse_hunk_header(&line);
     // Emit the hunk header, with any requested decoration
     if config.hunk_header_style.is_raw {
-        writeln!(painter.writer)?;
+        if config.hunk_header_style.decoration_style != DecorationStyle::NoDecoration {
+            writeln!(painter.writer)?;
+        }
```
dandavison, line 421: @yoichi do you recall the logic here -- why do we want a newline only if `hunk_header_style.decoration_style != DecorationStyle::NoDecoration`? I'm sorry, I should have asked you to add a comment at the time. Given that `hunk-header-style=omit` is failing (#274), I'm thinking we need to do more work to characterize the conditions under which the newline and other alterations are permitted. If you have any thoughts then great, but if not don't worry!

dandavison: Thanks @yoichi! I wonder whether I should get rid of that extra newline character completely, at least in order to get the code into an easy-to-understand state. And then consider re-introducing it. I think I added it in the big v0.2 update because I thought that the extra space made the diff easier to read.

Issue #288, opened 2020-08-13
🚀  ignore big file diff


swuecho:
really like delta to show the diff but not usable when file too big.

when file is too big, delta will be no response.  wonder if there is a way to stop diff specific file.  

also, is there a way to tell git to stop using delta by some command line option while keep the git config unchanged? (I want to use delta by default, but when it does not work, I want to use the default git diff).

Thanks.

dandavison:
Hi @swuecho, thank you!

Can you tell me, which of these is closest to your problem?

- problem with individual long lines (like https://github.com/dandavison/delta/issues/154)
- problem with many lines displayed (like https://github.com/dandavison/delta/issues/156)

> is there a way to tell git to stop using delta by some command line option while keep the git config unchanged? 

Yes, there is. You can use the `git config --global` command. For example:

```sh
# Disable delta
git config --global --unset core.pager

# enable delta
git config --global core.pager delta
```

You can use that to set any git config option (git options and delta options): there are a few examples in the delta README.

Kr1ss-XD:
FWIW, `git config --global` _does_ modify the (global) git configuration. While it is easy to revert the effect (e.g. set up aliases for the two mentioned commands - `delta-on` and `delta-off` for example), you can also switch off paging for a single git call via command line option.

Just call git as `git -P` or `git --no-pager`. this won't alter your configuration.

dandavison:
Hi @Kr1ss-XD, yes that's true that it modifies our gitconfig files on disk. However, `git --no-pager` doesn't use a pager at all, so it is no good for large diffs or `git log -p`. So perhaps an even better alternative is

```
git -c 'core.pager=less -R' {show/diff/log -p etc}
```

But... `git -c ...` cannot be used to set delta options. For example, `git -c 'delta.side-by-side=true'` does not work at all, because delta only sees git config on disk.

So I think for that reason I am currently embracing the idea of using my gitconfig as something that can be transiently mutated (I keep it in git, of course, so I can see the changes), and using aliases based on `git config --global` to  alter git/delta options transiently.

Kr1ss-XD:
Agreed, those are good points.

swuecho:
> Hi @swuecho, thank you!
> 
> Can you tell me, which of these is closest to your problem?
> 
> * problem with individual long lines (like #154)
> * problem with many lines displayed (like #156)
> 
> > is there a way to tell git to stop using delta by some command line option while keep the git config unchanged?
> 
[...]

PR #290, opened 2020-08-14
Truncate long lines

### LINKED ISSUE #154: 🐛 High memory usage when fed diff of minimized files
Version: 0.1.1

Steps to reproduce:

```
wget https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/js/bootstrap.min.js -O /tmp/bs-4.4.1.js
wget https://stackpath.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js -O /tmp/bs-4.0.0.js
diff -u /tmp/bs-4.0.0.js /tmp/bs-4.4.1.js | delta
```

Example above seems to use around 8GB and takes 10 seconds to fully display the diff on my system.

dandavison:
Fixes #154, #288


gasche:
In my experience this default setting makes it inconvenient to use `delta` on text documents without hard-wrapping, as is a de-facto standard for say markdown documents and some LaTeX writing styles. This makes the `delta` default settings unsuitable for git repositories versioning such documents (rather than code).

For example, calling `delta --help | delta` results in truncated output:

```
Options --line-numbers-left-format and --line-numbers-right-format allow you to change the contents of the line number columns. Their values are arbitrary format strings, which are allowed to contain the placeholders {nm} for the line number associated with the old version of the file and {np} for the line number associated with the new version of the file. The placeholders support a subset of the string formatting syntax documented here: https://doc.rust-lang.org/std/fmt/#formatting-parameters. Specifical→
```

It is not a great for a pager to hide output by default.

I think that it would be better if:
- the default could be made much larger; `delta --help | wc -L` reports a maximum line length of 701, maybe a default maximum of say 2048 would be safe for most markdown documents?
- or if you found a different way to avoid computation-time blowup that does not require truncating line. Maybe the whatever processing that is done per-line could be disabled for long lines, but the line content would still be shown.

dandavison:
Hi @gasche, I haven't had too many complaints about this and it was done very early on in the project. The main thing that caused me to do it was minified css and js files on a single line -- delta attempted to syntax highlight them and you can imagine what the performance was like.

> This makes the delta default settings unsuitable for git repositories versioning such documents (rather than code).

You probably know this, but in such repos this problem can be fixed permanently by adding

```gitconfig
[delta]
    max-line-length = 0
```

to the repo-specific git config (`.git/config`)

> the default could be made much larger; delta --help | wc -L reports a maximum line length of 701, maybe a default maximum of say 2048 would be safe for most markdown documents?

It seems to me that if these are paragraphs of prose then we can't predict an upper limit

> Maybe the whatever processing that is done per-line could be disabled for long lines, but the line content would still be shown.

Possibly. But I think that I like diffs *not* showing entire minified js/css files, with or without syntax highlighting.

So, I'm not quite sold personally on the need to change. But I could definitely believe that we could do a better job of letting people working in repos of plain text prose  know that they should set `max-line-length = 0`.

gasche:
> It seems to me that if these are paragraphs of prose then we can't predict an upper limit

Non-hard-wrapped text typically uses one line per paragraph, and paragraph tend to not be super-long for readability reasons (and because block structure such as lists, code or math elements require breaking the text paragraph). For `delta --help`, the maximum line length is 701. In my own documents I have found 999 to be the maximum line length. Finally, for physical books the standard character-per-page counts are reported to be between 1800 and 3700 characters -- some books have paragraphs running on several pages, but this is highly unusual.

To me this suggests that bumping the limit to 1024 or 2048 would be a reasonable choice.

In contrast, searching the web for "typical" css file sizes suggests that people aim for CSS files between 10Kio and 25Kio, so the minified CSS one-line files would typically start at around 10,000 characters. (People are generally discouraged from committing their minified CSS files in the version control system, but oh well.) Javascript file sizes are reported to be much larger than that.

gasche:
(The `delta --help` output contains four paragraphs over the current 512 limit.)

injust:
I got bit by this -- for a few days, I thought there was a bug in `delta` because changed lines were being truncated, hiding the actual change in the line. This default was surprising because my primary use case of `delta` is to highlight the smaller changes within longer lines.

> In contrast, searching the web for "typical" css file sizes suggests that people aim for CSS files between 10Kio and 25Kio, so the minified CSS one-line files would typically start at around 10,000 characters.

+1, 500 characters seems too low for a default. Also, 8GB of minifed JS takes ~10 seconds to diff (from #154), so there should be room to increase the default before performance becomes a factor.

> Maybe the whatever processing that is done per-line could be disabled for long lines, but the line content would still be shown.

+1

> Possibly. But I think that I like diffs not showing entire minified js/css files, with or without syntax highlighting.

That purpose seems a bit orthogonal to `delta`'s design (but you'd know about that more than me 😉). `.gitignore` could do the same thing in a much cleaner way.

> You probably know this, but in such repos this problem can be fixed permanently by adding
> 
> ```gitconfig
> [delta]
>     max-line-length = 0
> ```
> 
> to the repo-specific git config (`.git/config`)

I think we should document this in the README, or even better, add the default to the config so it's obvious that this functionality is in play.

gasche:
@injust thanks for the extra feedback. Do you have estimates of maximal line lengths for the documents you happen to be working with? (Note: `wc -L` returns the maximal line length.)

injust:
For reference, this uBlock Origin backup has max line length of approx 12k. `time git show` takes about half a second. It's also a `.txt` file, so I don't think `delta` is trying to syntax highlight it?

th1000s:
When looking at delta performance (a while ago), syntax highlighting was the most expensive part.

It might make sense to just limit syntax highlighting and not line length (within reason). In #1746 these are now separated, e.g.

  `--max-syntax-highlighting-length=500 --max-line-length=5000 `

A more complex solution would give delta "syntax fuel", which accumulates over normal lines and and can be spent on one or two overly long lines, but once it runs into a large block only the first 500 characters are highlighted.

gasche:
Indeed, I think that #1746 would certainly satisfy my use-case. Thanks!

th1000s:
While `--max-syntax-highlighting-length` can speed up the output, there is a good chance that the following lines have incorrect highlighting until the next hunk resets the highlighter state machine, but that as also the case with `--max-line-length`.

PR #321, opened 2020-09-13
Fix a bug that `git add -p` doesn't work when some options are enabled

### LINKED ISSUE #320: 🐛 git add -p one-to-one correspondence errors
When I try to use `git add -p`, I get the following error:
```
$ git add -p
fatal: mismatched output from interactive.diffFilter
hint: Your filter must maintain a one-to-one correspondence
hint: between its input and output lines.
```

I'm using `delta` with the following config:
```
$ delta --version
delta 0.4.3
$ delta --show-config
    commit-style                  = raw
    file-style                    = blue
    hunk-header-style             = syntax
    minus-style                   = syntax 52
    minus-non-emph-style          = syntax 52
    minus-emph-style              = syntax 88
    minus-empty-line-marker-style = normal 88
    zero-style                    = syntax
    plus-style                    = syntax black
    plus-non-emph-style           = syntax black
    plus-emph-style               = syntax 22
    plus-empty-line-marker-style  = normal 22
    whitespace-error-style        = reverse magenta
    24-bit-color                  = false
    file-added-label              = 'added:'
    file-modified-label           = ''
    file-removed-label            = 'removed:'
    file-renamed-label            = 'renamed:'
    hyperlinks                    = false
    inspect-raw-lines             = true
    keep-plus-minus-markers       = true
    line-numbers                  = true
    line-numbers-minus-style      = red
    line-numbers-zero-style       = brightgreen
    line-numbers-plus-style       = green
    line-numbers-left-style       = blue
    line-numbers-right-style      = blue
    line-numbers-left-format      = '{nm:^4}⋮'
    line-numbers-right-format     = '{np:^4}│'
    max-line-distance             = 0.6
    max-line-length               = 512
    navigate                      = false
    paging                        = auto
    side-by-side                  = false
    syntax-theme                  = Solarized (dark)
    width                         = 145
    tabs                          = 4
    word-diff-regex               = '\w+'
```

In my .gitconfig are the following relevant options:
```
[core]
        pager = delta
[delta]
        commit-style = raw
        file-style = blue
        hunk-header-style = syntax
        minus-style = syntax 52
        minus-emph-style = syntax 88
        zero-style = syntax normal
        plus-style = syntax black
        plus-emph-style = syntax 22
        keep-plus-minus-markers = true
        line-numbers = true
        line-numbers-minus-style = red
        line-numbers-zero-style = brightgreen
        line-numbers-plus-style = green
        whitespace-error-style = reverse magenta
        minus-empty-line-marker-style = normal 88
        plus-empty-line-marker-style = normal 22
[interactive]
[color]
        ui = true
```

ryuta69:
Fix #320

As same as side-by-side did already, `file-style` and `hunk-header-style` also need to be fixed because it cause a break line that is not acceptable to `git add -p` input. (Ref https://github.com/dandavison/delta/pull/272#discussion_r475024167)

```
# Minimal reproduce

[core]
    pager = delta
[delta]
    file-style = "#338ECF" ul
    side-by-side = true
    hunk-header-style = omit
[interactive]

# Once you build, test with this command.
git -c 'interactive.diffFilter="{{delta repository path}}delta/target/release/delta" --color-only' add -p
```

I added a test, but wonder if this is accurate to what we need... Even though option is changed as we code, we can't tell that output is acceptable to `git add -p`. I searched what kind of format is acceptable to `git add -p`, but couldn't find.

Any idea is appreciated.


src/options/set.rs
```diff
@@ -81,10 +81,13 @@ pub fn set_options(
...
     let mut builtin_features = features::make_builtin_features();
 
-    // --color-only is used for interactive.diffFilter (git add -p) and side-by-side cannot be used
+    // --color-only is used for interactive.diffFilter (git add -p) and
+    // side-by-side, file_style, and hunk_header_style cannot be used
     // there (does not emit lines in 1-1 correspondence with raw git output). See #274.
     if config::user_supplied_option("color-only", arg_matches) {
         builtin_features.remove("side-by-side");
+        builtin_features.remove("file-style");
+        builtin_features.remove("hunk-header-style");
```
dandavison, line 90: Have you confirmed that these changes are necessary? I don't think they have any effect, since the keys in `builtin_features` hash map should just be the [builtin features](https://github.com/dandavison/delta/tree/master/src/features), i.e. `side-by-side`, `line-numbers`, `navigate`, `color-only`, `diff-so-fancy`, `diff-highlight`, `hyperlinks`.

dandavison:
Thanks very much for working on this @ryuta69!

> I added a test, but wonder if this is accurate to what we need... Even though option is changed as we code, we can't tell that output is acceptable to `git add -p`. I searched what kind of format is acceptable to `git add -p`, but couldn't find.

So, the error message says

> hint: Your filter must maintain a one-to-one correspondence between its input and output lines.

Could you add a test that is similar to the tests in `src/tests/test_example_diffs.rs`, and does the following:

- Sets up a config with `--color-only`
- Choose an example diff for inout
- Call `run_delta`  and get the output
- Assert that the number of lines of output equals the number of lines of input
- Assert that, after stripping ANSI color escape sequences from the output, each line of output is identical to the corresponding line of input. (You can use `strip_ansi_codes` to remove the color escape sequences, like the other tests in `src/tests/test_example_diffs.rs`)

The test should fail on master, and pass in this branch, to prove that this branch fixes the bug.

src/options/set.rs
```diff
@@ -188,10 +191,13 @@ pub fn set_options(
...
     opt.computed.paging_mode = parse_paging_mode(&opt.paging_mode);
 
-    // --color-only is used for interactive.diffFilter (git add -p) and side-by-side cannot be used
+    // --color-only is used for interactive.diffFilter (git add -p) and
+    // side-by-side, file_style, and hunk_header_style cannot be used
     // there (does not emit lines in 1-1 correspondence with raw git output). See #274.
     if opt.color_only {
         opt.side_by_side = false;
+        opt.file_style = "raw".to_string();
+        opt.hunk_header_style = "raw".to_string();
```
dandavison, line 200: It would be good to understand _why_ we need to do this. I can see that adding decorations (e.g. box/lines) would break the one-to-one correspondence between input and output, but setting the color _should_ not break it.

src/options/set.rs
```diff
@@ -59,7 +59,7 @@ macro_rules! set_options {
                              &option_names - &expected_option_names));
             }
         }
-	}
+    }
```
dandavison, line 62: It looks like the lines above here were committed accidentally.

ryuta69:
@dandavison 
Thank you for reviewing dandavison, sorry I now think it was wrong approach.
I am going to re-create.

dandavison:
Hi @ryuta69, ok no problem! I know this code is confusing, that's why I haven't fixed the bug yet :)

ryuta69:
Moved to https://github.com/dandavison/delta/pull/323

PR #322, opened 2020-09-15
Add ppc64le glibc target

runlevel5:
## Context

PowerPC64 LE is a popular platform amongst devs and researchers. Because there are not many distro that package delta for this platform, it is inconvenient to install the app. In many cases, users are forced to compile the app from scratch. I think we should ship ppc64le binary for every new release.

## Changes

* Add ppc64le target in TravisCI manifest

### DIFF
--- a/src/bat/.travis.yml
+++ b/src/bat/.travis.yml
@@ -5,6 +5,9 @@ matrix:
     - os: linux
       rust: stable
       env: TARGET=x86_64-unknown-linux-gnu
+    - os: linux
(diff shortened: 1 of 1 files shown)


dandavison:
Hi @runlevel5 , thank you. I'm happy to do this (I don't think there's much downside to shipping binaries for more platforms). However, it looks like your build is failing at the moment.

runlevel5:
@dandavison I'm fixing it atm

PR #323, opened 2020-09-16
Fix `git add -p` to work when any options enabled

### LINKED ISSUE #293: 🐛 Line numbers incorrect/relative - Windows/Delta 0.4.1
Line numbers seem to be incorrect/relative in delta output. I believe this a bug, but apologies if just expected behavior.

Examples:
---
**git diff** (line numbers should be like 500 or so)
image

**git show** (line numbers should be like at like 330 or so)
image

**git show --no-pager**
image

Configs:
--
Delta 0.4.1 - Windows, executable from github release added to PATH.

**delta --show-config:**
image

**git config --list**
```
http.sslcainfo=C:/Program Files/Git/mingw64/ssl/certs/ca-bundle.crt
http.sslbackend=openssl
diff.astextplain.textconv=astextplain
filter.lfs.clean=git-lfs clean -- %f
filter.lfs.smudge=git-lfs smudge -- %f
filter.lfs.process=git-lfs filter-process
filter.lfs.required=true
credential.helper=manager
core.autocrlf=input
core.fscache=true
core.symlinks=false
pull.rebase=false
core.editor="C:\Users\User\AppData\Local\Programs\Microsoft VS Code\Code.exe" --wait
core.autocrlf=false
core.eol=lf
winupdater.recentlyseenversion=2.25.0.windows.1

include.path=~/.dotfiles/git/git_tweaks
diff.mnemonicprefix=true
diff.algorithm=patience
push.default=simple
log.date=auto:format:%a %Y-%h-%d %I:%M %p %z %Z

include.path=~/.dotfiles/git/git_deltadiff
interactive.difffilter=delta --color-only
delta.features=side-by-side line-numbers decorations
delta.syntax-theme=Dracula
delta.plus-style=syntax #003800
delta.minus-style=syntax #3f0001
delta.hunk-header-style=omit
delta.decorations.commit-decoration-style=bold yellow box ul
delta.decorations.file-style=bold yellow ul
delta.decorations.file-decoration-style=none
delta.decorations.hunk-header-decoration-style=cyan box ul
delta.line-numbers.line-numbers-left-style=cyan
delta.line-numbers.line-numbers-right-style=cyan
delta.line-numbers.line-numbers-minus-style=124
delta.line-numbers.line-numbers-plus-style=28

core.repositoryformatversion=0
core.filemode=false
core.bare=false
core.logallrefupdates=true
core.symlinks=false
core.ignorecase=true
remote.origin.url=<snipped>
remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
branch.master.remote=origin
branch.master.merge=refs/heads/master
user.name=<snipped>
user.email=<snipped>
```

### LINKED ISSUE #320: 🐛 git add -p one-to-one correspondence errors

ryuta69:
Fix #320

```
side-by-side
file-style
hunk-header-style
commit_style
hunk-header-decoration-style
```

These options will decorate the diff output, and when it comes to `git add -p`, diff.interactive.diffFilter will catch the error because it's different from input.

So when color_only is enabled, these decorations shouldn't handled.

## Screenshot

I keep user setting color while keeping output as same as input.

ss 2

## Remarks

1. While editing the code hunk-header-style.is_omitted part, I realized this PR also fixes #293 . It is because even user sets it as omitted, the process of line_numbers_paint actually should have processed. I skipped draw_fn part because `hunk-header-style.is_omitted` shouldn't. (I will add test for this, after this PR merged.)

ss 3

2. I made #321, but it was totally wrong approach....(sorry). Therefore I created this PR.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -388,9 +388,6 @@ fn handle_hunk_header_line(
     plus_file: &str,
     config: &Config,
 ) -> std::io::Result<()> {
-    if config.hunk_header_style.is_omitted {
-        return Ok(());
-    }
     let decoration_ansi_term_style;
     let draw_fn = match config.hunk_header_style.decoration_style {
         DecorationStyle::Box(style) => {
@@ -440,12 +437,14 @@ fn handle_hunk_header_line(
             config.hunk_header_style,
             decoration_ansi_term_style,
         )?;
-    } else {
+    } else if !config.hunk_header_style.is_omitted {
         let line = match painter.prepare(&raw_code_fragment, false) {
             s if s.len() > 0 => format!("{} ", s),
             s => s,
         };
-        writeln!(painter.writer)?;
+        if !config.color_only {
+            writeln!(painter.writer)?;
+        }
         if !line.is_empty() {
             let lines = vec![(line, State::HunkHeader)];
             let syntax_style_sections = Painter::get_syntax_style_sections_for_lines(
@@ -466,25 +465,35 @@ fn handle_hunk_header_line(
                 Some(false),
             );
             painter.output_buffer.pop(); // trim newline
-            draw_fn(
-                painter.writer,
-                &painter.output_buffer,
-                &painter.output_buffer,
-                &config.decorations_width,
-                config.hunk_header_style,
-                decoration_ansi_term_style,
-            )?;
-            if !config.hunk_header_style.is_raw {
-                painter.output_buffer.clear()
-            };
+            if !painter.output_buffer.is_empty() {
+                draw_fn(
+                    painter.writer,
+                    &painter.output_buffer,
+                    &painter.output_buffer,
+                    &config.decorations_width,
+                    config.hunk_header_style,
+                    decoration_ansi_term_style,
+                )?;
+            }
+            painter.output_buffer.clear();
         }
     };
+
+    if config.color_only && !config.hunk_header_style.is_raw {
+        painter.emit()?;
+        writeln!(
+            painter.writer,
+            "{}",
+            format::format_raw_line(&raw_line, config)
+        )?;
+    }
+
     // Emit a single line number, or prepare for full line-numbering
     if config.line_numbers {
         painter
             .line_numbers_data
             .initialize_hunk(line_numbers, plus_file.to_string());
-    } else if config.line_numbers_show_first_line_number && !config.hunk_header_style.is_raw {
+    } else if config.line_numbers_show_first_line_number && !config.color_only && !config.hunk_header_style.is_raw {
         let plus_line_number = line_numbers[line_numbers.len() - 1].0;
         let formatted_plus_line_number = if config.hyperlinks {
             features::hyperlinks::format_osc8_file_hyperlink(
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -193,14 +193,6 @@ where
 
 /// Should a handle_* function be called on this element?
 fn should_handle(state: &State, config: &Config) -> bool {
-    // If interactive.diffFilter is enabled (color-only=true),
-    // output must be same as input, and must not be handled.
-    if config.color_only && (
-        *state == State::FileMeta ||
-        *state == State::CommitMeta
-    ) {
-        return false;
-    }
     if *state == State::HunkHeader && config.line_numbers {
         return true;
     }
@@ -442,9 +434,7 @@ fn handle_hunk_header_line(
             s if s.len() > 0 => format!("{} ", s),
             s => s,
         };
-        if !config.color_only {
-            writeln!(painter.writer)?;
-        }
+        writeln!(painter.writer)?;
         if !line.is_empty() {
             let lines = vec![(line, State::HunkHeader)];
             let syntax_style_sections = Painter::get_syntax_style_sections_for_lines(
@@ -465,35 +455,24 @@ fn handle_hunk_header_line(
                 Some(false),
             );
             painter.output_buffer.pop(); // trim newline
-            if !painter.output_buffer.is_empty() {
-                draw_fn(
-                    painter.writer,
-                    &painter.output_buffer,
-                    &painter.output_buffer,
-                    &config.decorations_width,
-                    config.hunk_header_style,
-                    decoration_ansi_term_style,
-                )?;
-            }
+            draw_fn(
+                painter.writer,
+                &painter.output_buffer,
+                &painter.output_buffer,
+                &config.decorations_width,
+                config.hunk_header_style,
+                decoration_ansi_term_style,
+            )?;
             painter.output_buffer.clear();
         }
     };
-
-    if config.color_only && !config.hunk_header_style.is_raw {
-        painter.emit()?;
-        writeln!(
-            painter.writer,
-            "{}",
-            format::format_raw_line(&raw_line, config)
-        )?;
-    }
 
     // Emit a single line number, or prepare for full line-numbering
     if config.line_numbers {
         painter
             .line_numbers_data
             .initialize_hunk(line_numbers, plus_file.to_string());
-    } else if config.line_numbers_show_first_line_number && !config.color_only && !config.hunk_header_style.is_raw {
+    } else if config.line_numbers_show_first_line_number && !config.hunk_header_style.is_raw {
         let plus_line_number = line_numbers[line_numbers.len() - 1].0;
         let formatted_plus_line_number = if config.hyperlinks {
             features::hyperlinks::format_osc8_file_hyperlink(
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -193,6 +193,14 @@ where
 
 /// Should a handle_* function be called on this element?
 fn should_handle(state: &State, config: &Config) -> bool {
+    // If interactive.diffFilter is enabled (color-only=true),
+    // output must be same as input, and must not be handled.
+    if config.color_only && (
+        *state == State::FileMeta ||
+        *state == State::CommitMeta
+    ) {
+        return false;
+    }
     if *state == State::HunkHeader && config.line_numbers {
         return true;
     }
--- a/src/options/set.rs
+++ b/src/options/set.rs
@@ -19,8 +19,8 @@ use crate::options::option_value::{OptionValue, ProvenancedOptionValue};
 use crate::options::{self, theme};
 
 macro_rules! set_options {
-	([$( $field_ident:ident ),* ],
-     $opt:expr, $builtin_features:expr, $git_config:expr, $arg_matches:expr, $expected_option_name_map:expr, $check_names:expr) => {
+    ([$( $field_ident:ident ),* ],
+    $opt:expr, $builtin_features:expr, $git_config:expr, $arg_matches:expr, $expected_option_name_map:expr, $check_names:expr) => {
         let mut option_names = HashSet::new();
         $(
             let kebab_case_field_name = stringify!($field_ident).replace("_", "-");
@@ -59,7 +59,7 @@ macro_rules! set_options {
                              &option_names - &expected_option_names));
             }
         }
-	}
+    }
 }
 
 pub fn set_options(
--- a/src/options/set.rs
+++ b/src/options/set.rs
@@ -192,6 +192,10 @@ pub fn set_options(
     // there (does not emit lines in 1-1 correspondence with raw git output). See #274.
     if opt.color_only {
         opt.side_by_side = false;
+        opt.file_style = "raw".to_string();
+        opt.commit_style = "raw".to_string();
+        opt.hunk_header_style = "raw".to_string();
+        opt.hunk_header_decoration_style = "none".to_string();
     }
 }
 
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -260,6 +260,9 @@ impl<'a> Painter<'a> {
                 .iter()
                 .zip_eq(diff_style_sections.iter()),
         ) {
+            if state == &State::HunkHeader && config.color_only {
+                continue;
+            }
             let (mut line, line_is_empty) = Painter::paint_line(
                 syntax_sections,
                 diff_sections,
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -260,9 +260,6 @@ impl<'a> Painter<'a> {
                 .iter()
                 .zip_eq(diff_style_sections.iter()),
         ) {
-            if state == &State::HunkHeader && config.color_only {
-                continue;
-            }
             let (mut line, line_is_empty) = Painter::paint_line(
                 syntax_sections,
                 diff_sections,
--- a/src/tests/test_example_diffs.rs
+++ b/src/tests/test_example_diffs.rs
@@ -832,8 +832,9 @@ src/align.rs
     }
 
     #[test]
-    fn test_diff_interactive_filter_has_same_output_with_corresponding_input() {
-        let config = integration_test_utils::make_config_from_args(&[
+    fn test_color_only_output_is_in_one_to_one_correspondence_with_input() {
+        _do_test_output_is_in_one_to_one_correspondence_with_input(&["--color-only", "true"]);
+        _do_test_output_is_in_one_to_one_correspondence_with_input(&[
             "--color-only",
             "true",
             "--file-style",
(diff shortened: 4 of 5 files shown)


src/delta.rs
```diff
@@ -193,6 +193,14 @@ where
...
 /// Should a handle_* function be called on this element?
 fn should_handle(state: &State, config: &Config) -> bool {
+    // If interactive.diffFilter is enabled (color-only=true),
+    // output must be same as input, and must not be handled.
+    if config.color_only && (
+        *state == State::FileMeta ||
+        *state == State::CommitMeta
+    ) {
+        return false;
+    }
```
ryuta69, line 203: In any cases the state is FileMeta or CommitMeta when interactiveFilter, I confirmed it never should handle. However, when some big function will be added in handle_generic_file_meta_header_line or handle_commit_meta_header_line in future, this workaround should move in each function.

src/delta.rs
```diff
@@ -432,12 +437,14 @@ fn handle_hunk_header_line(
             config.hunk_header_style,
             decoration_ansi_term_style,
         )?;
-    } else {
+    } else if !config.hunk_header_style.is_omitted {
```
ryuta69, line 440: This is where fixes #293.

src/delta.rs
```diff
@@ -458,25 +465,35 @@ fn handle_hunk_header_line(
...
+            if !painter.output_buffer.is_empty() {
+                draw_fn(
+                    painter.writer,
+                    &painter.output_buffer,
+                    &painter.output_buffer,
+                    &config.decorations_width,
+                    config.hunk_header_style,
+                    decoration_ansi_term_style,
+                )?;
+            }
```
ryuta69, line 477: I found draw::write_*** is drawing empty line if the text is empty.
This part is gather text after paint_lines formatted, then empty line should removed because it causes difference error of diff.interactive.diffFilter.

I confirm there is no any case that text will be empty except when color_only.

src/paint.rs
```diff
@@ -260,6 +260,9 @@ impl<'a> Painter<'a> {
                 .iter()
                 .zip_eq(diff_style_sections.iter()),
         ) {
+            if state == &State::HunkHeader && config.color_only {
+                continue;
+            }
```
ryuta69, line 265: In specific, 
```
@@ -444,6 +444,8 @@ fn handle_hunk_header_line(
```

will be formatted to 

```
fn handle_hunk_header_line(
```

in this part. So I added only when state is hunkheader and color-only is enabled, skip the process.

dandavison:
Hi @ryuta69, this looks great! I'll review as soon as possible.

src/options/set.rs
```diff
@@ -59,7 +59,7 @@ macro_rules! set_options {
                              &option_names - &expected_option_names));
             }
         }
-	}
+    }
```
dandavison, line 62: The changes in this file are not intended, right? (Don't worry about adding more commits to fix, because I usually merge the PRs with squash-and-merge so it will become one commit in the end.)

ryuta69: @dandavison 
Oh, I'd like to make sure if this is mistake.

I knew you pointed out in previous PR as well, but I actually didn't get it.
https://github.com/dandavison/delta/pull/321#discussion_r487985240

This bracket is paired to here, starting bracket with 4 spaces, so I thought it was ok.
https://github.com/dandavison/delta/pull/323/files#diff-8b49f2774c51a77ad662f1c88bfd1579R22-R23

The reason why I changed tab into 4 spaces is that whole code looks ruled so. This `macro_rules! builtin_feature`'s argument line is so.
https://github.com/dandavison/delta/blob/master/src/features/mod.rs#L66

dandavison: Ah, OK, I didn't realize I had tabs in there. That's fine then, thanks. (The thing that was making me ask is that usually I like to only have relevant changes in a PR, and do reformatting changes separately. But no need to change it now!)

ryuta69: I'm sorry for not separating this. I thought this one is really tiny, but made you confused in fact. I'll be careful.

src/delta.rs
dandavison: Can you show me how to reproduce the bug/problem you're describing here?

ryuta69: @dandavison 
Yes, I made branch `repro_drawfn_bug_for_fix_diff_interactivefilter` in ryuta69/delta to reproduce the bug.

https://github.com/ryuta69/delta/pull/1

You can reproduce the bug by switching to this branch and run `cargo test tests::test_example_diffs::tests::test_diff_interactive_filter_has_same_output_with_corresponding_input -- --nocapture`.

I attached the picture in this PR to show how the bug is caused.

dandavison:
@ryuta69 this is really great work, thanks very much. Because this has been the source of bugs, I'm going to ask a few questions and have a bit of back and forth before we get it merged. At this point you probably understand these bugs better than me so please watch out for me getting something wrong or not seeing consequences etc!

dandavison:
OK, I have a question: what do you think about preserving some aspects of your #321 PR? Specifically, could we add code like this in `options/set.rs`:

```rust
    // --color-only is used for interactive.diffFilter (git add -p) and side-by-side cannot be used
    // there (does not emit lines in 1-1 correspondence with raw git output). See #274, #320
    if opt.color_only {
        opt.side_by_side = false;
        opt.commit_style = "raw".to_string();
        opt.file_style = "raw".to_string();
        opt.hunk_header_style = "raw".to_string();
        opt.hunk_header_decoration_style = "none".to_string();
    }
```

There are a couple of things that are making me suggest this:

1. I would _really_ love to simplify the code. Already in master, there are various `if` conditions and other logic related to `color-only` and `interactive.diffFilter` scattered around the  code and it is hard to understand the logic (in master, before your PR).

1. `hunk-header-style=omit` says to change the line structure but `color-only` promises to preserve the line structure. So they simply don't make sense together. So I think it makes sense to override it as early as possible.

1. Overriding the options in `set.rs` is like saying "OK, `color-only` is active, let's configure delta right at the beginning in a way that makes sense, and then the downstream code will work". I think that is a better design than allowing mixtures of options that don't make sense, and then attempting to make the downstream code do the right thing.

src/delta.rs
dandavison: I believe that if we override options in `set.rs` this conditional logic will not be needed, right?

src/delta.rs
dandavison: Awesome!

src/delta.rs
```diff
@@ -458,25 +465,35 @@ fn handle_hunk_header_line(
...
+
+    if config.color_only && !config.hunk_header_style.is_raw {
+        painter.emit()?;
+        writeln!(
+            painter.writer,
+            "{}",
+            format::format_raw_line(&raw_line, config)
+        )?;
+    }
+
```
dandavison, line 490: Similarly, I believe that if we override options in set.rs this code will not be needed.

src/tests/test_example_diffs.rs
```diff
@@ -831,6 +831,33 @@ src/align.rs
         );
     }
 
+    #[test]
```
dandavison, line 834: Thanks a lot for writing this test, this is extremely useful!

Can we change the name of the test to `test_color_only_output_is_in_one_to_one_correspondence_with_input`?

Also, how about adding a couple more tests, like

```rust
    #[test]
    fn test_color_only_output_is_in_one_to_one_correspondence_with_input_1() {
        let config = integration_test_utils::make_config_from_args(&["--color-only", "true"]);
        let output = integration_test_utils::run_delta(GIT_DIFF_SINGLE_HUNK, &config);

        let output_lines: Vec<&str> = output.split('\n').collect();
        let input_lines: Vec<&str> = GIT_DIFF_SINGLE_HUNK.split('\n').collect();

        assert_eq!(input_lines.len(), output_lines.len());

        for n in 0..input_lines.len() {
            assert_eq!(input_lines[n], output_lines[n]);
        }
    }

    #[test]
    fn test_color_only_output_is_in_one_to_one_correspondence_with_input_2() {
        let config = integration_test_utils::make_config_from_args(&[
            "--color-only",
            "true",
            "--file-style",
            "blue",
            "--commit-style",
            "red",
            "--hunk-header-style",
            "syntax",
            "--hunk-header-decoration-style",
            "box",
        ]);
        let output = integration_test_utils::run_delta(GIT_DIFF_SINGLE_HUNK, &config);

        let output_lines: Vec<&str> = output.split('\n').collect();
        let input_lines: Vec<&str> = GIT_DIFF_SINGLE_HUNK.split('\n').collect();

        assert_eq!(input_lines.len(), output_lines.len());

        for n in 0..input_lines.len() {
            assert_eq!(input_lines[n], output_lines[n]);
        }
    }

```

dandavison:
Could you run `cargo fmt` to auto-format the code? (There's one file `test_example_diffs.rs` where my code wasn't formatted; I just pushed that to master so you can ignore it.)

ryuta69:
@dandavison 
Hi Thank you for reviewing.

I actually agree with you. After I moved to this new PR, and coding on new approach, I realized this becomes too messy... The previous approach was better.

>hunk-header-style=omit says to change the line structure but color-only=true promises to preserve the line structure. So they simply don't make sense together. So I think it makes sense to override it as early as possible

Yes, after investigating the code, I think it's true.
I will fix up now.

src/delta.rs
ryuta69: Yes, it won't be needed anymore.

src/delta.rs
ryuta69: Yes, we don't need this anymore.

src/tests/test_example_diffs.rs
ryuta69: Of cource! Yes, the name is more clear. Thank you for pointing out.

src/delta.rs
ryuta69: We won't need this fix anymore, as same as https://github.com/dandavison/delta/pull/323#r490649549

src/paint.rs
ryuta69: this won't be needed as same as https://github.com/dandavison/delta/pull/323#r490649549

src/delta.rs
ryuta69: Fixed here 8a5fb65

src/delta.rs
ryuta69: Fixed here 8a5fb65

src/paint.rs
ryuta69: Fixed here 8a5fb65

src/delta.rs
ryuta69: Fixed here 8a5fb65

src/tests/test_example_diffs.rs
ryuta69: Added here 4fd75f1

ryuta69:
@dandavison 
Hi, I fixed the part you pointed out.
I run `cargo fmt` as well.

src/delta.rs
```diff
@@ -466,11 +463,10 @@ fn handle_hunk_header_line(
                 config.hunk_header_style,
                 decoration_ansi_term_style,
             )?;
-            if !config.hunk_header_style.is_raw {
-                painter.output_buffer.clear()
-            };
+            painter.output_buffer.clear();
```
dandavison, line 466: What bug is this fixing?

dandavison: Would you be able to add test coverage for the changes in this file? Specifically, when I revert the changes in this file like this:

```
git checkout master src/delta/src/delta.rs
```

right now, the full test suite still passes. I'd really like to have tests that fail when the changes in this file are reverted. Do you think you have time do do that?

ryuta69: I should have left comment.
This if statement is not needed because when config.hunk_header_style.is_raw, it won't come to this process.

https://github.com/ryuta69/delta/blob/4fd75f1b9f4abd1040126f2e87bdf4508947e200/src/delta.rs#L420

ryuta69: @dandavison

So, this is not a bug, but i just refactored (again, I should separate PR. I'm sorry...).
The tests were not failed before this fix.

dandavison: I see, thanks!

OK, so one of us should add test coverage for the #293 fix. Do you feel like doing that?

ryuta69: Yes!
As I stated `I will add test for this` in Remark, I am going to make a test for this!

src/tests/test_example_diffs.rs
```diff
@@ -831,6 +831,50 @@ src/align.rs
...
+        let output_lines: Vec<&str> = output.split('\n').collect();
+        let input_lines: Vec<&str> = GIT_DIFF_SINGLE_HUNK.split('\n').collect();
+
+        assert_eq!(input_lines.len(), output_lines.len());
+
+        for n in 0..input_lines.len() {
+            assert_eq!(input_lines[n], output_lines[n]);
+        }
+    }
+
```
dandavison, line 877: Nice!

src/delta.rs
dandavison: Oh yes :) Awesome, I'll let you do it in a separate PR.

dandavison (approved):
Thanks very much for this work @ryuta69!

src/delta.rs
ryuta69: Ok!

dandavison:
@ryuta69 I wonder whether the changes in #272 can be simplified after your PR? Just noting something to look at -- I haven't investigated carefully yet.

ryuta69:
@dandavison 
Sure!

For example, I think we can remove this if statement, 
https://github.com/dandavison/delta/blob/788c3636ece17612c7adf262f209990f603bb512/src/delta.rs#L139-L143

and this whole parts,

https://github.com/dandavison/delta/blob/788c3636ece17612c7adf262f209990f603bb512/src/delta.rs#L196-L198

Because when `*state == State::HunkHeader`, this `should_handle` returns false only when `!config.line-number && style.is_raw && style.decoration_style == DecorationStyle::NoDecoration` .

However, even if config.line-number isn't set and process `fn handle_hunk_header_line` , this if statement will block not to display line-number anyway.

https://github.com/dandavison/delta/blob/788c3636ece17612c7adf262f209990f603bb512/src/delta.rs#L473-L476

Plus, `fn handle_hunk_header_line` already handle `style.is_raw && style.decoration_style == DecorationStyle::NoDecoration`.

This can simplify the `should_handle` function, and we can put controller of hunk-header only in `fn handle_hunk_header_line`.

I'm on a trip right now, so I'll look into it in 2 days!

Issue #319, opened 2020-09-09
Is it possible to not use hunk header and see absolute (true) not relative line numbers?


nortliteborealis:
When I disable hunk header (**_set hunk-header-style = omit_**) the line numbers are calculated as relative numbers, but i want to see absolute numers in diff. I realy want see absolute line numbers without hunk header, please tell me how to set it up, if of course it's possible.
Thank you in advance))

With hunk headers
![with hunk headers](https://user-images.githubusercontent.com/44495093/92603170-aaff0c80-f2b7-11ea-8ce0-35bafb2926ac.png)
Without hunk headers
![without hunk headers](https://user-images.githubusercontent.com/44495093/92603500-0df0a380-f2b8-11ea-91f9-573c540aead0.png)

My git config

```
[interactive]
    diffFilter = delta --color-only

[diff]
    colorMoved = default

[delta]
    side-by-side = true
    features = line-numbers decorations
    whitespace-error-style = 22 reverse
    hunk-header-style = omit
    syntax-theme = base16

[delta "line-numbers"]
    line-numbers = true
    line-numbers-minus-style = red
    line-numbers-zero-style = "#444444"
    line-numbers-plus-style = green
    line-numbers-left-format = "{nm:^4}┊"
    line-numbers-right-format = "{np:^4}│"
    line-numbers-left-style = red
    line-numbers-right-style = green

[delta "decorations"]
    commit-decoration-style = bold yellow box ul
    file-style = bold yellow ul
    file-decoration-style = none
    hunk-header-decoration-style = yellow box
```

dandavison:
Hi @nortliteborealis! You're right, this is a bug with `hunk-header-style=omit`. It's actually the same as reported in #293 and #274 (but the issue titles make it hard to search for.) It's the top priority bug to fix next.

nortliteborealis:
Thank you very much for your reply and I apologize for duplicating the issue. I will wait for a solution of this problem. Delta is a super tool!

dandavison:
Hi @nortliteborealis, 

@ryuta69's work is fixing this issue (not released yet).

> I really want see absolute line numbers without hunk header,

Can you tell me more about how you use delta without hunk headers. If there's a file with multiple hunks, how do you tell when one hunk ends and the next starts?

Would you like there to be a blank line separating the hunks? Or do you want the next hunk to literally follow straight on, as if there was no break in the underlying code?

nortliteborealis:
Hi @dandavison 

[...]

PR #325, opened 2020-09-18
[Suggestion] Add break line in each part of lines when hunk header style is omit

ryuta69:
While creating the hunk-header-style's line-number test, I realized that hunk-header-style is kind of hard to read if there are two or more parts like below, because there is no break line between each parts.

ss 1

I should have refactored this [here](https://github.com/dandavison/delta/pull/323), but I didn't find that. So I fixed now like below. I think it's easier to read.
ss 2

## Remark
As you can see the code in picture, the hunk-header-style's line-number test is already created.

I will send this test PR after this suggestion PR.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -429,7 +429,9 @@ fn handle_hunk_header_line(
             config.hunk_header_style,
             decoration_ansi_term_style,
         )?;
-    } else if !config.hunk_header_style.is_omitted {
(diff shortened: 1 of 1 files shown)


dandavison (commented):
Thanks! I asked a question in https://github.com/dandavison/delta/issues/319 to someone who uses this option, but this makes sense to me.

PR #327, opened 2020-09-18
Add test for line numbers with hunk header style is omit

ryuta69:
Hi, this is related to https://github.com/dandavison/delta/pull/323#discussion_r490999066 .


dandavison (approved):
LGTM! Thanks again @ryuta69.

(I see that this test fails as expected when cherry-picked on to a commit prior to your PR #323 and passes in this branch.)

PR #331, opened 2020-09-22
Add identical test when color_only and line_numbers are enabled at same time

ryuta69:
This's not a bug, but the test does an unnecessary checking. The test used to fail with line-numbers, but it shouldn't fail. (I found this while I was trying simplifying the code from https://github.com/dandavison/delta/pull/323#issuecomment-695789313.)

### Why delete assert_eq of each lines

I thought `git add -p` checks each lines are identical, but it actually checks only line-counts.

From [Git source code](https://github.com/git/git/blob/master/add-patch.c), the flow is `git add -p` -> [run_add_p](https://github.com/git/git/blob/master/add-patch.c#L1667) -> [parse_diff](https://github.com/git/git/blob/master/add-patch.c#L385).

Inside `parse_diff`, they run these commands.

```
# for input
git diff-files -p --no-color

# for output
git diff-files -p --color | delta --color-only
```

Then, it checks only their line-counts.
https://github.com/git/git/blob/master/add-patch.c#L609-L615
(Perl code is more understandable https://github.com/git/git/blob/master/git-add--interactive.perl#L717-L731)

I confirmed above by inserting a trash string inside and add --line-numbers option.
ss 3

### Remarks
I should have known about this when I created this test, sorry.

After this PR gets merged, my simplified code of https://github.com/dandavison/delta/pull/323#issuecomment-695789313 will pass the all tests.

### DIFF
--- a/src/tests/test_example_diffs.rs
+++ b/src/tests/test_example_diffs.rs
@@ -876,9 +876,25 @@ src/align.rs
         let output = integration_test_utils::run_delta(GIT_DIFF_SINGLE_HUNK, &config);
         let output = strip_ansi_codes(&output);
 
-        let input_lines = GIT_DIFF_SINGLE_HUNK.split('\n').count();
-        let output_lines = output.split('\n').count();
-        assert_eq!(input_lines, output_lines);
+        let input_lines: Vec<&str> = GIT_DIFF_SINGLE_HUNK.split('\n').collect();
+        let output_lines: Vec<&str> = output.split('\n').collect();
+        assert_eq!(input_lines.len(), output_lines.len());
+
+        // Although git patch options only checks the line counts of input and output,
+        // we should check if they are identical as well to avoid unexpected decoration.
+        for n in 0..input_lines.len() {
+            let input_line = input_lines[n];
+            // If config.line_numbers is enabled,
+            // we should remove line_numbers decoration while checking.
+            let output_line = if config.line_numbers && n > 11 && n < input_lines.len() - 1 {
+                &output_lines[n][14..]
+            } else {
+                output_lines[n]
+            };
+            // TODO: this trim() can be removed by simplifing width_boxed of draw_fn.
+            assert_eq!(input_line.trim(), output_line.trim());
+            // assert_eq!(input_line, output_line);
+        }
     }
 
     #[test]
--- a/src/tests/test_example_diffs.rs
+++ b/src/tests/test_example_diffs.rs
@@ -841,6 +841,13 @@ src/align.rs
         _do_test_output_is_in_one_to_one_correspondence_with_input(&[
             "--color-only",
             "true",
+            "--hunk-header-style",
+            "normal",
+            "--line-numbers",
+        ]);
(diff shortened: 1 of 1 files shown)


dandavison:
> I thought git add -p checks each lines are identical, but it actually checks only line-counts.

My worry is that delta could cause an incorrect patch to be applied, so I was thinking that it was actually great to have that test, even if it is stronger than necessary. What do you think about asserting equality after chopping off the line numbers (or using a regular expression to make an equivalent assertion)?

Specifically, here's what I'm worrying about. Does my worry make sense?

Raw git output:

```
patch line 1
patch line 2
patch line 3
```

Delta output

```
patch line 1
patch line 2 changed somehow by delta
patch line 3
```
 
My worry is that `git add -p` would say "OK, fine, the number of lines is the same", and then a patch would be committed containing incorrect content altered by delta.

ryuta69:
@dandavison 
Yes! That makes sense.
Although we already have a test `test_git_diff_is_unchanged_under_color_only`, `color_only + SomeOtherOptions` perhaps causes something wrong as you worry.

https://github.com/dandavison/delta/blob/788c3636ece17612c7adf262f209990f603bb512/src/tests/test_example_diffs.rs#L1227-L1236

The reason why I created this PR is that it actually leaded me confused while debugging, like "Oh a test fails. So this is not something right  ---->>>> OMG it actually works".

> (or using a regular expression to make an equivalent assertion)

Yes, I'm going to add some process inside to relieve your concerns at the same time it works as `git add -p` actually behaves.

dandavison:
> Yes, I'm going to add some process inside to relieve your concerns at the same time it works as git add -p actually behaves.

OK, great!

ryuta69:
@dandavison 

Hi I added identical checking.
This test works when `color-only + line-numbers` are enabled as well.

d6ccfcc

As I commented, trim() can be removed soon.
Right now, it inserts unnecessary whitespace into hunk-header like below,
ss 4
because of this part.

https://github.com/dandavison/delta/blob/788c3636ece17612c7adf262f209990f603bb512/src/delta.rs#L426-L427

I already made a fix while simplifying the code, so I will send PR after this test gets merged.

dandavison (approved):
Awesome,  LGTM!

dandavison:
@ryuta69 is there anything more you would like to do before the next delta release?

ryuta69:
@dandavison 
Hi!

I would love to add

1. Refactoring should handle, and test for hunk_header_line. https://github.com/dandavison/delta/pull/333 . You pointed out that this PR caused breaking change, but now fixed.

2. Skip bool config from command-line. https://github.com/dandavison/delta/pull/340 . You pointed out that naming is kind of confusing, so I add some another suggestion. I can fix in a day.

2 is not necessary, but I think 1 should get resolved before release.

dandavison:
OK cool, that's great if you'd still like to do more work! No immediate hurry.

ryuta69:
@dandavison 
Thank you!

Oh, I replied to you in those PR, so I would love you to look at when you have a time....

PR #333, opened 2020-09-23
Refactor handle_hunk_header_line and should_handle

ryuta69:
Related to https://github.com/dandavison/delta/pull/331, and related to simplifying handle_hunk_header_line.

~~It used to handle whitespaces of box from handler in delta.rs, not from write_boxed in draw.rs.~~

~~This causes a whitespaces into hunk-header when `line_numbers=true and hunk-header-style=raw`, and while I'm simplifying the code, it makes assert_eq tests failed. So I decided to refactor this part. In any cases you draw a box, you always insert a whitespace after the last character. Therefore, we can move this whitespace handle into draw.rs.~~

~~This could prevent us from adding whitespaces or using pad flag at every time we want to write_box.~~

↑
This refactoring whitespaces was wrong approach.
Now I just refactored handle_hunk_header_line and should_handle.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -231,25 +231,20 @@ fn handle_commit_meta_header_line(
         return Ok(());
     }
     let decoration_ansi_term_style;
-    let mut pad = false;
     let draw_fn = match config.commit_style.decoration_style {
         DecorationStyle::Box(style) => {
-            pad = true;
             decoration_ansi_term_style = style;
             draw::write_boxed
         }
         DecorationStyle::BoxWithUnderline(style) => {
-            pad = true;
             decoration_ansi_term_style = style;
             draw::write_boxed_with_underline
         }
         DecorationStyle::BoxWithOverline(style) => {
-            pad = true;
             decoration_ansi_term_style = style;
             draw::write_boxed // TODO: not implemented
         }
         DecorationStyle::BoxWithUnderOverline(style) => {
-            pad = true;
             decoration_ansi_term_style = style;
             draw::write_boxed // TODO: not implemented
         }
@@ -287,8 +282,8 @@ fn handle_commit_meta_header_line(
 
     draw_fn(
         painter.writer,
-        &format!("{}{}", formatted_line, if pad { " " } else { "" }),
-        &format!("{}{}", formatted_raw_line, if pad { " " } else { "" }),
+        &format!("{}", formatted_line),
+        &format!("{}", formatted_raw_line),
         &config.decorations_width,
         config.commit_style,
         decoration_ansi_term_style,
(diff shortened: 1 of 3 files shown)


src/tests/test_example_diffs.rs
```diff
@@ -891,9 +891,7 @@ src/align.rs
             } else {
                 output_lines[n]
             };
-            // TODO: this trim() can be removed by simplifing width_boxed of draw_fn.
-            assert_eq!(input_line.trim(), output_line.trim());
-            // assert_eq!(input_line, output_line);
+            assert_eq!(input_line, output_line);
```
ryuta69, line 894: Resolve TODO of https://github.com/dandavison/delta/pull/331#issuecomment-698022161 here.

dandavison:
Hi @ryuta69, is this PR intended to change the behavior of delta or is it just a refactor?

Looking at git blame takes me to this commit: https://github.com/dandavison/delta/commit/470fb46a0a23052ea40ddc0906175d9ca54a18cb "Bug fix: hunk header background color was not reaching box line" and it looks like this PR does add blank space at the end of the hunk header, e.g. compare

```
git show | delta --hunk-header-style='blue "#dddddd"'
```

before and after this commit.

before
image

after:
image

I like the code simplification but can you clarify whether you're intending to make this change to behavior?

ryuta69:
@dandavison 
Thank you for reviewing, actually I didn't intend that changing behavior.

I actually didn't notice that commit, but looked https://github.com/dandavison/delta/commit/0d593c2d one. 

So this PR shouldn't get merged, but I want you to see my purpose, and decide which way is better.

Please look for this https://github.com/ryuta69/delta/pull/2. I wrote the detail about my purpose and the two plans of simplifying.

1. Simplify the `should_handle` and `pad`, but causes changing behavior https://github.com/dandavison/delta/pull/333#issuecomment-699079852 .
2. Simplify the `should_handle` and not causing changing behavior, but needs to add `pad` in handle-hunk-header-line.

Right now I think 2nd one makes sense because it should have had a `pad` anyway.

I'll change this PR to the plan you choose.

ryuta69:
@dandavison 
Hi, as I said yesterday, I wish you could review this again...

src/delta.rs
```diff
@@ -193,9 +191,6 @@ where
 
 /// Should a handle_* function be called on this element?
 fn should_handle(state: &State, config: &Config) -> bool {
-    if *state == State::HunkHeader && config.line_numbers {
-        return true;
-    }
```
dandavison, line 198: :)

dandavison: That means: this change makes me happy!

dandavison (approved):
Great, I really like this clean-up!

ryuta69:
Thank you for your review!

Issue #307, opened 2020-09-01
🐛 Can't disable line numbering from command line


HaleTom:
If I run:

    git diff | delta --line-numbers=false | cat

I see line numbers.

How do I disable line numbers from the command line?

I tried `--no-line-numbers`, but no joy.  It would be great to be able to disable any boolean with `--no-*`, to override something that had been previously enabled (eg by a script or bash alias)

dandavison:
Hi @HaleTom, thanks for this. I agree it would be nice if this were more convenient. I think the best that's available right now is:

1. `git config --global delta.line-numbers false; git diff`
  Not perfect but I've been finding this approach quite useful. Of course one can go further in this direction with custom shell aliases / wrapper functions.

1. `delta --no-gitconfig`
  But then you have to build back up the options you do require

A couple of tangential thoughts:

Sometimes I wonder if it might have been better not to use boolean flags at all in delta and only use flags with explicit values, especially seeing as delta is mostly not used directly at the command line. However, I don't think there's a path there that retains backwards compatibility.

Is adding `--no-*` negating versions automatically something that one could hope that Rust's `clap` (and hence `structOpt`) package could do automatically? I don't _think_ it does.

dandavison:
Ref https://github.com/clap-rs/clap/issues/815 ("Add support for automatic negation flags")

dandavison:
The current plan for addressing this is https://github.com/dandavison/delta/pull/340#issuecomment-744257639 (and see  https://github.com/dandavison/delta/issues/446).

dandavison:
@HaleTom is the following a workaround for you?

```gitconfig
[delta]
    features = line-numbers

[delta "no-line-numbers"]
    line-numbers = false
```

Now by default delta will have `line-numbers` and they can be disabled with `delta --features=no-line-numbers`, or `DELTA_FEATURES=no-line-numbers git show` etc. There's similar discussion at https://github.com/dandavison/delta/issues/492#issuecomment-755821580 and https://github.com/dandavison/magit-delta/issues/13#issuecomment-754736108.

dandavison:
In the master branch we can now do `git -c delta.line-numbers=false` (see #493).

So I think that's the most convenient way to do it for the case of using delta with git. It _should_ also be possible to disable it on the delta command line somehow, but perhaps that's less of  a priority since we can do it conveniently in the git case, and we have the features workaround https://github.com/dandavison/delta/issues/307#issuecomment-755823411 otherwise.

[...]

PR #340, opened 2020-09-28
Skip specified settings in .gitconfig

ulwlu:
Fix https://github.com/dandavison/delta/issues/307.

Bool option couldn't be disabled by command.
There was three options to solve this,

1. Make all bool options into string <- kind of messy.
2. Make --no-** flag <- kind of messy. Need to add each no-flags. Even if https://github.com/TeXitoi/structopt/issues/320 will be able to do, still need to handle in set_option.
3. [This PR] Change `--no-gitconfig` option to Vec<String>, and it will ignore those specified keys. <- This one option can disable as much options as you want. Also, you can skip some non-bool config at same time. It will be useful.

I'm still not sure if this way is the best, so I titled 'suggestion'.
I would like to hear any ideas.

ss 1
ss

### DIFF
--- a/src/git_config.rs
+++ b/src/git_config.rs
@@ -7,6 +7,7 @@ use git2;
 pub struct GitConfig {
     config: git2::Config,
     pub enabled: bool,
+    pub ignored_keys: Vec<String>,
     pub repo: Option<git2::Repository>,
 }
 
@@ -30,6 +31,7 @@ impl GitConfig {
                     config,
                     repo,
                     enabled: true,
+                    ignored_keys: vec![],
                 })
             }
             None => None,
@@ -42,6 +44,7 @@ impl GitConfig {
             config: git2::Config::open(path).unwrap(),
             repo: None,
             enabled: true,
+            ignored_keys: vec![],
         }
     }
 
@@ -49,11 +52,18 @@ impl GitConfig {
     where
         T: GitConfigGet,
     {
-        if self.enabled {
-            T::git_config_get(key, self)
-        } else {
-            None
+        if !self.enabled {
+            return None;
         }
+
+        if !self.ignored_keys.is_empty() {
+            let key_name = &key.rsplit('.').collect::<Vec<&str>>()[0];
+            if self.ignored_keys.contains(&key_name.to_string()) {
+                return None;
+            }
+        }
+
+        T::git_config_get(key, self)
     }
 }
 
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ BENCHMARK_INPUT_FILE = /tmp/delta-benchmark-input.gitdiff
 BENCHMARK_COMMAND = git log -p 23c292d3f25c67082a2ba315a187268be1a9b0ab
 benchmark: build
 	$(BENCHMARK_COMMAND) > $(BENCHMARK_INPUT_FILE)
-	hyperfine 'target/release/delta --no-gitconfig < $(BENCHMARK_INPUT_FILE) > /dev/null'
+	hyperfine 'target/release/delta --no-gitconfig=all < $(BENCHMARK_INPUT_FILE) > /dev/null'
 
 # https://github.com/brendangregg/FlameGraph
 flamegraph: build
--- a/README.md
+++ b/README.md
@@ -581,7 +581,10 @@ FLAGS:
         --show-syntax-themes         Show all available syntax-highlighting themes, each with an example of highlighted
                                      diff output. If diff output is supplied on standard input then this will be used
                                      for the demo. For example: `git show --color=always | delta --show-syntax-themes`
-        --no-gitconfig               Do not take any settings from git config. See GIT CONFIG section
+        --no-gitconfig               Do not take specified settings from git config.
+                                     For example: `git show | delta --no-gitconfig=line-numbers,file-style`.
+                                     You can ignore all settings by specifying `--no-gitconfig=all`.
+                                     See GIT CONFIG section.
         --raw                        Do not alter the input in any way. This is mainly intended for testing delta
         --color-only                 Do not alter the input structurally in any way, but color and highlight hunk lines
                                      according to your delta configuration. This is mainly intended for other tools that
--- a/etc/bin/diagnostics
+++ b/etc/bin/diagnostics
@@ -8,8 +8,8 @@ commands=(
     "printf '\e[38;5;19m\e[48;5;226mtext\e[m\n'"
     "printf '\e[38;2;0;0;255m\e[48;2;255;255;0mtext\e[m\n'"
     "printf '⋮│─'"
-    "delta --no-gitconfig <(echo a) <(echo b) | cat -A"
-    "delta  --no-gitconfig <(echo a) <(echo b)"
+    "delta --no-gitconfig=all <(echo a) <(echo b) | cat -A"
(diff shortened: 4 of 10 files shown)


dandavison:
Hi @ryuta69, isn't this confusingly similar to the existing `--no-gitconfig`?

ryuta69:
@dandavison 
Hi, the biggest difference is  that --no-gitconfig disables whole settings in git-config while --skip-config disables only settings you specified (eg, only line-numbers as original issue says.)

But as looking at the names, they are maybe kinda similar as you said.

I gonna ask the issue owner if this is clear or not.

ryuta69:
@dandavison 
This is an another suggestion, how do you think making --no-gitconfig into String.

Then if its value is `all`, it will skip all config as current `--no-gitconfig` does.
If it specifies the keys of gitconfig, it will skip only those specified key like this PR does.
Default value would be `none`.

ryuta69:
@dandavison 
Hi, sorry to ask you many times, but as I stated yesterday I would like to hear your opinion.

There is an another suggestion, how do you think below plans?

1. Make option's name be `--ignore`
2. Don't make new option. Make `--no-gitconfig` into string, and if value is `all` it will skip all gitconfig. If value is specific config, it will skip only that config.

dandavison:
Hi @ryuta69, thanks for working on this and thinking about it. I still need to think about it a bit more, but I do like your idea changing `--no-gitconfig` to be a string. Would you agree that this isn't urgent, and the first priority is releasing your improvements to `git add -p`?

ryuta69:
@dandavison 
I agree!

As I stated here https://github.com/dandavison/delta/pull/331#issuecomment-703334112, this one is not urgent for release in my opinion.

I'm going to work for this, but there's no problem to release without this because the --pager is the first priotiry.

src/git_config.rs
```diff
@@ -42,18 +44,26 @@ impl GitConfig {
...
-        if self.enabled {
-            T::git_config_get(key, self)
-        } else {
-            None
+        if !self.enabled {
+            return None;
         }
+
+        if !self.ignored_keys.is_empty() {
+            let key_name = &key.rsplit('.').collect::<Vec<&str>>()[0];
```
ryuta69, line 60: keys will be 

1. core.pager
2. delta.line-numbers
3. delta.features.line-numbers

I think users don't want to specify parent key, so I split key here.

ryuta69:
@dandavison 
Hi, I fixed this PR by changing --no-gitconfig to be a string.
I would be happy if you could review this when you have a time.

dandavison:
Thanks @ryuta69, I'll review as soon as I can.

Your other changes are all merged, right? So we could do a delta release now?

ryuta69:
@dandavison 
Absolutely!

All other PRs are merged, so there are no problems for release.

dandavison:
OK, all your changes are released in delta 0.4.4, thanks very much for all the work!

ulwlu:
@dandavison 
Hi, just a reminder.
I'd be happy if you could review this when you have a time.

dandavison:
Hi @ulwlu, I'm hesitant about this one. Here is what I'm asking myself:

1. Is the added complexity to the interface justified?
1. Is it clear to users, or is it confusing?
1. In #307 @HaleTom said
    > It would be great to be able to disable any boolean with --no-*, to override something that had been previously enabled (eg by a script or bash alias)
    However, this approach would work for neither scripts nor bash aliases since it would not override a command line arg.
1. This wouldn't be a standard pattern for a UNIX tool; it somehow feels a bit convoluted to me to have an option specifying which entries in a config file to disregard.
1. I wonder whether we should consider environment variables instead? For example, perhaps for every delta flag, delta could look an an environment variable with a predictable name: `DELTA_NO_LINE_NUMBERS`, `DELTA_NO_SIDE_BY_SIDE`, in general if there is a boolean setting `xxx-yyy` then delta would automatically look for `DELTA_NO_XXX_YYY`.

But I don't have a clear sense here of what is best.

ulwlu:
@dandavison 
I agree with your concern, and now thinking below is the cleanest way to solve.

> I wonder whether we should consider environment variables instead? For example, perhaps for every delta flag, delta could look an an environment variable with a predictable name: DELTA_NO_LINE_NUMBERS, DELTA_NO_SIDE_BY_SIDE, in general if there is a boolean setting xxx-yyy then delta would automatically look for DELTA_NO_XXX_YYY

The purpose of creating this PR was written in description (I am trying to achieve with less maintenance. Since there is no automatic negative boolean in clap).

I'm going to close this, and will keep above plan in mind.

dandavison:
Thanks @ulwlu, sounds like we agree. Let's keep thinking about the environment variable option. Above I suggested `DELTA_NO_XXX_YYY`, but another option would be `DELTA_XXX_YYY`. This was the env var could be used to override gitconfig in both a positive and negative direction. We would have to decide on the rules for which strings mean true and which mean false though, which is always a slightly difficult decision.

HaleTom:
Could we have a 2nd config file, possibly using the same format as gitconfig, and overriding anything set there?

dandavison:
@HaleTom can you expand on that -- what's an example of a problem that would be solved by having a second config file?

dandavison:
@HaleTom I think I see what you're suggesting -- we could have a second config file containing an alternative config that we sometimes need to use as a variant, overriding values in the main config file. I think that we actually already have that: that is essentially what [named custom "features"](https://github.com/dandavison/delta#custom-features) are. There are already precedence rules built into the features implementation, so they allow overriding in a predictable way. And we can define as many variant configs as we want as named features. The only question seems to be -- how do we select our features conveniently at the command line (which would also be a question for your suggestion of having a second config file, right?)

I am thinking that we want this to be possible without having to pipe git into delta to apply a command-line arg, in which case an environment variable seems like the way forwards, since that will affect e.g. `git diff` without needing to explicitly invoke a delta process.

So as a third variant of the above environment variable ideas, I wonder whether we only need one new environment variable: `DELTA_FEATURES`. This would allow us to activate a named custom feature (with higher priority than any features already named in gitconfig). Then we can do anything we want in the named feature sections, including disabling boolean flags like `line-numbers`. I'm thinking this might make sense because in practice, each user will only have a small number of variants that they need, and so those variants can be hard-coded in gitconfig and given a convenient name.

dandavison:
Just giving that a very quick go I see that a bit of work might be required: the following would ideally have worked first time but did not:

```gitconfig
[delta]
    line-numbers = true
    features = no-line-numbers

[delta "no-line-numbers"]
    line-numbers = false
```

HaleTom:
You got me, @dandavison .  Cheers!

ulwlu:
@dandavison Hi shall I check that behaviour?

> the following would ideally have worked first time but did not

I actually found this bug when fixing git add pager, because some styles like hunk_header_style breaks pager while hunk_header_style in features won't break it.

I totally forgot where it happens, but maybe can fix.

dandavison:
> @dandavison Hi shall I check that behaviour?

@ulwlu , thanks! Yes, that would be great. I opened #446 for it. In addition to #307 I think that `DELTA_FEATURES` might be a solution for #447 also (as you know), so it would be great to get it working exactly as expected for all options.

ulwlu:
Okay! I'm gonna start working this in a few days because I need take a surgery today.

dandavison:
Awesome, no hurry! I hope that goes smoothly and you're feeling good after.

Issue #339, opened 2020-09-26
🐛 Delta crash on Windows when using as stand-alone diff tool


frifix:
For me delta version 0.4.3 crash on Windows 10 when I compare two files with command: `delta file_1 file_2`
Here is a backtrace:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "Nie można odnaleźć określonego pliku." }', src/main.rs:117:37
stack backtrace:
   0:     0x7ff624923cce - git_filter_source_path
   1:     0x7ff624940b1c - git_filter_source_path
   2:     0x7ff62491f553 - git_filter_source_path
   3:     0x7ff6249271eb - git_filter_source_path
   4:     0x7ff624926e38 - git_filter_source_path
   5:     0x7ff6249279bf - git_filter_source_path
   6:     0x7ff62492753f - git_filter_source_path
   7:     0x7ff62493f3b0 - git_filter_source_path
   8:     0x7ff62493f1e3 - git_filter_source_path
   9:     0x7ff6247abc1e - <unknown>
  10:     0x7ff6247b9ed2 - <unknown>
  11:     0x7ff624927bb6 - git_filter_source_path
  12:     0x7ff6247b3b17 - <unknown>
  13:     0x7ff624a2d704 - onig_unicode_define_user_property
  14:     0x7ffdd9df7bd4 - BaseThreadInitThunk
  15:     0x7ffddbb0ce51 - RtlUserThreadStart
```

frifix:
After installing `difftools` on Windows it works. However I don't know if the crash is an intended behavior in this situation.

dandavison:
Thank you, yes I think we can handle this a bit more gracefully!

PR #341, opened 2020-10-01
Handle diff execution error gracefully.

yoichi:
Output following message instead of panic when diff command is not exist:

```
Failed to execute the command: diff: No such file or directory (os error 2)
```

fixes https://github.com/dandavison/delta/issues/339


dandavison:
Thanks very much @yoichi!

PR #344, opened 2020-10-08
Fix whitespace in Makefile

Minnozz:
Replace 4 spaces with a tab.

Was going to work on Subversion support, but this was the first thing I noticed.


dandavison:
Right, thanks!

PR #347, opened 2020-10-09
Added the command to install with cargo

ZloeSabo:
The command to install delta with cargo was missing, this change adds it.


dandavison:
Thanks @ZloeSabo!

PR #367, opened 2020-10-27
Fix diff interactivefilter when decoration option is enabled

ulwlu:
Fix https://github.com/dandavison/delta/issues/320#issuecomment-715521903

When ***-decoration-style is enabled, it occurs error. So I set them to none while color-only.

Also, I added all configs to test(excluding `show-config`, `show-syntax-theme`, or something not related).

### DIFF
--- a/src/options/set.rs
+++ b/src/options/set.rs
@@ -192,6 +192,8 @@ pub fn set_options(
     // there (does not emit lines in 1-1 correspondence with raw git output). See #274.
     if opt.color_only {
         opt.side_by_side = false;
+        opt.file_decoration_style = "none".to_string();
+        opt.commit_decoration_style = "none".to_string();
         opt.file_style = "raw".to_string();
         opt.commit_style = "raw".to_string();
         opt.hunk_header_style = "raw".to_string();
--- a/src/tests/test_example_diffs.rs
+++ b/src/tests/test_example_diffs.rs
@@ -838,16 +838,16 @@ src/align.rs
     #[test]
     fn test_color_only_output_is_in_one_to_one_correspondence_with_input() {
         let user_suppliable_configs: &[&[&str]] = &[
-            &["--color-only", "--light", "true"],
-            &["--color-only", "--dark", "true"],
-            &["--color-only", "--line-numbers", "true"],
-            &["--color-only", "--side-by-side", "true"],
-            &["--color-only", "--diff-highlight", "true"],
-            &["--color-only", "--diff-so-fancy", "true"],
-            &["--color-only", "--navigate", "true"],
-            &["--color-only", "--hyperlinks", "true"],
-            &["--color-only", "--keep-plus-minus-markers", "true"],
-            &["--color-only", "--raw", "true"],
+            &["--color-only", "--light"],
+            &["--color-only", "--dark"],
+            &["--color-only", "--line-numbers"],
+            &["--color-only", "--side-by-side"],
+            &["--color-only", "--diff-highlight"],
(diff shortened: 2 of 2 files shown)


src/tests/test_example_diffs.rs
```diff
@@ -837,38 +837,115 @@ src/align.rs
...
+                "--hunk-header-decoration-style",
+                "bold yellow box ul",
+            ],
+            &[
+                "--color-only",
+                "--line-numbers",
+                "true",
+                "--line-numbers-minus-style",
+                "#444444",
+            ],
```
ryuta69, line 883: Apparently, these `--line-numbers-***` is not necessary because `line-numbers` would be disabled. However, I can't make sure if the option's behavior will be changed, so I added anyway.

It's just my opinion.

src/options/set.rs
```diff
@@ -192,6 +192,8 @@ pub fn set_options(
     // there (does not emit lines in 1-1 correspondence with raw git output). See #274.
     if opt.color_only {
         opt.side_by_side = false;
+        opt.file_decoration_style = "none".to_string();
+        opt.commit_decoration_style = "none".to_string();
```
dandavison, line 196: Hi @ryuta69, thanks for this PR! I have one question: aren't these already set to "none" by the `color-only` feature? (see the way that feature "inherits" from the `raw` feature)

ryuta69: @dandavison 
Thank you for review!
Oh, is that supposed to be? If so, currently it seems not working when a user supply some options in gitconfig.

In order to give priority to the options in features/** even when a user supply option in gitconfig, I think we must switch the conditions below.

https://github.com/dandavison/delta/blob/1ca06156570914b85875f95ab6ae1dcf685f082c/src/options/get.rs#L54-L72

Currently, both not working as you suggested.

```
[delta]
	file-decoration-style = ul
# git diff | delta --color-only
# => It still prints file-decoration-style as ul.

------------

[delta]
	minus-style = blue
# git diff | delta --raw
# => It still prints minus-style as blue.
```

dandavison: Hi @ryuta69, sorry to be slow -- I was meaning to look into what we're discussing here but didn't find time yet. Definitely looking forward to having this PR in the next release.

ryuta69: @dandavison 
Sure! The first priority is your schedule, and I'll help when you have a time. Again, thank you for this great OSS.

dandavison: OK, this sounds like a potential headache -- let's just keep the hard-coded override to "none" that you have, and we can maybe worry in the future about whether the none-value was supposed to be inherited from the feature automatically.

dandavison (approved):
LGTM, thanks @ulwlu!

I pushed one commit a39b58cd61793b3d611b5ad16a232511008674ae which removed the `true` arguments to the boolean flags, and deleted one dup test case. I believe that that "true" was not actually being used as an option value but rather was being interpreted as the first of two possible file name positional args. E.g. `gs | delta --light true true` "works" but `gs | delta --light true true true` fails with `error: Found argument 'true' which wasn't expected, or isn't valid in this context`. Let me know if anything in a39b58cd61793b3d611b5ad16a232511008674ae doesn't seem right to you!

src/options/set.rs
ulwlu: @dandavison 
I understand, thank you for review.
If I find a way to solve, I'm going to create PR.

ulwlu:
@dandavison 
Oh! I didn't notice there is that kind of case.
thank you very much!

PR #372, opened 2020-10-30
replace nonexistent option in `--help` output

Kr1ss-XD:
Recently, I noticed that  the 'after_help' text mentions a `number` op-
tion, which doen't seem to be implemented.

Let's fix this.


dandavison:
Oh thanks very much @Kr1ss-XD. (Yes, when implementing that feature @clnoll and I debated some different names and that was one of them).

Issue #378, opened 2020-11-04
🐛 Delta spawns too many threads, causing thrasing


jyn514:
```
$ git diff HEAD~ --stat
 Cargo.lock                                            |  15 +++++++++++++-
 library/alloc/tests/slice.rs                          | 120 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 library/backtrace                                     |   2 +-
 src/bootstrap/tool.rs                                 |   1 +
 src/doc/book                                          |   2 +-
 src/doc/edition-guide                                 |   2 +-
 src/doc/nomicon                                       |   2 +-
 src/doc/reference                                     |   2 +-
 src/doc/rust-by-example                               |   2 +-
 src/librustdoc/passes/collect_intra_doc_links.rs      |   3 +--
 src/llvm-project                                      |   2 +-
 src/test/ui/array-slice-vec/subslice-patterns-pass.rs | 126 ---------------------------------------------------------------------------------------------------------------
 src/tools/cargo                                       |   2 +-
 src/tools/miri                                        |   2 +-
 14 files changed, 145 insertions(+), 138 deletions(-)
$ PAGER= time -v git diff HEAD~ >/dev/null
	Command being timed: "git diff HEAD~"
	User time (seconds): 0.09
	System time (seconds): 0.18
	Percent of CPU this job got: 234%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.11
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 33516
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 9333
	Voluntary context switches: 131
	Involuntary context switches: 22
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
$ PAGER=delta time -v git diff HEAD~
... still running as I write this issue ...
```

This is https://github.com/rust-lang/rust/commit/1bdee96c5e6de445f09df34447a42553294f21ed. ltrace shows that there are hundreds of delta processes running:

```
$ ps -ef | grep delta | wc -l
996
```

jyn514:
[...]

Issue #326, opened 2020-09-18
Not picking up bat configuration file


ahmedelgabri:
As far as I understand, `delta` uses `bat` for its syntax highlighting, etc… And `bat` supports having a config file (`bat --config-file /path/to/file` or by setting `BAT_CONFIG_PATH`), I'm interested in the second case `BAT_CONFIG_PATH`.

In this file you can have some file mapping like this for example:

```
--map-syntax '.gitconfig.local:Git Config'
--map-syntax '.zsh*:Bourne Again Shell (bash)'
--map-syntax '.vimrc.local:VimL'
--map-syntax 'vimrc:VimL'
```

But it seems like this is not being picked up by the `bat` version used by `delta`

dandavison:
Hi @ahmedelgabri, thanks for this, but delta doesn't use bat for syntax highlighting. So, currently at least, delta doesn't look at the bat config file. So, perhaps this is best turned into a feature request for `map-syntax` functionality in delta? I'll have to think about whether we should go down the path of honoring the bat config `map-syntax` entries. Although in practice I know one would almost always want the same mappings in both, it may not make sense to link the two applications like that -- what do you think? There is one point of shared configuration at the moment -- delta honors the `BAT_THEME` env var.

ahmedelgabri:
If delta honors `BAT_THEME` already why not honor the config too. The config can actually contain the theme too.

ahmedelgabri:
Unless you plan not to use bat, I don’t see why you shouldn’t pick up it’s config. Yes it will couple the projects together but bat can be used as a library already so it’s not like you are going to be tied to an external app. It can be a direct dependency of the project which you can swap later if you want.

nickspoons:
I found this confusing too, when trying to work out how to map my `~/.vim/vimrc` syntax to `VimL`, which worked as expected for `bat ~/.vim/vimrc` but not `git diff ~/.vim/vimrc`.

+1 for some form of `map-syntax` for delta. I don't understand where delta is using bat and where it is not, as a new user it _seems_ like the bat config file would have been used here. Clearly not, but if parsing the bat config file is an option, that might make things "just work" for a lot of users. For 2 of us, at least :smile:

SuperCuber:
> I don't understand where delta is using bat and where it is not, as a new user it seems like the bat config file would have been used here

[...]

Issue #116, opened 2020-03-22
Make use of bat as a library


dandavison:
See https://github.com/sharkdp/bat/releases/tag/v0.13.0

Currently, delta vendors some bat code, e.g. related to paging and custom themes/syntax definitions.

MarcoIeni:
I will try to work on this in my spare time. Let's see what I can do :)

dandavison:
Hi @MarcoIeni, that sounds great!

MarcoIeni:
Started with some clippy warnings fix. See #380 

Anyway I think that the best strategy might be to keep the `bat/` directory and incrementally substitute parts of its code with the bat external library.

My first approach was to remove the `bat/` directory entirely and try to replace it with the bat library, but I failed because the library doesn't do everything.
In fact, there will be some code that will remain in the `bat/` directory, because the bat library doesn't expose everything (for example the function `bat::assets::list_languages`).

So we will have both `use crate::bat` (the `bat/` dir) and `use bat` (the actual bat library added in cargo.toml) mixed in the code. I will try to keep things clean as much as possible. In the future we might think to remane the `bat/` dir to something like `bat_utils` to avoid confusion.

dandavison:
Yes, that all makes sense. I was aware that some things aren't exposed, and also that there are some delta-specific changes in `src/bat/output.rs` that may make it difficult to avoid code duplication there?

MarcoIeni:
From what I understood this is the function that prints to output:
https://github.com/dandavison/delta/blob/b14942527ac7f839f16523f7fe390fe922c20553/src/paint.rs#L438-L442

So here I should call the bat `print()` function as shown in the docs [example](https://docs.rs/bat/0.16.0/bat/).

dandavison:
Hi @MarcoIeni, I'm worried that this ticket is misleading and/or that I've been slow to give input! Delta cannot use bat for syntax highlighting. The places where I think delta may be able to use bat are related to (a) handling of the sublime syntax and theme binary assets that the bat project extremely helpfully makes available, and (b) handling of the pager process that delta, just like bat, spawns to receive its output. So that is the code that is vendored under the `bat/` directory, that originated as a licensed direct copy from bat's source, and which has diverged slightly since. I am not sure whether the bat library currently exposes the necessary functionality for these; if not, that wouldn't be terrible as it's only a small part of the delta code base.

[...]

PR #379, opened 2020-11-05
Document configuration via environment

Kr1ss-XD:
As described in the discussion of #378, certain constellations of envi-
ronment variable configuration may lead to an endless recursion cycle.

This adds a note  to `README.md`, and to the application's help output,
to make users aware of that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,9 @@ Behind the scenes, delta uses `less` for paging. The version of `less` that come
 
 ## Configuration
 
+
+#### Git's config files
+
 Set delta to be git's pager in your `.gitconfig`. Delta has many options to alter colors and other details of the output. An example is
 ```gitconfig
 [core]
@@ -293,6 +296,27 @@ delta a.txt b.txt
 
 diff -u a.txt b.txt | delta
 ```
+
+#### Environment
+
+##### a) Delta
+Delta respects the setting of some environment variables to derive the default pager to use. These are in descending order of priority :
+- `DELTA_PAGER`,
+- `BAT_PAGER` and
+- `PAGER`
+
+Hence if e.g. `$DELTA_PAGER` is unset, delta will use `$BAT_PAGER`, or `$PAGER` if that is unset, too. In environments wthout any of the three variable set, delta's fallback is `less`.
+
+_**NOTE:** Be aware that setting one of those environment variables to `delta` while leaving the ones with higher priority unset (or also pointing at `delta`) can lead to an infinite cycle of forked processes by delta recursively calling itself._
+
+##### b) Third party apps
+Note that `$BAT_PAGER` (as well as `$PAGER`) is _also_ used by the standalone `bat` app. Please see `bat(1)` for a description. However, `$BAT_THEME`, `$BAT_STYLE` and `$BAT_CONFIG_PATH` are **not** regarded by delta.
+
+The behavior of delta's default pager, less, can be controlled using the `LESS` environment variable. It may contain any of the `less` command line options and/or interactive less-commands (prefixed by a leading `+` sign) which will be executed every time right after less is launched. A real-life example for the latter would be `LESS='+Gg'`, causing the pager to briefly jump to the bottom of the buffer and back to the top right afterwards. The purpose of this example is to enable the full metadata display in less' status line, including the buffer contents length and relative position of the current view port, in percent and lines/characters.
+
+Less also uses a bunch of other environment variables as well as config files, although most of them are not relevant for use in combination with delta. For an in-depth explanation of these configuration options please see the `less(1)` manual.
+
+It should also be mentioned that, besides using the `core.pager` setting, you can also set `GIT_PAGER=delta` in your environment in order to define the default `git` pager. If set, this variable will take priority above the setting in the git config file(s).
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -729,7 +729,8 @@ OPTIONS:
         --paging <paging-mode>
             Whether to use a pager when displaying output. Options are: auto, always, and never. The default pager is
             `less`: this can be altered by setting the environment variables DELTA_PAGER, BAT_PAGER, or PAGER (and that
-            is their order of priority) [default: auto]
+            is their order of priority; *NOTE* be aware that setting one of those environment variables to `delta` while
+            the higher prioritized ones are unset will cause delta to endlessly fork itself) [default: auto]
         --minus-empty-line-marker-style <minus-empty-line-marker-style>
             Style for removed empty line marker (used only if --minus-style has no background color) [default: normal auto]
         --plus-empty-line-marker-style <plus-empty-line-marker-style>
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ Delta respects the setting of some environment variables to derive the default p
 Hence if e.g. `$DELTA_PAGER` is unset, delta will use `$BAT_PAGER`, or `$PAGER` if that is unset, too. In environments without any of the three variable set, delta's fallback is `less`.
 
 ##### b) Third party apps
-Note that `$BAT_PAGER` (as well as `$PAGER`) is _also_ used by the standalone `bat` app. Please see `bat(1)` for a [description](https://jlk.fjfi.cvut.cz/arch/manpages/man/community/bat/bat.1.en). However, `$BAT_THEME`, `$BAT_STYLE` and `$BAT_CONFIG_PATH` are **not** regarded by delta.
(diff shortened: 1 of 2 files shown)


dandavison:
Thanks @Kr1ss-XD! You know, I think it would be even better to add a new section in the README, introducing the fact that delta starts a separate pager process and saying something about the pager-related env vars and options that are relevant (e.g. `DELTA_PAGER`, `BAT_PAGER`, `PAGER`, and also `LESS`). I think it's especially important because `less` is the cause of quite a few problems for people (especially on Windows). Do you feel like taking a stab at that? It doesn't have to be perfect -- documentation can always evolve. If not, we can merge this and do it later.

Kr1ss-XD:
Sure, I've actually considered something like that but was hesitant to introduce too much of a chance.

I'm going to add a draft tonight, after work. 🙂

Kr1ss-XD:
I'd greatly appreciate a review regarding linguistical (and content-wise) correctness, if you can spare some time for that.

Thx

README.md
```diff
@@ -294,6 +299,27 @@ delta a.txt b.txt
...
+##### a) Delta
+Delta respects the setting of some environment variables to derive the default pager to use. These are in descending order of priority :
+- `DELTA_PAGER`,
+- `BAT_PAGER` and
+- `PAGER`
+
+Hence if e.g. `$DELTA_PAGER` is unset, delta will use `$BAT_PAGER`, or `$PAGER` if that is unset, too. In environments wthout any of the three variable set, delta's fallback is `less`.
+
+_**NOTE:** Be aware that setting one of those environment variables to `delta` while leaving the ones with higher priority unset (or also pointing at `delta`) can lead to an infinite cycle of forked processes by delta recursively calling itself._
+
```
dandavison, line 313: OK I hope this doesn't sound like I'm being difficult :)  but I don't think we need to mention this in the README as it's an unlikely scenario and doesn't correspond to any recommended usage. If it turns out to be a problem that users encounter in practice, then we should just put a fix in the code so that delta refuses to call itself.

README.md
```diff
@@ -294,6 +299,27 @@ delta a.txt b.txt
...
 
+#### Environment
+
+##### a) Delta
+Delta respects the setting of some environment variables to derive the default pager to use. These are in descending order of priority :
+- `DELTA_PAGER`,
+- `BAT_PAGER` and
+- `PAGER`
+
+Hence if e.g. `$DELTA_PAGER` is unset, delta will use `$BAT_PAGER`, or `$PAGER` if that is unset, too. In environments wthout any of the three variable set, delta's fallback is `less`.
```
dandavison, line 310: typo: wthout

README.md
```diff
@@ -294,6 +299,27 @@ delta a.txt b.txt
...
+
+_**NOTE:** Be aware that setting one of those environment variables to `delta` while leaving the ones with higher priority unset (or also pointing at `delta`) can lead to an infinite cycle of forked processes by delta recursively calling itself._
+
+##### b) Third party apps
+Note that `$BAT_PAGER` (as well as `$PAGER`) is _also_ used by the standalone `bat` app. Please see `bat(1)` for a description. However, `$BAT_THEME`, `$BAT_STYLE` and `$BAT_CONFIG_PATH` are **not** regarded by delta.
```
dandavison, line 313: I think the word you were looking for was "respected". But, actually, shall we just say "used by" -- that makes it easy to understand; after all the audience has many (most?) non-native English speakers.

README.md
```diff
@@ -294,6 +299,27 @@ delta a.txt b.txt
 diff -u a.txt b.txt | delta
 ```
 
+#### Environment
+
+##### a) Delta
+Delta respects the setting of some environment variables to derive the default pager to use. These are in descending order of priority :
+- `DELTA_PAGER`,
+- `BAT_PAGER` and
```
dandavison, line 307: What's your opinion about the way delta respects `BAT_PAGER`? Maybe we should start saying that it is deprecated and delta users should use `DELTA_PAGER` or `PAGER`. I may have introduced some confusion with all the connections to bat; see e.g. https://github.com/dandavison/delta/issues/326#issuecomment-718708442

README.md
```diff
@@ -294,6 +299,27 @@ delta a.txt b.txt
 diff -u a.txt b.txt | delta
 ```
 
+#### Environment
+
+##### a) Delta
+Delta respects the setting of some environment variables to derive the default pager to use. These are in descending order of priority :
```
dandavison, line 305: nit: "decide which pager to use" 

I guess the "default" is actually "less" with some default options.

README.md
```diff
@@ -294,6 +299,27 @@ delta a.txt b.txt
...
+
+_**NOTE:** Be aware that setting one of those environment variables to `delta` while leaving the ones with higher priority unset (or also pointing at `delta`) can lead to an infinite cycle of forked processes by delta recursively calling itself._
+
+##### b) Third party apps
+Note that `$BAT_PAGER` (as well as `$PAGER`) is _also_ used by the standalone `bat` app. Please see `bat(1)` for a description. However, `$BAT_THEME`, `$BAT_STYLE` and `$BAT_CONFIG_PATH` are **not** regarded by delta.
+
+The behavior of delta's default pager, less, can be controlled using the `LESS` environment variable. It may contain any of the `less` command line options and/or interactive less-commands (prefixed by a leading `+` sign) which will be executed every time right after less is launched. A real-life example for the latter would be `LESS='+Gg'`, causing the pager to briefly jump to the bottom of the buffer and back to the top right afterwards. The purpose of this example is to enable the full metadata display in less' status line, including the buffer contents length and relative position of the current view port, in percent and lines/characters.
+
```
dandavison, line 316: > A real-life example for the latter would be `LESS='+Gg'`, causing the pager to briefly jump to the bottom of the buffer

Is that something you use with delta?

README.md
Kr1ss-XD: NP :smile:

README.md
```diff
@@ -294,6 +299,27 @@ delta a.txt b.txt
...
+Less also uses a bunch of other environment variables as well as config files, although most of them are not relevant for use in combination with delta. For an in-depth explanation of these configuration options please see the `less(1)` manual.
```
dandavison, line 317: It would be really nice to link to an online less manual here. Honestly, I consider myself decent with Unix command line stuff and I've never learned how to use `man` or what the numbers really mean. And we do have Windows users of course.

dandavison (approved):
Great! I made some suggestions but we can always improve the docs whenever so make just make the changes you feel like.

README.md
Kr1ss-XD: I couldn't find one at the upstream site, but I guess we can link to linux.die.net ?

README.md
Kr1ss-XD: Not specifically, but I do use it in my default `less` config.

README.md
Kr1ss-XD: +1

I'd actually think it would be good if you'd abandon the use of the variable at some time.

Kr1ss-XD:
Nice !

Thank you very much for the review. I'll call it a day and will take care of your points tomorrow. Need to get get up soon :(  See you

README.md
dandavison: >  I guess we can link to linux.die.net 

Yes, I agree that's a well-known resource. Maybe we don't need detail here on `LESS` if we can link to good external docs (I'd even link to a random article / blog post on configuring less if there were a good one, seeing as there are so many possibilities).

README.md
dandavison: Thanks for the second opinion there.

OK, so maybe let's remove all mention of `BAT_PAGER` from the main instructions and just have a footnote / final sentence somewhere saying something like

> Delta does currently respect the `BAT_PAGER` environment variable, but that is deprecated: please use `DELTA_PAGER` or `PAGER` instead.

dandavison: cc @MarcoIeni ref #116 it doesn't necessarily affect your refactoring work but just wanted to mention to you that we're planning on making delta a bit more independent of bat by deprecating the usage of the `BAT_PAGER` env var to configure delta's pager.

README.md
Kr1ss-XD: Sure. Guess I was trying to prevent repetition, as i already said 'used by' in the first sentence.

dandavison:
Let me know when you're happy for me to merge this @Kr1ss-XD. Like I said, we can always merge first and tweak later, so no pressure.

Kr1ss-XD:
If you're confident with it, I'd say go forward. :)

dandavison:
Great! Thanks very much for doing this.

Kr1ss-XD:
No. Thank **you** :exclamation:

PR #380, opened 2020-11-05
fix some clippy warnings

MarcoIeni:
I fixed 60+ clippy warnings. 40 are left, but I stopped because I think it is already too much for a single PR.


dandavison:
Awesome. Thanks very much @MarcoIeni! The fact I hadn't done this was not an indication that I don't appreciate the sophisticated linting that's available; I really had been meaning to do this for a while. I went through all these changes and my reaction was almost entirely "Oh that's great that it can be simplified like that."

Of course, when we're done clippifying, I'd like to have a test (or make target called by the `make test` target) to enforce it forever.

Issue #384, opened 2020-11-07
🚀 Add a test that enforces clippy and rustfmt compliance


dandavison:
Once the codebase is clippy-compliant, let's add a test that ensures it stays that way. (And the same for rustfmt).  cc @MarcoIeni ref #380 #383

MarcoIeni:
I know how to do it by using github actions. See [this](https://github.com/rust-github/template/blob/master/.github/workflows/ci.yml).
I don't know if you have plans to use them.

dandavison:
I'm happy to switch to GitHub actions if that makes sense. But I'm thinking that's pretty much orthogonal to this issue since we can add shell commands here https://github.com/dandavison/delta/blob/b14942527ac7f839f16523f7fe390fe922c20553/etc/ci/script.sh#L10-L15

MarcoIeni:
> I'm happy to switch to GitHub actions if that makes sense.

I don't know, that's just what I happen to use for my open source projects, but yes, it's a separate issue.

Anyway, the commands should be the following:
```
cargo fmt --all -- --check
cargo clippy -- -D warnings
```

dandavison:
Update: we've got a check for rustfmt, and a check for clippy commented out; the only remaining thing here is getting the codebase to pass clippy.

dandavison:
Fixed at #422

PR #393, opened 2020-11-22
ci: verify code is well formatted

MarcoIeni:
Partially solves #384 

Downside: In this way `cargo fmt` runs for every target, but it could be run only once.


dandavison:
Thanks! I verified this passes on master (since travis builds are not running currently: #396)

PR #395, opened 2020-11-22
implement From for Vec<Element>

MarcoIeni:
This solves a clippy warning: to_vec doesn't respect naming conventions, because it consumes self (i.e. the function doesn't take `&self`, but `self`).

Anyway, in my opinion a much simpler approach would be just to delete `to_vec` without implementing `From` and to use just `collect` like this in tests:
```rust
        let actual_vec: Vec<Element> = AnsiElementIterator::new(minus_line).collect();
        assert_eq!( actual_vec,  vec![..]);
```

### DIFF
--- a/src/ansi/iterator.rs
+++ b/src/ansi/iterator.rs
@@ -38,6 +38,12 @@ pub enum Element {
     Text(usize, usize),
 }
 
+impl From<AnsiElementIterator<'_>> for Vec<Element> {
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks again @MarcoIeni. Yes, I agree your simpler option is the best -- especially since the conversion is only used in tests.

MarcoIeni:
Closed in favor of #397

Issue #317, opened 2020-09-07
🚀 Using delta as library


extrawurst:
This library looks great! Is it possible to use it as a crate in another rust project? `gitui` could really benefit from such state of the art diff visualisation

Kr1ss-XD:
Nice one !  That would be great.

dandavison:
Hi @extrawurst, thanks! Delta-as-a-library isn't possible yet, but shall we sketch out what the minimal version of this could look like? The delta test suite already contains a function that takes an input diff as a string, and a delta config struct, and returns delta output as a string. So, as a quick first thought, If we exposed functions like these, is there any chance that that could suffice as a minimal first version of the library?
  https://github.com/dandavison/delta/blob/6912b4df9507ca43fdcacbaf4b9e07fdb6a26646/src/tests/integration_test_utils.rs#L74-L85

https://github.com/dandavison/delta/blob/6912b4df9507ca43fdcacbaf4b9e07fdb6a26646/src/tests/integration_test_utils.rs#L51-L53

extrawurst:
@dandavison ok I might need more context here. what kind of `input` would that require? and what output does it generate?

dandavison:
The input would be a string containing the stdout of a git diff command (`git diff`, `git show`, `git log -p`, `git stash show -p` etc) or `diff -u`. The output would be the same as the input, but with ANSI color escape sequences inserted to effect delta's coloring, and possibly with  additional decoration elements (boxes, lines, reconfigured hunk header). The details of exactly which colors and decoration elements would be applied to the input would be determined by the `config` struct. Basically every option mentioned in the delta README would be configurable by setting a value on the `config` struct.

dandavison:
Incidentally, lazygit already uses delta, but as I understand it lazygit calls git as an external process whereas you are using libgit2. Is that right? So lazygit didn't need delta as a library; a delta process is just inserted in between git's output and lazygit's input.

dandavison:
More precisely, I think the input might be three things:

1. A string containing git's diff output
2. A [`git2::Config`](https://docs.rs/git2/0.13.10/git2/struct.Config.html) struct representing the user's git config (which might have relevant delta config).
3. Some additional delta command line options (optional)

So my question at this stage is: would an API something like this give you what you need?

```rust
/// Run delta on `git_output` (a string containing a diff in git/unified diff format), respecting
/// relevant configuration in `git_config` and `command_line_options`. Return a string containing
[...]

PR #398, opened 2020-11-22
Refactor to provide a library crate

dandavison:
@MarcoIeni this is a partial progress towards refactoring Delta as a library crate with a binary (#317).

This branch in its current form passes all tests and exposes a public library as well as a binary. I believe that, if this branch were published to crates.io, then delta would be available as a library. However, I don't want to actually do that right now, because the library API in this branch is not ready yet.  

What I would ideally like to do is merge this branch into master now (so the refactoring is in master), but _without_ actually publishing a library to crates.io for now, because this branch has not yet finalized the library API.

So I was thinking we could make one small change to this branch, something like

rename lib.rs -> delta_lib.rs

Does that make sense and can you suggest what change I should make?


dandavison:
We can chat in gitter by the way, if that's convenient: https://gitter.im/dandavison-delta/community

MarcoIeni:
Are you able to do that rename?
I tried it but modules declarations fail.
From what I've seen if you don't have a `lib.rs` you have to move those modules declaration inside `main.rs` again.

But maybe I am missing something.

MarcoIeni:
The simplest solution would be to keep everything as it is but place big warnings in the documentation where we say that everything is experimental and it is going to change really soon.

dandavison:
> you have to move those modules declaration inside main.rs again.

Thanks, maybe this is what I want to do -- create one commit that will basically be reverted when the library API is finalized.

I'd prefer not to have a library published to crates.io if it is not the correct API, even with warnings in the docs.

hnorkowski:
Any updates on this?

rmacklin:
Possibly superseded by https://github.com/dandavison/delta/pull/1013 and https://github.com/dandavison/delta/pull/1017

PR #400, opened 2020-11-22
ci: add end to end tests

MarcoIeni:
I left `cargo test` as it is instead of calling `make test` because in this way we use the `actions-rs/cargo` action instead of just calling `cargo` from the command line.
Main benefit of doing this:
> Warnings and errors issued by cargo will be displayed in GitHub UI


dandavison (approved):
LGTM!

dandavison:
Yes, thanks that sounds like the right way of doing this.

PR #411, opened 2020-11-28
GitHub Continuous Integration: run tests on linux, mac and windows

MarcoIeni:
[Open Source Saturday Italy](https://oss-italy.github.io/)

With this PR we add windows and mac to tests.
Integration tests are split from unit tests, because I think you can't run them on windows (and on other linux targets as well).
All other linux targets except `x86_64-unknown-linux-gnu` are still missing and will be added in the future.
We added `--release` flag to the last step of integration tests in order to avoid to compile the project in debug mode, tell us if there is a problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,14 +36,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        os: [macos-latest, ubuntu-latest]
         include:
           - os: macos-latest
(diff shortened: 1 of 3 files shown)


dandavison:
Fantastic, so this is now better than what we had on Travis. Thanks @MarcoIeni @surveyor3 @DanySpin97!

What do you think of the way the audits are blocking merge, do we need to tweak something there? Shall I go ahead and merge this?

image

MarcoIeni:
Test Suite is yellow because we deleted it.
I don't know why audits are yellow..They seem totally unrelated to this PR, so I would merge this 😅

PR #413, opened 2020-11-29
ci: add unit tests for other linux targets

MarcoIeni:
From my understanding by using [cross](https://github.com/rust-embedded/cross) we are able to cross-compile to all the targets we want, without executing `/etc/ci/before_install.sh` :exploding_head:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,32 +5,17 @@ jobs:
 
   unit_tests:
     name: Unit tests
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.job.os }}
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest, windows-latest]
-        include:
-          - os: macos-latest
-            target: x86_64-apple-darwin
-            use-cross: false
-          - os: windows-latest
-            target: x86_64-pc-windows-msvc
-            use-cross: false
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-gnu
-            use-cross: false
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-musl
-            use-cross: true
-          - os: ubuntu-latest
-            target: i686-unknown-linux-gnu
-            use-cross: false
-          - os: ubuntu-latest
-            target: arm-unknown-linux-gnueabihf
-            use-cross: true
-          - os: ubuntu-latest
-            target: aarch64-unknown-linux-gnu
-            use-cross: true
+        job:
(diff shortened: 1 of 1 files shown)


MarcoIeni:
Regarding this:
image

- Security audit should run only on push, why does it appear here? :/
- Test Suite was removed in #411. Why is it still checked?

dandavison:
I'm sure I'm going to like this, I just need to understand what's going on first :) Can you explain a bit more what should be happening with the test builds configured like this? This is quite unfamiliar territory for me so feel free to ELI5! For example, two questions I have are

- What is the logic behind the pattern of `true`s and `false`s in the `use-cross` entries?
- When I look at https://github.com/dandavison/delta/runs/1470532514, I just see a single invocation like this `cross test --target aarch64-unknown-linux-gnu --verbose`

and then there's a bunch of compilation, and finally the test suite is just run once I think. I was kind of expecting to see the test suite run multiple times?

MarcoIeni:
I just discovered [cross](https://github.com/rust-embedded/cross) a few hours ago, so it's new also for me :)

> What is the logic behind the pattern of trues and falses in the use-cross entries?

So in the include section I just set the use-cross variable to true or false, depending if I want to use cargo (false) or cross (true).

For example here on mac os I set this variable to false, because in this case I want to use cargo as usual:

https://github.com/MarcoIeni/delta/blob/c7a57674b0e85a16aac07c57c2e549cf95fd0c9d/.github/workflows/ci.yml#L13-L15

This variable is read by the cargo action:

https://github.com/MarcoIeni/delta/blob/c7a57674b0e85a16aac07c57c2e549cf95fd0c9d/.github/workflows/ci.yml#L48

If it is true, it will use `cross`, otherwise it will use `cargo` as usual. [Here](https://github.com/actions-rs/cargo#cross-compilation) you can find the docs related to this flag.

## ELI5 part

Cross compilation means to compile for a different target with respect to the machine you are using. For example in this case we want to compile from linux x86 to linux arm 64 bit.

From my understanding the `cross` tool, depending on the target compiles on a docker container which has all the correct dependencies. So for example, if you give him `aarch64-unknown-linux-gnu` as target, it will run cargo on a docker container which has all the dependencies and the setup required to build for that target, so you don't have to install them by yourself.

## embarrassing part

> I was kind of expecting to see the test suite run multiple times?

Me too honestly 😅

MarcoIeni:
Those green checks looks so sweet :sunglasses:

dandavison:
Awesome, and thanks for explaining!

PR #420, opened 2020-12-03
Fix variable assign with bash

ulwlu:
This caused make test be errored.

```
make: *** [end-to-end-test] Error 1
```

### Reason

As `man bash`

```
${parameter:-word}
Use Default Values. If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.

${parameter:=word}
Assign Default Values. If parameter is unset or null, the expansion of word is assigned to parameter. The value of parameter is then substituted. Positional parameters and special parameters may not be assigned to in this way.
```

This means, $1 cannot be assigned any variables. It should be - but =.

### DIFF
--- a/tests/test_deprecated_options
+++ b/tests/test_deprecated_options
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # if first arg provided, use it, otherwise default to release
-DELTA_BIN=${1:=./target/release/delta}
+DELTA_BIN=${1:-./target/release/delta}
 DELTA="$DELTA_BIN --no-gitconfig"
 
 foreground_color=red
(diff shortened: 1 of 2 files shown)


dandavison (approved):
LGTM! I'm sorry, we made exactly the same change!

dandavison:
This exact change was made in parallel in ed39e4c8137e9d760eee49ac346603354902eeb4.

ulwlu:
Oh, Really!?
We missed each other at same timing lol.

ulwlu:
> 18 minutes ago

I see...

ulwlu:
just memo

discussion goes to below.
https://github.com/dandavison/delta/pull/421

Issue #405, opened 2020-11-23
🐛 Some theme options are not working with `--color-only` enabled


renard:
When running `delta` in a standalone mode (no git) with `--color-only` to keep a *diff* / *patch* compatible patch some theming options such as `--file-style` or `--hunk-header-style` (same with `--hunk-header-decoration-style`) are not operational:

<img width="720" alt="Screenshot 2020-11-23 at 14 58 17" src="https://user-images.githubusercontent.com/383821/99971457-77ebf580-2d9d-11eb-8702-84a308993593.png">

<img width="718" alt="Screenshot 2020-11-23 at 14 59 53" src="https://user-images.githubusercontent.com/383821/99971482-7f130380-2d9d-11eb-8fb1-35fb50088c4c.png">

Is there a way to highlight file headers and hunks delimiter in  a specific colour? 

Diff file:

```diff
--- 1	2020-11-23 12:51:34.434055926 +0000
+++ 2	2020-11-23 12:52:06.242055415 +0000
@@ -11,7 +11,7 @@
                                      style plain --hunk-style plain`.
         --dark                       Use default colors appropriate for a dark terminal background. For more control,
                                      see the other color options.
-    -h, --help                       Prints help information
+    -h, --HELP                       Prints help information
         --highlight-removed          Apply syntax highlighting to removed lines. The default is to apply syntax
                                      highlighting to unchanged and new lines only.
         --keep-plus-minus-markers    Prefix added/removed lines with a +/- character, respectively, exactly as git does.
```

TIA.

dandavison:
Hi @renard, thanks for this! Short answer: yes, I think there are more improvements we can make to `--color-only`. We should be able to color all elements as long as we are not making any "structural" changes (e.g. decorations).

Until recently, there had been some bugs causing `--color-only`  to create invalid patch output, when used in combination with certain options. For many delta users this comes up when using `git add -p`. Thanks to the work of @ulwlu, these are mostly, or perhaps entirely, gone now. See e.g. #320. I think perhaps we can characterize this program of work as follows:

0. There were bugs relating to invalid patches under `--color-only`
1. We introduced much better test coverage and fixed bugs, but erring on the side of aggressively eliminating features from `--color-only` to ensure that a valid (if vanilla in terms of coloring) patch was produced.
2. Now that we have good test coverage we should be in a better position to revisit and allow color where possible without allowing any decorations/structural changes to the patch.

cc @ulwlu what do you think?

ulwlu:
[...]

PR #421, opened 2020-12-03
Fix end-to-end test CI

dandavison:
cc @MarcoIeni  @ulwlu 

This is an attempt to fix the end-to-end test CI. Currently, I believe that it is not failing even when it should fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
           command: build
           args: --target ${{ matrix.target }} --release
       - name: End to end tests
-        run: |
+        run: >-
           DELTA_BIN=target/${{ matrix.target }}/release/delta
           ./tests/test_raw_output_matches_git_on_full_repo_history $DELTA_BIN &&
           ./tests/test_deprecated_options $DELTA_BIN > /dev/null
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,10 +29,9 @@ jobs:
           command: build
           args: --target ${{ matrix.target }} --release
       - name: End to end tests
-        run: |
-          DELTA_BIN=target/${{ matrix.target }}/release/delta
-          ./tests/test_raw_output_matches_git_on_full_repo_history $DELTA_BIN
-          ./tests/test_deprecated_options $DELTA_BIN > /dev/null
+        run: >-
+          ./tests/test_raw_output_matches_git_on_full_repo_history target/${{ matrix.target }}/release/delta;
+          ./tests/test_deprecated_options target/${{ matrix.target }}/release/delta > /dev/null
       - name: Run executable
         run: cargo run --release --target ${{ matrix.target }} -- < /dev/null
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,9 +29,7 @@ jobs:
           command: build
           args: --target ${{ matrix.target }} --release
       - name: End to end tests
-        run: >-
-          bash -c "./tests/test_raw_output_matches_git_on_full_repo_history target/${{ matrix.target }}/release/delta &&
-                   ./tests/test_deprecated_options target/${{ matrix.target }}/release/delta > /dev/null"
+        run: make DELTA_BIN=target/${{ matrix.target }}/release/delta end-to-end-test
       - name: Run executable
         run: cargo run --release --target ${{ matrix.target }} -- < /dev/null
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,36 +3,6 @@ name: Continuous Integration
 
 jobs:
 
-  unit_tests:
-    name: Unit tests
-    runs-on: ${{ matrix.job.os }}
-    strategy:
-      matrix:
-        job:
-          - { os: macos-latest,   target: x86_64-apple-darwin,         use-cross: false }
-          - { os: windows-latest, target: x86_64-pc-windows-msvc,      use-cross: false }
-          - { os: ubuntu-latest , target: x86_64-unknown-linux-gnu,    use-cross: false }
-          - { os: ubuntu-latest,  target: x86_64-unknown-linux-musl,   use-cross: true }
-          - { os: ubuntu-latest,  target: i686-unknown-linux-gnu,      use-cross: true }
-          - { os: ubuntu-latest,  target: arm-unknown-linux-gnueabihf, use-cross: true }
-          - { os: ubuntu-latest,  target: aarch64-unknown-linux-gnu,   use-cross: true }
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-          target: ${{ matrix.job.target }}
-          profile: minimal
-          override: true
-      - name: Unit tests
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          use-cross: ${{ matrix.job.use-cross }}
-          args: --target ${{ matrix.job.target }} --verbose
-
   integration_tests:
     name: Integration tests
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           command: build
           args: --target ${{ matrix.target }} --release
       - name: End to end tests
-        run: make DELTA_BIN=target/${{ matrix.target }}/release/delta end-to-end-test
+        run: make end-to-end-test DELTA_BIN=target/${{ matrix.target }}/release/delta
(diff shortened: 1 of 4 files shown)


tests/test_deprecated_options
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # if first arg provided, use it, otherwise default to release
-DELTA_BIN=${1:-./target/release/delta}
+DELTA_BIN=${1:=./target/release/delta}
```
dandavison, line 4: This change is here to make it fail deliberately; I will revert this before merging.

tests/test_raw_output_matches_git_on_full_repo_history
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # if first arg provided, use it, otherwise default to release
-DELTA_BIN=${1:-./target/release/delta}
+DELTA_BIN=${1:=./target/release/delta}
```
dandavison, line 4: This change is here to make it fail deliberately; I will revert this before merging.

.github/workflows/ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       - name: End to end tests
         run: |
           DELTA_BIN=target/${{ matrix.target }}/release/delta
-          ./tests/test_raw_output_matches_git_on_full_repo_history $DELTA_BIN
+          ./tests/test_raw_output_matches_git_on_full_repo_history $DELTA_BIN &&
           ./tests/test_deprecated_options $DELTA_BIN > /dev/null
```
dandavison, line 65: I thought that this fix was correct, but the build doesn't look like I was expecting: https://github.com/dandavison/delta/runs/1493996100

.github/workflows/ci.yml
```diff
@@ -59,9 +59,9 @@ jobs:
           command: build
           args: --target ${{ matrix.target }} --release
       - name: End to end tests
-        run: |
+        run: >-
```
dandavison, line 62: https://stackoverflow.com/questions/15540635/what-is-the-use-of-the-pipe-symbol-in-yaml
https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-over-multiple-lines

ulwlu:
@dandavison 
This is my suggestion.

## Ideal

**end-to-end** can be executed **with or without** passing bin-path.

## What's the problem currently

**end-to-end** **cannot** be executed **without** passing bin-path.

## What this PR aims, and what it should be

This PR aims to fix CI to be error when it should be. 

However, I think CI's alright already. 
It will fail when `test` is failed. On the other hand, bash syntax is correct for CI, It's alright that CI won't error. CI passes bin-path, so it's supposed to be pass.

So, there's no need to fix ci since we fix end-to-end.

dandavison:
> However, I think CI's alright already.

I'm not sure I agree -- with the latest commit 462fc56cf4557137cfb594d81a6e6cc2e63644cc CI fails as expected: 

https://github.com/dandavison/delta/runs/1494106155
image

but previously it was passing when it should have failed:

https://github.com/dandavison/delta/runs/1493996100

.github/workflows/ci.yml
dandavison: I believe the combination of the two fixes above is what we want.

ulwlu:
@dandavison 
I will look into previous succeeded one, but currently it fails because it should be like

```
# ; is required
DELTA_BIN=target/x86_64-apple-darwin/release/delta; ./tests/test_raw_output_matches_git_on_full_repo_history $DELTA_BIN && ./tests/test_deprecated_options $DELTA_BIN > /dev/null
```

in my guess. But I will look into it more carefully.

ulwlu:
And previous one is supposed to succeed. So it's as we expected.

It passes $DELTA_BIN correctly, so `assigning variable to $1` won't occur at `${1:=./target/release/delta}`.

dandavison:
> `; is required in bash not like rust.`

@ulwlu I don't agree that `;` is required -- but you made me notice an error in my code. Do you agree with this:

`DELTA_BIN=foo ./do_this` means:

 Create a child process to run the executable "./do_this", and set the env var DELTA_BIN=foo in that child process.

Whereas `DELTA_BIN=foo; ./do_this` means:

Set `DELTA_BIN=foo` in the current shell process. Then run the executable "./do_this". But, notice that `DELTA_BIN=foo` will not be set in the child process.

Now, I had made a mistake! At ec18620beb1ee91773973350dbe400816821cc1d I had this:

`DELTA_BIN=foo ./do_this && ./do_that`. But that means:

Create child process 1 running executable "./do_this" with the env var set in that child process. Then, if that succeeds create child process 2 running executable "./do_that" but _without_ setting the env var.

So that wasn't what we wanted. I've switched it to use `make` now as I think we can rely on a suitable make being present in MacOS and Linux build envs, and this way we can add to the make target without having to duplicate that in the CI config.

ulwlu:
@dandavison 
I agree with that, if we want to execute in same process, we need to 'source'.

>I've switched it to use make now as I think we can rely on a suitable make being present in MacOS and Linux build envs, and this way we can add to the make target without having to duplicate that in the CI config.

Yes! I see it works fine.

BTW,
>　I don't agree that ; is required

This actually is, below is the example.

ss

An article about similar topic is here https://unix.stackexchange.com/questions/56444/how-do-i-set-an-environment-variable-on-the-command-line-and-have-it-appear-in-c.

dandavison:
> This actually is, below is the example.

Oh I see! Yes you're right, it was being passed as an argument. I forgot about that, and was thinking about the default assignment construct in the shell script.

dandavison:
> I thought `./{script}` is same as `source ./{script}`, so they should be in same process?

No, that's not true. `./script` runs an executable named "script" as a child process. The "executable" might happen to be a shell script with a shebang line (`#/bin/bash` or `#!/some/interpreter`) but that doesn't change things -- "some/interpreter" is started as a new child process and it is given the contents of the script.

```
~ cat print-pid
#!/bin/bash
echo $$
~ source print-pid 
5292
~ source print-pid
5292
~ ./print-pid
60932
~ ./print-pid
60954
~ bash ./print-pid 
60992
~ source print-pid
5292
```

ulwlu:
>~ cat print-pid
#!/bin/bash
echo $$
~ source print-pid 
5292
~ source print-pid
5292
~ ./print-pid
60932
~ ./print-pid
60954
~ bash ./print-pid 
60992
~ source print-pid
5292

@dandavison 
Oh sorry, you are correct.
I made mistake for the dot below.

```
# These are same.
source script.sh
. script.sh

# This creates new shell
./script.sh

```

dandavison:
Oh right, sorry, you already knew what I wrote :)

Yeah, incidentally I switched to zsh recently and I don't think the `. script` syntax exists.;l just `source`.

ulwlu:
> you already knew what I wrote :)

I completely forgot and learn again now thank to you (there is still a lot of things to learn for me... ).

With the fix, I can start moving again https://github.com/dandavison/delta/issues/405.
Thank you very much!

dandavison:
I haven't quite got a correct version of this. The problem I'm encountering is that I don't understand this error output. When this happens, the integration tests don't really run, but they also don't fail.

```
shell: /bin/bash -e {0}
```

dandavison:
OK, the current commit is the closest I've got to the behaviour we want with the deliberate error that's in this branch. It's need one more tweak though as I don't think it's setting the env var correctly in both processes. 

MarcoIeni:
Sorry I haven't read all the discussion in detail. Are you trying to setup an environment variable in a GitHub action step?

In this case that's how you do it:

https://github.com/dandavison/delta/blob/0ddd0648190f2d0497a7f3a9e89b26d04186feaf/.github/workflows/cd.yml#L53-L57

dandavison:
> Sorry I haven't read all the discussion in detail. Are you trying to setup an environment variable in a GitHub action step?

Haha don't worry -- this PR is completely incomprehensible :) 

The problem that it's addressing is that the end-to-end tests are passing in Github Actions CI even when there is an error in them. The error is fixed in master, but in this branch I'm trying to change things so the end-to-end tests will fail in Github Actions CI if an error is introduced again.

dandavison:
I'm wondering whether the problem is that `&&` is not being parsed correctly (it's in shell code in YAML). But we need to ensure that the test fails if either of those shell scripts exit with an error status. As far as I know, this will succeed if the first shell script fails and the second one succeeds:

```yaml
        run: |
          DELTA_BIN=target/${{ matrix.target }}/release/delta
          ./tests/test_raw_output_matches_git_on_full_repo_history $DELTA_BIN
          ./tests/test_deprecated_options $DELTA_BIN > /dev/null
```

MarcoIeni:
If we want to avoid using && we could try with `set -e`

dandavison:
Thanks @MarcoIeni. So looks like bash is the default on non-Windows builds, and by default they use `set -e` semantics, so what you had in the CI should be perfectly correct then. It's not currently failing when the scripts error but I'll get there eventually :)

PR #422, opened 2020-12-03
Disable some clippy warnings

dandavison:
This makes clippy pass, by whitelisting several clippy complaints about complex types and long argument lists. I think it might be more beneficial to have the clippy test running and open an issue to fix my poor code style that has been whitelisted. cc @MarcoIeni


MarcoIeni:
Great idea, I opened #424 in order to keep track of this issue :)

PR #431, opened 2020-12-05
Update to vte v0.9.0

dandavison:
Update to vte v0.9.0.

cc @chrisduerr just in case you have time to take a glance at this. If so, thanks very much! Delta uses vte to create an internal library of utilities for operating on strings containing ANSI escape sequences. I was originally using the [console](https://docs.rs/console/0.13.0/console/) crate for this, but I needed to handle OSC hyperlinks, which led me to switch to using vte's parser. (The code is messy currently, but on the positive side delta has lots of unit tests for these utilities).

See
    https://github.com/alacritty/vte/releases/tag/v0.9.0
    https://github.com/alacritty/vte/compare/v0.8.0...v0.9.0
    https://github.com/alacritty/alacritty/commit/576252294d09c1f52ec73bde03652349bdf5a529#diff-49ac9e6f6e6a855312bfcd393201f18ca53e6148c4a22a3a4949f1f9d1d137a8

### DIFF
--- a/src/ansi/iterator.rs
+++ b/src/ansi/iterator.rs
@@ -1,4 +1,7 @@
 use core::str::Bytes;
+use std::convert::TryFrom;
+use std::iter;
+use vte::{Params, ParamsIter};
 
 pub struct AnsiElementIterator<'a> {
     // The input bytes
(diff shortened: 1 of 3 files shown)


src/ansi/iterator.rs
```diff
@@ -116,19 +119,19 @@ impl<'a> Iterator for AnsiElementIterator<'a> {
...
-    fn csi_dispatch(&mut self, params: &[i64], intermediates: &[u8], ignore: bool, c: char) {
+    fn csi_dispatch(&mut self, params: &Params, intermediates: &[u8], ignore: bool, c: char) {
         if ignore || intermediates.len() > 1 {
             return;
         }
 
         if let ('m', None) = (c, intermediates.get(0)) {
             if params.is_empty() {
-                // Attr::Reset;
+                // Attr::Reset; TODO: does this need to be handled?
```
chrisduerr, line 131: Technically it can never be empty, but that's not a guarantee I would give. But with the current VTE version this is unreachable. Since even `\e[m` will give you the parameter `[0]`.

src/ansi/iterator.rs
```diff
@@ -162,139 +165,111 @@ impl vte::Perform for Performer {
...
-        if i >= parameters.len() {
-            break;
-        }
+    while let Some(param) = params.next() {
+        match param {
+            // [0] => Some(Attr::Reset),
+            [1] => style.is_bold = true,
+            [2] => style.is_dimmed = true,
+            [3] => style.is_italic = true,
+            [4] => style.is_underline = true,
```
chrisduerr, line 177: Technically this should probably be as follows:

```suggestion
            [4, ..] => style.is_underline = true,
```

Since it's possible to have underlines with more than one parameter separated by colons.

chrisduerr (commented):
The next VTE release will also remove the need to implement all trait methods, so that will probably help clean things up a bit. Something to look forward to maybe.

dandavison:
Thanks @chrisduerr!

PR #436, opened 2020-12-07
Make file style to be colored with same structure while color_only #405

ulwlu:
This pr changes below.

- Make file style to be colored with same structure while color_only

ss 4

ss 6

I confirmed below apart from test.

- `git -c 'interactive.diffFilter="./target/release/delta" --color-only' add -p` still works.

Remarks.

- Related to #405 
- I still need to resolve `commit-style` and `hunk-header-style`.
- Already ready for pr of `commit-style` and `hunk-header-style`, but they depend on this pr.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -188,7 +188,8 @@ where
             continue;
         }
 
-        if state == State::FileMeta && should_handle(&State::FileMeta, config) && !config.color_only {
+        if state == State::FileMeta && should_handle(&State::FileMeta, config) && !config.color_only
+        {
             // The file metadata section is 4 lines. Skip them under non-plain file-styles.
             continue;
         } else {
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -115,6 +115,10 @@ where
                     &minus_file,
                 ));
             }
+            if config.color_only {
+                handle_generic_file_meta_header_line(&mut painter, &line, &raw_line, config)?;
+                continue;
+            }
         } else if (state == State::FileMeta || source == Source::DiffUnified)
             && (line.starts_with("+++ ")
                 || line.starts_with("rename to ")
@@ -127,6 +131,10 @@ where
                 &plus_file,
             ));
             current_file_pair = Some((minus_file.clone(), plus_file.clone()));
+            if config.color_only {
+                handle_generic_file_meta_header_line(&mut painter, &line, &raw_line, config)?;
+                continue;
+            }
             if should_handle(&State::FileMeta, config)
                 && handled_file_meta_header_line_file_pair != current_file_pair
             {
@@ -180,7 +188,7 @@ where
             continue;
         }
 
-        if state == State::FileMeta && should_handle(&State::FileMeta, config) {
+        if state == State::FileMeta && should_handle(&State::FileMeta, config) && !config.color_only {
             // The file metadata section is 4 lines. Skip them under non-plain file-styles.
             continue;
         } else {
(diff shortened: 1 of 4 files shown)


src/delta.rs
```diff
@@ -180,7 +188,7 @@ where
             continue;
         }
 
-        if state == State::FileMeta && should_handle(&State::FileMeta, config) {
+        if state == State::FileMeta && should_handle(&State::FileMeta, config) && !config.color_only {
```
ulwlu: I think this can be refactored after `commit style` and `hunk header style` are also resolved.

src/delta.rs
```diff
@@ -317,7 +325,7 @@ fn handle_generic_file_meta_header_line(
     raw_line: &str,
     config: &Config,
 ) -> std::io::Result<()> {
-    if config.file_style.is_omitted {
+    if config.file_style.is_omitted && !config.color_only {
         return Ok(());
     }
```
ulwlu, line 331: This requires anyway, otherwise we need to refactor `should_handle`.
We can discuss and make another pr if we need.

src/options/set.rs
```diff
@@ -195,7 +195,6 @@ pub fn set_options(
         opt.side_by_side = false;
         opt.file_decoration_style = "none".to_string();
```
ulwlu, line 196: This is required.
file_decoration_style will break structures anyway. We should prevent any of them.

dandavison:
Great! Thanks for working on this @ulwlu.

> Already ready for pr of commit-style and hunk-header-style, but they depend on this pr.

You could open a second PR and use the PR target branch selector so that the second PR targets this branch. Unless I imagined it, I think GitHub is clever enough nowadays that the second one automatically switches to target master when the first merges.

src/delta.rs
```diff
@@ -127,6 +131,10 @@ where
                 &plus_file,
             ));
             current_file_pair = Some((minus_file.clone(), plus_file.clone()));
+            if config.color_only {
+                handle_generic_file_meta_header_line(&mut painter, &line, &raw_line, config)?;
+                continue;
+            }
```
dandavison, line 137: Would this be clearer as an `else if`?

dandavison (changes requested):
This is great, there's just one thing I'd ask for -- do you think you could add a failing test that demonstrates the problem that you are fixing? If you don't mind rebasing, it's nice to have the failing test be the first commit and the fixes come after (then I probably won't use squash-and-merge because the history is useful.)

The utility `ansi_test_utils::assert_line_has_style` might be useful for the test.

src/delta.rs
dandavison: OK. Feel free to add explanatory comments to the code while it is fresh in your brain. If we have some test coverage of this then we'll feel even better about refactoring :)

ulwlu: Oh yes! I'm gonna add.

ulwlu:
@dandavison 
Sure!

> You could open a second PR and use the PR target branch selector so that the second PR targets this branch. Unless I imagined it

Right, but I was not sure if this approach will be changed, and need to fix 2nd branch entirely or fix conflict (and I don't want so).

> do you think you could add a failing test that demonstrates the problem that you are fixing?

Sure, just a minute.

src/delta.rs
ulwlu: @dandavison 
Add comment here def9554

ulwlu:
@dandavison 

Add test here.
b48fa01

confirmed it's succeed now, and failed before the first commit.

dandavison:
Perfect, thanks! Merged.

PR #437, opened 2020-12-08
Make hunk header style to be colored with same structure while color_only #405

ulwlu:
## This pr changes below.

- Make hunk-header style to be colored with same structure while color_only

ss 6

## Remarks.

- Related to #405 
- I still need to resolve `commit-style`
- I will refactor too many conditions of color_only

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -434,7 +434,7 @@ fn handle_hunk_header_line(
     };
     let (raw_code_fragment, line_numbers) = parse::parse_hunk_header(&line);
     // Emit the hunk header, with any requested decoration
-    if config.hunk_header_style.is_raw {
+    if config.hunk_header_style.is_raw || config.color_only {
         if config.hunk_header_style.decoration_style != DecorationStyle::NoDecoration {
             writeln!(painter.writer)?;
         }
@@ -491,7 +491,10 @@ fn handle_hunk_header_line(
         painter
             .line_numbers_data
             .initialize_hunk(line_numbers, plus_file.to_string());
-    } else if config.line_numbers_show_first_line_number && !config.hunk_header_style.is_raw {
+    } else if config.line_numbers_show_first_line_number
+        && !config.hunk_header_style.is_raw
+        && !config.color_only
+    {
         let plus_line_number = line_numbers[line_numbers.len() - 1].0;
         let formatted_plus_line_number = if config.hyperlinks {
             features::hyperlinks::format_osc8_file_hyperlink(
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -327,7 +327,6 @@ fn handle_generic_file_meta_header_line(
     raw_line: &str,
     config: &Config,
 ) -> std::io::Result<()> {
-    // FIXME: this may be able to be refactored by moving to should_handle.
     if config.file_style.is_omitted && !config.color_only {
         return Ok(());
     }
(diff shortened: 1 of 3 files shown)


ulwlu:
I will rebase this `commit-style` PR https://github.com/ulwlu/delta/pull/4 to master, after this pr gets merged.

src/delta.rs
```diff
@@ -491,7 +491,10 @@ fn handle_hunk_header_line(
         painter
             .line_numbers_data
             .initialize_hunk(line_numbers, plus_file.to_string());
-    } else if config.line_numbers_show_first_line_number && !config.hunk_header_style.is_raw {
+    } else if config.line_numbers_show_first_line_number
+        && !config.hunk_header_style.is_raw
+        && !config.color_only
+    {
```
dandavison, line 497: If you're able to add a comment explaining this boolean logic (here and in other places) that would be really helpful. I'm finding that when I haven't worked on the code for a while that these boolean conditions are very hard to understand!

dandavison:
Do you think it would be nice to have `hunk-header-style = syntax` working in color-only mode? Maybe that can be for a future PR.

src/delta.rs
ulwlu: I see. I'm going to add.

ulwlu:
@dandavison 

> Do you think it would be nice to have hunk-header-style = syntax working in color-only mode?

like this?

ss

Yes it makes sense, this is more understandable. As you said if we'll do this, it will be future pr because we need to change test a bit.

src/delta.rs
ulwlu: @dandavison 
added comment here.
ed9f3c9

src/delta.rs
```diff
@@ -327,7 +327,6 @@ fn handle_generic_file_meta_header_line(
     raw_line: &str,
     config: &Config,
 ) -> std::io::Result<()> {
-    // FIXME: this may be able to be refactored by moving to should_handle.
```
ulwlu, line 330: Sorry, I added fixme, but I realized this can't be refactored.

When file style is assinged to `omit` we should return OK, however, if it's while color_only mode, we shouldn't return OK.

If we move this condition to should_handle it would be messy because not only State::FileMeta is using should_handle.

src/delta.rs
```diff
@@ -491,7 +490,12 @@ fn handle_hunk_header_line(
         painter
             .line_numbers_data
             .initialize_hunk(line_numbers, plus_file.to_string());
-    } else if config.line_numbers_show_first_line_number && !config.hunk_header_style.is_raw {
+    } else if config.line_numbers_show_first_line_number
+        && !config.hunk_header_style.is_raw
+        && !config.color_only
+    {
+        // With raw mode or color-only mode,
+        // we should prevent the output from creating new line for printing line number
```
dandavison, line 498: 👍

dandavison:
Awesome, thanks again!

> Yes it makes sense, this is more understandable. As you said if we'll do this, it will be future pr because we need to change test a bit.

Great. (I thought the current test using "red" would not need changing?)

ulwlu:
@dandavison 
Yes, but I think it would be better to add this to `_do_test_output_is_in_one_to_one_correspondence_with_input` for confirming not changing line-number.

Currently it won't be a problem, however, I worry about the future if `syntax` behavior changes.

PR #443, opened 2020-12-10
Enable hunk header style with syntax when color only

ulwlu:
## this pr changes below.

- Enable `hunk-header-style syntax` with `color-only`

ss 4

## remarks
Related to https://github.com/dandavison/delta/pull/437#issuecomment-740691252.

I have said I need to change test to achieve this, but I found way to keep structure with syntax. Therefore I don't add any test.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -115,6 +115,11 @@ where
                     &minus_file,
                 ));
             }
+
+            // In color_only mode, raw_line's structure shouldn't be changed.
+            // So it needs to avoid fn handle_file_meta_header_line
+            // (it connects the plus_file and minus_file),
+            // and to call fn handle_generic_file_meta_header_line directly.
             if config.color_only {
                 handle_generic_file_meta_header_line(&mut painter, &line, &raw_line, config)?;
                 continue;
@@ -131,6 +136,11 @@ where
                 &plus_file,
             ));
             current_file_pair = Some((minus_file.clone(), plus_file.clone()));
+
+            // In color_only mode, raw_line's structure shouldn't be changed.
+            // So it needs to avoid fn handle_file_meta_header_line
+            // (it connects the plus_file and minus_file),
+            // and to call fn handle_generic_file_meta_header_line directly.
             if config.color_only {
                 handle_generic_file_meta_header_line(&mut painter, &line, &raw_line, config)?;
                 continue;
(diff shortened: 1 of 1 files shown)


src/delta.rs
```diff
@@ -448,11 +448,19 @@ fn handle_hunk_header_line(
...
-            s if !s.is_empty() => format!("{} ", s),
-            s => s,
+        // Keep the raw_line structure when color_only mode
+        let line = if config.color_only {
+            format!(" {}", &line)
+        } else {
+            match painter.prepare(&raw_code_fragment, false) {
+                s if !s.is_empty() => format!("{} ", s),
+                s => s,
+            }
```
dandavison, line 458: Question about code before your change: do you know why `painter.prepare()` is being called here? (Is it to expand tabs?).

dandavison: For documentation purposes, can you explain the reason that we don't call `painter.prepare()` under `color-only`?

dandavison:
@ulwlu, awesome, thanks for the continued work!

Could I ask you a favour? Would you mind adding comments to all the `if` statements involving `color-only`, explaining the reason for the boolean logic in some detail?? It's fine if the comments are a bit long -- I can edit them if I want to. And of course I can edit grammar so don't worry about that!

The reason I ask is that this area has been quite tricky and hard to understand, and I've found that a few weeks/months later I no longer understand the boolean logic.

ulwlu:
Sure, I'm gonna add the comments.

src/delta.rs
ulwlu: @dandavison 
The reason is this (it may be long, sorry).

#### ① `can you explain the reason that we don't call painter.prepare() under color-only?`

Simply it's because it doesn't need to use `painter.prepare()`. Because we won't change structure from raw_line, and will not use decoration.

#### ② then `do you know why painter.prepare() is being called here? (Is it to expand tabs?)`

I think it exists for expanding tabs `just in the case there are tabs` (just my speculation, I couldn't find exact purpose in `git blame`). However, this doesn't need it because hunk header will never have tabs even when it's nested.

So, it actually could be written like below.
```rust
match raw_code_fragment {
    s if !s.is_empty() => format!("{} ", s),
    s => s,
}

// This change passes all tests in fact.
```

This match condition is necessary, because some options will be broken such as box.
ss

ulwlu:
@dandavison 
Thank you for reviewing.

I added comments here.
a76e450

src/delta.rs
dandavison: > So, it actually could be written like below.
 > // This change passes all tests in fact.

Great. Let's bear that in mind for a little future cleanup. Then we can get rid of that second argument to `prepare`.

dandavison (approved):
These comments are fantastic! They're going to be really useful, I'm sure of it.

LGTM!

PR #445, opened 2020-12-13
Add cd to update brew formula in homebrew core

ulwlu:
This pr changes below.

- add https://github.com/marketplace/actions/homebrew-bump-formula
- it will send pr to homebrew-core automatically on pushing tags (it means release).

Example
https://github.com/jesseduffield/lazygit/blob/master/.github/workflows/cd.yml
https://github.com/ablinov/declutter/blob/master/.github/workflows/bump_homebrew_formula.yml
https://github.com/asciidoc/asciidoc-py3/blob/master/.github/workflows/release.yml
https://github.com/bow-swift/nef/blob/master/.github/workflows/bump-formula.yml

This way is same as delta used to. This uses `brew bump-formula-pr`, and delta used too.
https://github.com/Homebrew/homebrew-core/pull/62680

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -65,3 +65,14 @@ jobs:
             git-delta*.deb
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  bump:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dandavison:
Great, thanks! This should make release much easier and quicker. I've added a commit deleting the Makefile targets I was using for this.

ulwlu:
Oh I missed that makefile. Thank you for adding commit and review!

PR #453, opened 2020-12-19
Add file path to hunk header

### LINKED ISSUE #309: Is it possible to show the file path in the hunk header?
Hi,

Right now, the hunk header only shows the line number. diff-so-fancy also shows the file path, which I like because it gives me better orientation in a large diff (although it can be redundant with the commit/file box).
Is it possible to emulate this in delta?

Thanks!

dandavison:
Fixes #309  (cc @ulwlu just in case you feel like reviewing).

This PR allows a special style attribute `file` to be used in `hunk-header-style`. It causes the file path to be included in the hunk header. For example: `hunk-header-style = file syntax`. The style for the file path is determined by `file-style`.

- 3c31223dbfcbcb32c3c2da990091cc65e9f6eca1 adds a failing test for the feature.
- 334dec742d78d6d72715eaefaaa1c8313c6315fb implements the feature and makes the test pass

It is possible to omit the main file section and display hunk headers only. Here's an example:

```gitconfig
[delta]
	hunk-header-style = file syntax  # use new attribute
	file-style = red omit  # set the color for the file path, but omit the main file section
	hunk-header-decoration-style = blue box ul
```

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::fmt::Write as FmtWrite;
 use std::io::BufRead;
 use std::io::Write;
 
@@ -485,6 +486,13 @@ fn handle_hunk_header_line(
             writeln!(painter.writer)?;
         }
         if !line.is_empty() {
+            if config.hunk_header_style_include_file_path {
+                let _ = write!(
+                    &mut painter.output_buffer,
+                    "{}: ",
+                    config.file_style.paint(plus_file)
+                );
+            };
             let lines = vec![(line, State::HunkHeader)];
             let syntax_style_sections = Painter::get_syntax_style_sections_for_lines(
                 &lines,
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -509,7 +509,7 @@ fn handle_hunk_header_line(
                 &painter.output_buffer,
                 &painter.output_buffer,
                 &config.decorations_width,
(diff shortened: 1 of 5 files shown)


src/delta.rs
```diff
@@ -509,7 +517,7 @@ fn handle_hunk_header_line(
                 &painter.output_buffer,
                 &painter.output_buffer,
                 &config.decorations_width,
-                config.hunk_header_style,
+                config.null_style,
```
dandavison, line 520: This change is orthogonal to this PR. Basically we were painting the hunk header text twice, since the text being passed to `draw_fn` has already been painted. So I believe it's a bug fix, but not a bug likely to have been noticed by anyone. It caused the unit test here to fail, because I was expecting the line to start with `file-style`, but in fact it started with `hunk-header-style`. I.e. the line looked like

```
(hunk-header-style) (file-style) file.rs (hunk-header-style) fn my_fun () {
```
and with this fix it looks like
```
(file-style) file.rs (hunk-header-style)  fn my_fun () {
```
The commit message:
```
commit 31066fc8803e47e5f992c53ab611e8077093693a
Author: Dan Davison <dandavison7@gmail.com>
Date:   Sat Dec 19 12:32:30 2020 +0000

    Do not paint hunk header twice
    
    The text being passed to the draw function has already been painted.

 src/delta.rs | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)
```

PR #464, opened 2020-12-24
fix issue 463

### LINKED ISSUE #463: 🐛 delta 0.4.5 binary release messes up terminal on Ubuntu 20 04
Release 0.4.5, when run on Ubuntu 20 04, messes up the terminal.  After running `delta`, I see the diff output; however, any subsequent typing within the terminal is not echo'd back.

```
salman@byteevo-scm:~/Scratch$ ls update-pcs*
update-pcs  update-pcs-new
salman@byteevo-scm:~/Scratch$ ./delta -s update-pcs update-pcs-new 

comparing: update-pcs ⟶   update-pcs-new
───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

│ 1  │#!/bin/sh                                                │ 1  │#!/bin/sh
│ 2  │for i in $(seq 3); do                                    │ 2  │for i in $(seq 4); do
│8m 3  │    rsync -av -e "ssh -i $HOME/.ssh/id_rsa" $HOME/.local→│ 3  │    rsync -av -e "ssh -i $HOME/.ssh/id_rsa" $HOME/.local→
│ 4  │    rsync -av -e "ssh -i $HOME/.ssh/id_rsa" $HOME/.local→│ 4  │    rsync -av -e "ssh -i $HOME/.ssh/id_rsa" $HOME/.local→
│ 5  │done                                                     │ 5  │done
salman@byteevo-scm:~/Scratch$ 
Command 'lll' not found, did you mean:

  command 'dll' from deb brickos (0.9.0.dfsg-12.2)
  command 'lld' from deb lld (1:10.0-50~exp1)
  command 'lli' from deb llvm-runtime (1:10.0-50~exp1)
  command 'llt' from deb storebackup (3.2.1-1+deb8u1build0.20.04.1)
  command 'llc' from deb llvm (1:10.0-50~exp1)

Try: sudo apt install <deb name>

salman@byteevo-scm:~/Scratch$ delta 0.4.5
salman@byteevo-scm:~/Scratch$ 
```

As you can see, I intentionally typed a non-existant command `lll`, which is not echo'd back.  Then I typed `./delta --version`, which is also not echo'd back.  The result of the command is, however, seen in the output above.

MarcoIeni:
Fix #463
Ugly, but it works. I am sure there is a better solution, so if you can find it, great!

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -133,21 +133,14 @@ fn diff(
             process::exit(config.error_exit_code);
         });
 
+    let diff_output = BufReader::new(diff_process.stdout.unwrap()).byte_lines();
     let delta_result = match writer {
         None => {
             let mut output_type = OutputType::from_mode(config.paging_mode, None, &config).unwrap();
             let writer = output_type.handle().unwrap();
-            delta(
-                BufReader::new(diff_process.stdout.unwrap()).byte_lines(),
-                writer,
-                &config,
-            )
+            delta(diff_output, writer, &config)
         }
-        Some(writer) => delta(
-            BufReader::new(diff_process.stdout.unwrap()).byte_lines(),
-            writer,
-            &config,
-        ),
+        Some(writer) => delta(diff_output, writer, &config),
     };
 
     if let Err(error) = delta_result {
--- a/src/main.rs
+++ b/src/main.rs
@@ -76,17 +76,17 @@ fn main() -> std::io::Result<()> {
         process::exit(0);
     }
 
-    let mut output_type = OutputType::from_mode(config.paging_mode, None, &config).unwrap();
-    let mut writer = output_type.handle().unwrap();
-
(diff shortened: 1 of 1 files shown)


dandavison:
@MarcoIeni can you explain why this fixes the problem?

MarcoIeni:
Not really! :sweat_smile: 
I just noticed that if `output_type.handle().unwrap()`, i.e. `writer` is passed as function argument it doesn't work. I think it has to do with lifetimes, but the rust compiler doesn't complain.
I think that a better fix would be to edit the `handle` function.

dandavison:
I think I (eventually!) found a cleaner solution: https://github.com/dandavison/delta/pull/469

dandavison:
Thanks very much @MarcoIeni  for locating the bug and pointing out that the fix was to stop that `writer` getting destroyed too early.

PR #467, opened 2020-12-26
ci: run also on push

MarcoIeni:


dandavison:
Hi @MarcoIeni, with `[push, pull_request]` it was running everything twice when we pushed commits to an open pull request (see https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662). So I removed `push` in https://github.com/dandavison/delta/pull/450. What I was thinking was that with `pull_request` alone, it will run once when we push to an open PR, and I will be careful not to push commits that need CI directly to master.

MarcoIeni:
Ok!

If you put something like this it will run ci only when you push to master:

```yaml
on:
  push:
    branches:
      - master
  pull_request:
```

If you don't like this and you prefer to leave only `pull_request` just close this PR :)

dandavison:
Thanks @MarcoIeni, I changed it to the example you gave.

Issue #462, opened 2020-12-24
🐛  delta 0.4.5 binary release does not run on macOS 10.11 (El Capitan)


phil-blain:
To replicate: 
1. download https://github.com/dandavison/delta/releases/tag/0.4.5 asset delta-0.4.5-x86_64-apple-darwin.tar.gz
2. untar and `cd` to the extracted folder
3. run `./delta --version`:
~~~bash
$ ./delta --version
dyld: lazy symbol binding failed: Symbol not found: ____chkstk_darwin
  Referenced from: /Users/Philippe/Downloads/delta-0.4.5-x86_64-apple-darwin 2/./delta
  Expected in: /usr/lib/libSystem.B.dylib

dyld: Symbol not found: ____chkstk_darwin
  Referenced from: /Users/Philippe/Downloads/delta-0.4.5-x86_64-apple-darwin 2/./delta
  Expected in: /usr/lib/libSystem.B.dylib

Trace/BPT trap: 5
~~~

dandavison:
Thanks @phil-blain. That binary is working for me on MacOS 10.15.7 (Catalina). Quick googling brings up https://github.com/esy/esy/issues/1169#issuecomment-691799356  and https://github.com/nodegui/nodegui/issues/391#issuecomment-582328912 which suggest that the problem is that our GitHub Actions CD is building the MacOS binary on a recent version of MacOS:

```yaml
        job:
          - { os: macos-latest,   target: x86_64-apple-darwin,         use-cross: false }
```
 
So one possibility would be to re-release with that `os:` set to an older version. However https://github.com/dandavison/delta/issues/463 reports that a binary is not working on Ubuntu 20.04, which I am still investigating. cc @MarcoIeni

MarcoIeni:
> So one possibility would be to re-release with that `os:` set to an older version.

Sadly GitHub actions only support:
- macOS Catalina 10.15 (`macos-latest` or `macos-10.15`) (this is what we are using)
- macOS Big Sur 11.0 (`macos-11.0`)

See https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources

I believe there is no older version.

dandavison:
Hm, that's a shame. What should we do? MacOS users do have Homebrew and MacPorts as installation options. We could revert to Travis if we had to.

MarcoIeni:
[...]

PR #468, opened 2020-12-26
cd: set MACOSX_DEPLOYMENT_TARGET to 10.7

phil-blain:
The binary release of delta 0.4.5 for macOS created by the "Continuous
Deployment" GitHub Actions workflow does not run on older macOS, like
10.11 (El Capitan) [1].

The Rust compiler, rustc, by default, will build macOS binaries
compatible with macOS 10.7 or newer [2], [3], but some *-sys crates invoke a
C compiler and thus will target the macOS version the C compiler is
running on [4], thus limiting the compatibility of the resulting
binary. The macOS virtual machines used on GitHub Actions run macOS
10.15.7, and since delta depends on some *-sys crates, the delta binary
built by cargo targets macOS 10.15, as can be verified by using the
following command:

    $ otool -l ./target/x86_64-apple-darwin/release/delta | grep LC_BUILD_VERSION -A4
    cmd LC_BUILD_VERSION
    cmdsize 32
    platform 1
    sdk 10.15.6
    minos 10.15

or, if building locally on an older macOS versions, (here 10.11):

    $ otool -l ./target/debug/delta | grep  LC_VERSION_MIN_MACOSX -A2
    cmd LC_VERSION_MIN_MACOSX
    cmdsize 16
    version 10.11

To restore compatibility with older macOS version, explicitely set the environment
variable `MACOSX_DEPLOYMENT_TARGET` to the default for rustc, "10.7".
This will make the `clang` C compiler invoked by *-sys crates also
target 10.7, thus restoring the compatibility of the delta binary.

For simplicity, add this variable to the environment for all platforms,
which should not have any effect on non-macOS platforms.

[1] https://github.com/dandavison/delta/issues/462
[2] https://github.com/rust-lang/rust/blob/65d053ab74d8c8c9c502b678acc265f3d7e2ac49/compiler/rustc_target/src/spec/apple_base.rs#L15-L17
[3] https://github.com/rust-lang/rust/blob/65d053ab74d8c8c9c502b678acc265f3d7e2ac49/compiler/rustc_target/src/spec/apple_base.rs#L53-L61
[4] https://users.rust-lang.org/t/compile-rust-binary-for-older-versions-of-mac-osx/38695/5

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -30,6 +30,8 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
       - name: Cargo build
+        env:
+          MACOSX_DEPLOYMENT_TARGET: 10.7
         uses: actions-rs/cargo@v1
         with:
           command: build


MarcoIeni:
Thank you for the detailed explanation!
I don't have ways to test this, but it looks good.

Just one question, do you know if we can support older MACOSX versions, like 10.5 or 10.6 for example? What's the minimum?

phil-blain:
From what I was able to understand while researching this fix, as of now Rust itself is compatible with 10.7 and later only, or maybe 10.6 if building Rust from source with some special `configure` (or similar) options (I'm not sure): 

- https://github.com/rust-lang/rust/issues/11927
- https://github.com/rust-lang/rust/issues/25342

So I think that 10.7 should be enough, being the minimum version supported by Rust.

MarcoIeni:
Sure, thank you! It was also written in link [2].

For me this can be merged.

phil-blain:
Also, I did not modify the README to revert 2bbf189, but it might be good to do so before the next release... I can submit a subsequent PR once this one is merged.

dandavison:
Thanks very much @phil-blain -- for researching the proper solution, implementing, and providing detailed explanations.

The two Rust command line projects I tend to check against are bat and ripgrep; I should have thought to look at those when we were first discussing this. Looks like ripgrep may be coming to the same solution as here in https://github.com/BurntSushi/ripgrep/issues/1737 (there is some discussion there that may be relevant). However, on a quick search I didn't see any issues regarding this in bat, nor is it obviously addressed in https://github.com/sharkdp/bat/blob/master/.github/workflows/CICD.yml.

I'll aim to get a fixup release out soon (that will be 0.5.0).

PR #472, opened 2020-12-28
handle hunk header refactor

dandavison:
Pure refactoring; no functional changes. cc @ulwlu

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -442,7 +442,7 @@ fn handle_hunk_line(
             };
             painter
                 .minus_lines
-                .push((painter.prepare(&line, true), state.clone()));
+                .push((painter.prepare(&line), state.clone()));
             state
         }
         Some('+') => {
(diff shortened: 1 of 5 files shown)


ulwlu:
It really looks more understandable. Thank you very much. (I noticed clippy ci causes 3 warnings.)

IMO: Maybe other draw_** such as `handle_commit_meta_header_line` will be separated someday, so would it be cleaner if we move `src/hunk_header.rs` to `src/draw/hunk_header.rs` ?

dandavison:
Thanks @ulwlu. I've fixed the clippy errors and put the `handle_*` function back in delta.rs with the other ones in https://github.com/dandavison/delta/pull/474. I agree that there's more straightening out that could be done but this is a start (I was on a flight today...)

ulwlu:
#474 is great! What a fast response. I hope you have a nice trip.

PR #480, opened 2020-12-29
refactor main in order to call process::exit once

MarcoIeni:
Just a little refactoring.
I tried to follow the [example](https://doc.rust-lang.org/std/process/fn.exit.html#examples) on the rust docs about `process::exit`.
This is the cleanest way to avoid other issues like #463 in my opinion.
Manually calling `mem::drop` is not that elegant. With this refactoring is not needed anymore. See line 91.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -28,7 +28,6 @@ mod syntect_color;
 mod tests;
 
 use std::io::{self, ErrorKind, Read, Write};
-use std::mem;
 use std::path::PathBuf;
 use std::process;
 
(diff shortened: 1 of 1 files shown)


MarcoIeni:
Anyway the work is not finished, there are other `process::exit` in the code, so if you like this idea, we could open an issue that says to remove them.

src/main.rs
```diff
@@ -88,17 +88,23 @@ fn main() -> std::io::Result<()> {
...
         }
     };
-    Ok(())
+    Ok(0)
+}
+
+#[cfg(not(tarpaulin_include))]
+fn main() -> std::io::Result<()> {
+    let exit_code = run_app()?;
+    // when you call prcess::exit, no destructors are called, so we want to do it only once, here
```
dandavison, line 106:
```suggestion
    // when you call process::exit, no destructors are called, so we want to do it only once, here
```

dandavison (approved):
Excellent, thanks for doing this the proper way. LGTM!

PR #495, opened 2021-01-08
Honor GIT_CONFIG_PARAMETERS env var

### LINKED ISSUE #307: 🐛 Can't disable line numbering from command line

### LINKED ISSUE #493: 🐛 Delta does not respect git -c
If I have the following in git's global `.gitconfig`:

```ini
[delta]
side-by-side = true
```
When I issue the following git command:

```sh
$ git -c delta.side-by-side=false diff ./somefile
```
I still get a side by side view. `git -c` should always override the config file.

dandavison:
cc @wyuenho @ulwlu 

When git is invoked as `git -c aaa.bbb=ccc -c ddd.eee=fff` then git
sets the env var `GIT_CONFIG_PARAMETERS` containing the changed config
entries, so that child processes can honor them.

libgit2 doesn't yet honor the env var: see
https://github.com/libgit2/libgit2/issues/3854.

Fixes #493
Fixes #307
Ref https://github.com/dandavison/magit-delta/issues/13


dandavison:
This allows us to do things like

```
git -c delta.line-numbers=false -c delta.max-line-distance=0.8 show

# note: while quotes around color hex codes are required in gitconfig, they can be omitted in the `git -c` context
git -c delta.plus-style='syntax "#aa0000"' show
git -c delta.plus-style='syntax #aa0000' show
```

Thanks @wyuenho for pointing out that this should be made possible; I think this now becomes the preferred method for changing delta config on-the-fly. I'll need to remember to update the README at the next release.

wyuenho:
Nice! Can't wait to test it out in the next release.

Issue #115, opened 2020-03-05
Diff highlighting overflows onto the next line (2 chars too long)


gibfahn:
It looks like delta's syntax highlighting is now overrunning the line by two characters.

[![asciicast](https://asciinema.org/a/307651.svg)](https://asciinema.org/a/307651)

gibfahn:
`git show | delta --width=$(($COLUMNS-2))` works, so I'm guessing that what happens is that the diff is indented by 2 chars, and those 2 chars aren't being removed from the line length.

`git bisect` shows the commit that introduced the issue is 185dbd5e87e35a625ceaaea7fd36a35e948e44b1 (unsurprisingly, as that's the one that changed how the text is printed).

dandavison:
Hey @gibfahn, thanks! I'm not seeing this but there's clearly a bug. What are your values of `PAGER` and `BAT_PAGER`? Problems like this in the past have been related to `less` options:

https://github.com/dandavison/delta/issues/61
https://github.com/dandavison/delta/issues/41
https://github.com/dandavison/delta/issues/10

dandavison:
This makes me think, perhaps we should add a `delta --environment` option which prints out all the relevant environment variables and anything else delta is responding to, to help diagnose issues.

gibfahn:
>What are your values of PAGER and BAT_PAGER and LESS? Problems like this in the past have been related to less options:

Yeah I have some stuff :)

```shell
PAGER=less
LESS=--tabs=4 --ignore-case --RAW-CONTROL-CHARS --quit-if-one-screen --status-column --LONG-PROMPT --HILITE-UNREAD --window=-4
BAT_PAGER=less --tabs=4 --ignore-case --RAW-CONTROL-CHARS --quit-if-one-screen --status-column --LONG-PROMPT --HILITE-UNREAD --window=-4
```

Using `git show | LESS= BAT_PAGER= delta` does "fix" things.

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/15943089/76007967-cbb06900-5f06-11ea-9781-78572f642602.png">


<img width="1680" alt="image" src="https://user-images.githubusercontent.com/15943089/76007939-c2270100-5f06-11ea-846e-746f889f43c2.png">



>This makes me think, perhaps we should add a delta --environment option which prints out all the relevant environment variables and anything else delta is responding to, to help diagnose issues.

[...]

PR #512, opened 2021-02-01
Option to set the background extension mode to ANSI or spaces

th1000s:
(Original topic: Fix right panel background extending beyond the truncation symbol)

In side-by-side mode, if `background_color_extends_to_terminal_width`
is set, the left panel color is extended via spaces, but the right
one via an ANSI sequence which instructs the terminal emulator to
fill the background color rightwards.

However due to the padding after the right panel the color will
extend beyond a possible truncation symbol:

`TEXT#AB→#`, not `TEXT#AB→ ` with `#` being the green background.

This is fixed by using the same logic as in the left panel for the
right panel.

---------------------

Before-After comparison in Konsole:

sbs2

### DIFF
--- a/src/features/line_numbers.rs
+++ b/src/features/line_numbers.rs
@@ -383,7 +383,7 @@ pub mod tests {
         let mut lines = output.lines().skip(7);
         let (line_1, line_2) = (lines.next().unwrap(), lines.next().unwrap());
         assert_eq!(strip_ansi_codes(line_1), " 1  ⋮    │a = 1");
-        assert_eq!(strip_ansi_codes(line_2), " 2  ⋮    │b = 2");
+        assert_eq!(strip_ansi_codes(line_2), " 2  ⋮    │b = 23456");
     }
 
     #[test]
@@ -407,7 +407,7 @@ pub mod tests {
         let mut lines = output.lines().skip(7);
         let (line_1, line_2) = (lines.next().unwrap(), lines.next().unwrap());
         assert_eq!(strip_ansi_codes(line_1), "    ⋮ 1  │a = 1");
-        assert_eq!(strip_ansi_codes(line_2), "    ⋮ 2  │b = 2");
+        assert_eq!(strip_ansi_codes(line_2), "    ⋮ 2  │b = 234567");
     }
 
     #[test]
@@ -489,7 +489,7 @@ pub mod tests {
         let mut lines = output.lines().skip(5);
         let (line_1, line_2) = (lines.next().unwrap(), lines.next().unwrap());
         assert_eq!(strip_ansi_codes(line_1), " 1  ⋮    │-a = 1");
-        assert_eq!(strip_ansi_codes(line_2), " 2  ⋮    │-b = 2");
+        assert_eq!(strip_ansi_codes(line_2), " 2  ⋮    │-b = 23456");
     }
 
     #[test]
@@ -514,7 +514,7 @@ index 223ca50..e69de29 100644
 +++ w/a.py
 @@ -1,2 +0,0 @@
 -a = 1
--b = 2
+-b = 23456
 ";
 
     pub const TWO_PLUS_LINES_DIFF: &str = "\
@@ -525,7 +525,7 @@ index 0000000..223ca50
 +++ i/a.py
 @@ -0,0 +1,2 @@
 +a = 1
-+b = 2
++b = 234567
 ";
 
     pub const ONE_MINUS_ONE_PLUS_LINE_DIFF: &str = "\
--- a/src/features/line_numbers.rs
+++ b/src/features/line_numbers.rs
@@ -6,9 +6,10 @@ use regex::Regex;
 use crate::config;
 use crate::delta::State;
 use crate::features::hyperlinks;
-use crate::features::side_by_side;
+use crate::features::side_by_side::PanelSide;
 use crate::features::OptionValueFunction;
 use crate::format;
+use crate::plusminus::*;
 use crate::style::Style;
 
 pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
@@ -65,11 +66,11 @@ pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
 pub fn format_and_paint_line_numbers<'a>(
     line_numbers_data: &'a mut LineNumbersData,
     state: &State,
-    side_by_side_panel: Option<side_by_side::PanelSide>,
+    side_by_side_panel: Option<PanelSide>,
     config: &'a config::Config,
 ) -> Vec<ansi_term::ANSIGenericString<'a, str>> {
-    let m_ref = &mut line_numbers_data.hunk_minus_line_number;
-    let p_ref = &mut line_numbers_data.hunk_plus_line_number;
+    let nr_left = line_numbers_data.line_number[Minus];
+    let nr_right = line_numbers_data.line_number[Plus];
     let (minus_style, zero_style, plus_style) = (
         config.line_numbers_minus_style,
         config.line_numbers_zero_style,
@@ -77,20 +78,17 @@ pub fn format_and_paint_line_numbers<'a>(
     );
     let ((minus_number, plus_number), (minus_style, plus_style)) = match state {
         State::HunkMinus(_) => {
-            let m = *m_ref;
-            *m_ref += 1;
-            ((Some(m), None), (minus_style, plus_style))
+            line_numbers_data.line_number[Minus] += 1;
+            ((Some(nr_left), None), (minus_style, plus_style))
         }
         State::HunkZero => {
-            let (m, p) = (*m_ref, *p_ref);
-            *m_ref += 1;
-            *p_ref += 1;
-            ((Some(m), Some(p)), (zero_style, zero_style))
+            line_numbers_data.line_number[Minus] += 1;
+            line_numbers_data.line_number[Plus] += 1;
+            ((Some(nr_left), Some(nr_right)), (zero_style, zero_style))
         }
         State::HunkPlus(_) => {
-            let p = *p_ref;
-            *p_ref += 1;
-            ((None, Some(p)), (minus_style, plus_style))
+            line_numbers_data.line_number[Plus] += 1;
+            ((None, Some(nr_right)), (minus_style, plus_style))
         }
         _ => return Vec::new(),
     };
@@ -99,14 +97,14 @@ pub fn format_and_paint_line_numbers<'a>(
 
     let (emit_left, emit_right) = match (config.side_by_side, side_by_side_panel) {
         (false, _) => (true, true),
-        (true, Some(side_by_side::PanelSide::Left)) => (true, false),
-        (true, Some(side_by_side::PanelSide::Right)) => (false, true),
+        (true, Some(PanelSide::Left)) => (true, false),
+        (true, Some(PanelSide::Right)) => (false, true),
         (true, None) => unreachable!(),
     };
 
     if emit_left {
         formatted_numbers.extend(format_and_paint_line_number_field(
-            &line_numbers_data.left_format_data,
+            &line_numbers_data.format_data[PanelSide::Left],
             &config.line_numbers_left_style,
             minus_number,
             plus_number,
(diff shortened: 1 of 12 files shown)


th1000s:
The background is now only filled with spaces if any line in a block gets truncated (or, soon, wrapped):

ansi_spaces

This required moving some code written for wrapping here, I will now rebase the rest onto this.

dandavison:
@th1000s can you explain why we ever need to fill with spaces? Sorry if I'm being slow! I don't think I've understood the motivation for this PR. I see this problem:

But can't that be fixed by setting the background color of the continuation character to the fill color? I think it's correct that the background color extends beyond the continuation character; to my mind, the bug here is just that the background is white for one character, when it should be green.

th1000s:
No problem. Hereafter "ANSI" means the ANSI Sequence which extends the color until the end of the line.

So, ANSI placed after the truncation symbol is the problem [1] (adapted from your screenshot), ANSI placed on lines which are not cut off is fine [2].

But, just selectively skipping ANSI at e.g. [3] but still emitting it where the line is not cut of [5] would result in a [jagged right edge](https://user-images.githubusercontent.com/78380144/106798758-b97f0380-665e-11eb-9888-2729328b73b4.png), here crossed out in red.

Not using ANSI or spaces doesn't work either, because then the cut off lines are fine [3], but shorter lines are not [4]. 

This means the only option is to not use ANSI for truncated lines, and filling up lines which are too short with **spaces**, i.e. [3] and [5] up to the dotted line.

image

dandavison:
Thanks very much for explaining that carefully. Would another alternative be to place the continuation character on the green (or whatever) background color instead of on the default (here white) background color? I was thinking that would then allow the code to be almost unchanged, with ANSI sequences used on all lines to fill the background color to the terminal width.

dandavison:
Continuing my comment above, here are a couple of screenshots (on this branch at 306621bac7d3e5494dcd163fd8b9ca5f242132c5). Please correct me if I'm wrong but what I'm thinking is that even with this plan:

> not use ANSI for truncated lines, and filling up lines which are too short with spaces

we are still going to get "jagged right edges" in a sense, because the column to which the right fill extends will vary between  files and diff hunks. Hence my question about whether we should consider just placing everything, including the continuation character, on the colored background and always filling rightwards to the maximum extent, using the ANSI sequence in the right panel. 

th1000s:
Yes, making the background of a continuation character (`↵ / ↴ / →`) the same color as the rest of the line does make it visually more consistent. I will update it to that.

> jagged right edges

For larger visual component these are still there, that is true.

But back to the initial issue, for my personal taste the inconsistency of having the background _always_ end with exactly the last character on the red/left side, but having it extend further on the right side is worse -  hence my initial fix of _always_ padding with spaces.

And while this ruins terminal resizing/shrinking, in side-by-side mode that is going to happen much sooner anyhow. To help with that, maybe the pager could be started with `--chop-long-lines`. I also do not thing users change the size that often :). On the long run the only real fix is going to be using some terminal-ui which can redraw on resizes, maybe after some parts have been extracted into libdelta.

dandavison:
> for my personal taste the inconsistency of having the background always end with exactly the last character on the red/left side, but having it extend further on the right side is worse - hence my initial fix of always padding with spaces.

I'm still not sure about this (in the sense that I'm genuinely not sure, not a euphemism for being against!). One consideration is that we haven't had any complaints from other users, and also this was the explicit motivation for the ANSI code in the first place. Is it possible for this decision to be made independently of adding the `side-by-side-wrapped` feature? Switching to spaces seems to be quite a large code change whereas I think changing the background color of the continuation character would be a small change. Is it possible for the `side-by-side-wrapped` branch to not be based on top of the more radical (switching to spaces) change?

th1000s:
No problem, I'll adjust the background color of the continuation character to match that of the rest of the line. I'd also want to set the foreground color to blue to remove any possible ambiguities and still have it stand out a bit, similar how vim can highlight non-printable characters. Blue seems to be quite readable on both dark and light backgrounds (maybe tab-highlighting can work that way in delta as well).

nonprint

As for the for the ANSI code motivation, my impression was that these were more relevant for the unified diff view and were then re-used for the right panel of the side-by-side mode.

dandavison:
> the inconsistency of having the background always end with exactly the last character on the red/left side, but having it extend further on the right side is worse -

OK, I think it's important that I understand this exactly. What I had thought was that side-by-side splits the user's terminal window exactly in half (modulo odd number width) and then applies equal (visual) widths of color in each panel. However, I'm looking at this, and I think I understand what you're saying: there seem to be two "extra" characters width of green beyond the truncation symbol on the right. So, why is that? Is the terminal width measurement or splitting calculation wrong?

> As for the for the ANSI code motivation, my impression was that these were more relevant for the unified diff view and were then re-used for the right panel of the side-by-side mode.

Yes, I think that's true. And I do agree with you that we want equal width color on each side (filling the full terminal width ideally, and modulo odd-number terminal widths)

dandavison:
> No problem, I'll adjust the background color of the continuation character to match that of the rest of the line.

Great.

Regarding the spaces vs ANSI I am worried about causing you unnecessary work in the event that we both agree in the end that spaces is the right solution! So please, if you feel it makes sense to finish the conversation above before altering code, that would be absolutely fine of course.

> I'd also want to set the foreground color to blue to remove any possible ambiguities and still have it stand out a bit, similar how vim can highlight non-printable characters.

That sounds good.

> Blue seems to be quite readable on both dark and light backgrounds (maybe tab-highlighting can work that way in delta as well).

Right. Perhaps we'll want to make that color user-customizable? Side-by-side users will see it a lot (and delta seems to have taken it upon itself to be customizable to the last degree!)

th1000s:
Indeed, the right padding is either 1 or 2, never 0 or 1, I thought it was a stylistic choice and never questioned it :)

If at most 1 line of padding is the goal, maybe that can be inserted in the middle, before the right line-numbers. That would make all the issues here disappear.

> unnecessary work

This patchset already makes using spaces or ANSI a function parameter (`BgFillWidth`: whether to fill the line at all, containing a `BgFillMethod`: ANSI or via spaces). And spaces are required either way for the left panel.

> user-customizable

Of course. The same goes for the continuation/termination chars, maybe someone likes other symbols more or wants to only use 7-bit ASCII.

dandavison:
> If at most 1 line of padding is the goal, maybe that can be inserted in the middle, before the right line-numbers. That would make all the issues here disappear.

This sounds great. There is one thing that we need to bear in mind. Basically, I think that the fact that delta's width calculations are leaving extra columns on the right is due to this line:https://github.com/dandavison/delta/blob/e7d1e281ea539adb91a57c6945dea123fff1bd16/src/options/set.rs#L517-L518

I just wanted to highlight that line of code. Everything that follows in the rest of this comment is detail, just in case it is helpful.

-----------------------------------------------------------------

The issue being addressed there is that `less --status-column` inserts two columns on the left, and so if a user is using `less --status-column`, we have to be careful that delta does not overflow on the right.

To understand the history, the best issue to look at is this one: https://github.com/dandavison/delta/issues/115, which led to delta switching to use the ANSI fill-color-right sequence instead of manually padding with spaces. But that still leaves the issue for things like horizontal dividing lines, hence the `computed.available_terminal_width` subtraction line is still needed even after delta switched to using the ANSI fill-right sequence.

I think it may be relevant to mention that [bat](https://github.com/sharkdp/bat/issues/376) has a solution for this that we could consider introducing to delta. Bat's solution is to allow the user to declare how many columns are available to the application (bat/delta) via a +/- offset: 
```
--terminal-width <width>
            Explicitly set the width of the terminal instead of determining it automatically. If
            prefixed with '+' or '-', the value will be treated as an offset to the actual terminal
            width. See also: '--wrap'.
```

See https://github.com/sharkdp/bat/commit/dda27b253be828c9c52f040c815e3a5f45d80010 and https://github.com/sharkdp/bat/issues/376.

th1000s:
I suspected there was a history to that behavior. Also, well, it is broken again, or maybe never worked with side-by-side mode:
Set the Terminal to any odd width of less than 173 (see `echo $COLUMNS`) and run `git log -p e1ef7f54 | LESS=--status-column target/release/delta -s` with exactly that commit.

src/features/side_by_side.rs
```diff
@@ -24,62 +27,99 @@ pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
...
+}
+
+impl<T> IndexMut<PanelSide> for LeftRight<T> {
+    fn index_mut(&mut self, side: PanelSide) -> &mut Self::Output {
+        match side {
+            PanelSide::Left => &mut self.left,
+            PanelSide::Right => &mut self.right,
+        }
+    }
+}
```
dandavison, line 61: Very nice. Thanks, I didn't know about these traits.

src/features/line_numbers.rs
```diff
@@ -7,6 +7,8 @@ use crate::config;
 use crate::delta::State;
 use crate::features::hyperlinks;
 use crate::features::side_by_side;
+use crate::features::side_by_side::LeftRight;
+use crate::features::side_by_side::PanelSide::{Left, Right};
```
dandavison, line 11: Since `Left` and `Right` and `LeftRight` are used here in `line_numbers.rs`, I think that raises the questions:

1. Should they be called `Minus` and `Plus` rather than `Left` and `Right`? (because they are used in contexts other than side-by-side, and because the rest of the codebase uses "minus" and plus" with this meaning)
2. Should they be defined somewhere more central than `side-by-side.rs`?

However, even if those questions are reasonable, I am happy to leave them unresolved for now and let you decide in the main side-by-side PR.

src/paint.rs
```diff
@@ -138,18 +161,36 @@ impl<'a> Painter<'a> {
...
+            let syntax_left_right = LeftRight::new(
                 minus_line_syntax_style_sections,
-                minus_line_diff_style_sections,
-                self.minus_lines.iter().map(|(_, state)| state).collect(),
                 plus_line_syntax_style_sections,
+            );
+            let diff_left_right = LeftRight::new(
+                minus_line_diff_style_sections,
                 plus_line_diff_style_sections,
+            );
```
dandavison, line 171: If "left" basically means the same thing as "minus" here, then the two sets of equivalent variable names here are making this code harder to understand. I'll wait for your opinion on this topic before merging this PR.

dandavison (changes requested):
This looks fantastic, I really appreciate the refactoring and data structures introduced to make the code clearer and more compact.

I'm just marking this "request changes" because I have that question about whether "left/right" and "minus/plus" have genuinely distinct semantics, or whether we can simplify to use "minus/plus" in the situations where we have a data structure holding data from removed and added lines respectively (which, if side-by-side is active will be destined for left and right panels respectively).

th1000s:
Here is my attempt to have my cake and eat it too: The data structure is now `PlusMinus` (MinusPlus reads weird), it is indexed by `Plus` or `Minus`, but in a side-by-side context (which are actually most of them) it can also be index with `PanelSide::{Left, Right}`.

I can now squash this as-is into the previous commits or drop the `Left`/`Right` indexing.

dandavison:
> The data structure is now PlusMinus (MinusPlus reads weird), it is indexed by Plus or Minus, but in a side-by-side context (which are actually most of them) it can also be index with PanelSide::{Left, Right}.

Great, thanks!

> I can now squash this as-is into the previous commits or drop the Left/Right indexing.

I'd like to squash and keep the Left/Right indexing.

Looks like some tests are failing now, but once you've squashed and CI is green, I'll merge this.

```
failures:

---- features::side_by_side::tests::test_two_minus_lines_truncated stdout ----
thread 'features::side_by_side::tests::test_two_minus_lines_truncated' panicked at 'index out of bounds: the len is 0 but the index is 0', src/features/side_by_side.rs:335:18

---- features::side_by_side::tests::test_two_minus_lines stdout ----
thread 'features::side_by_side::tests::test_two_minus_lines' panicked at 'index out of bounds: the len is 0 but the index is 0', src/features/side_by_side.rs:335:18

---- features::side_by_side::tests::test_one_minus_one_plus_line stdout ----
thread 'features::side_by_side::tests::test_one_minus_one_plus_line' panicked at 'String mismatch encountered while superimposing style sections: ' ' vs 'b'', src/paint.rs:761:17


failures:
    features::side_by_side::tests::test_one_minus_one_plus_line
    features::side_by_side::tests::test_two_minus_lines
    features::side_by_side::tests::test_two_minus_lines_truncated
```

th1000s:
Ah, I mean the other left of course, should be fixed!

dandavison (approved):

dandavison:
Merged -- thanks very much for this work @th1000s, and I'm really looking forward to having the new wrapping `side-by-side` in master.

Issue #498, opened 2021-01-10
🚀 multiline with --side-by-side


jonathanpoelen:
Currently lines that are too long are truncated with `--side-by-side`. I propose `--multi-line-side-by-side` which would give like this example with icdiff:

![2021-01-10-025048_956x201_scrot](https://user-images.githubusercontent.com/1436727/104112643-b3149a80-52f1-11eb-8918-fe9384369aad.png)

nghialm269:
I suggest that we also have an option to fallback to unified view if the line is too long / or the terminal width is small.

dandavison:
Thanks @jonathanpoelen. The good news is that there's a branch in progress for this: https://github.com/dandavison/delta/pull/515

dandavison:
That's right! Thanks, I forgot about this one; will close it now. (Hi @waldyrious!)

Issue #359, opened 2020-10-20
🚀 automatic side-by-side based on column width?


aschmolck:
Hi, many thanks for writing delta -- I've long longed for something like it and am very happy to discover it.

One question: side-by-side views are great for looking at more complicated diffs but require substantially more horizontal space, so are IMO not a good default for git diff with typical terminal widths On the other hand I'm not aware to of a way to pass the `-s` flag through when using delta as git diff backend.

Assuming something like it does not already exist, how about a mechanism to provide context dependent side-by-side in git diff based on either a user-specified minimal column width or an easy way to pass flags through something like a `DELTA` env var (similar to `LESS`)?

dmaahs2017:
Or instend of using an env variable it could be specified in the gitconfig for the delta options. ex: `side-by-side = auto` or `side-by-side = someNumberOfPixels` such that when the terminal reaches the specified width it switches to side-by-side mode

aschmolck:
Yes, my thinking was something like a `side-by-side-min-columns = 120` option in gitconfig rather than an env var, I think env var makes most sense for passing through flags directly for an individual run (or git alias).

If there is interest in principle, I'd also be happy to attempt to provide a MR for this feature.

dandavison:
Hi @aschmolck @dmaahs2017, thanks! I agree that a `DELTA` env var (analogous to `LESS`, i.e. holding a string which behaves exactly like command-line flags) could indeed make sense. There have been some similar discussions, let's just loop in some others: @ryuta69 @Kr1ss-XD @navarroaxel @gibfahn any opinions there?

It's not at all ideal, but what I currently do is something like

```
git config --global delta.side-by-side true
```

(as is in the README). 

Also your idea of an auto-side-by-side mode sounds interesting; I hadn't thought of that. Would the ideal be to re-use the same option, rather than to introduce another one? I.e.

```
side-by-side = true|false|120
```

> If there is interest in principle, I'd also be happy to attempt to provide a PR for this feature.

@aschmolck great! I'll be happy to discuss designs and review code.

dmaahs2017:
> Also your idea of an auto-side-by-side mode sounds interesting; I hadn't thought of that. Would the ideal be to re-use the same option, rather than to introduce another one? I.e.
> 
> ```
> side-by-side = true|false|120
> ```
@dandavison Yeah I think reusing this option would be the best UX.

ryuta69:
I think this is a good and useful idea!

> ```
> side-by-side = true|false|120
[...]

Issue #299, opened 2020-08-21
🚀 Optional line wrapping


augustobmoura:
Sorry if this is repeated. I didn't find any similar issues.

Currently, when the line length exceeds its column, an arrow indicates that the line continues
![image](https://user-images.githubusercontent.com/5845160/90934186-3d0da680-e3d7-11ea-9d37-344a76334891.png)

I wish we could wrap the line like `bat` currently does and keep the diff highlighting.

Bat with line wrapping
![image](https://user-images.githubusercontent.com/5845160/90934279-67f7fa80-e3d7-11ea-83ae-3a416ef44c12.png)

makeworld-the-better-one:
I have a similar issue.

![2020-08-23-183028_1920x1080_scrot](https://user-images.githubusercontent.com/25111343/90990362-e4711180-e56e-11ea-9fc7-070651b918b6.png)

My lines are not wrapped at all when not in side-by-side mode, and so they go over where the line numbers would be. It'd be nice if the terminal size could be detected and then accounted for.

mnsgs:
I am seeing the same issue as referred by @augustobmoura - pretty much a standard Ubuntu 20.04 installation with the git/delta config:

```
[delta]
    line-numbers = true
    side-by-side = true
    plus-color = "#012800"
    minus-color = "#340001"
    syntax-theme = Monokai Extended

```
delta is of version 0.4.1.

dandavison:
Hi all, thanks for this. I agree implementing line wrapping like bat does would be a nice feature. I don't think I'm going to be able to work on this in the near future so I'd welcome a pull request and am also happy to discuss what the code change might involve.
 
Some notes: There's a couple of related aspects to this:

1. line wrapping respecting the line numbers panel in normal mode
2. line wrapping in side-by-side mode, in the left panel and right panel 

In normal mode we don't lose content today, it's just that it's not wrapped prettily in line-numbers mode. Personally I have the `-S` flag set in less, i.e. it's `less -RS`, so that content never wraps in normal mode (I personally find wrapped code very hard to read and prefer to scroll right when necessary).

However in side-by-side mode we currently lose content, because we can't allow the left panel to flow into the right panel's area and I did not attempt to implement line wrapping in the initial implementation of side-by-side because it would have been made the code change significantly more complicated than it already was.

[...]

PR #515, opened 2021-02-01
Add side-by-side line wrapping mode

th1000s:
If the current line does not fit into the panel, then
it is not truncated but split into multiple lines. A
wrapping symbol is placed at the end of the line.

Wrapping is limited to a certain number of lines, if this
is exceeded the line is truncated by a now highlighted
truncation symbol.

Commandline argument `-S` / `--side-by-side-wrapped`.

Also adapted `--keep-plus-minus-markers` logic, required
to calculate the exact remaining panel width.

-----------------------

Example:

wrapped

In the current state is is perfectly functional, however the limit of 5 wrapped lines is currently hard coded and I have not tested the behavior with `--tabs` yet. Also documentation beyond the one in `--help` stub has yet to be written, plus more tests.

Ping #359 #299 #498

Let me know what you think!


Kr1ss-XD:
**Thank you**, this looks ace !

I've just checked out your PR and am building it right now to do some testing :)

Kr1ss-XD:
So, one thing I noticed is, that on wrapped lines, the `plus-style`/`minus-style` is only applied until the EOL while otherwise it's applied to the while line. Please see the screenshot, and compare source code lines 5 and 183 in the right pane.

delta_sbs-wrap 1

dandavison:
Wow, thanks very much for this PR and #512 @th1000s. I'll review as soon as I can. Meanwhile, I see that they merge together without conflicts, so that combined branch will be what I'm running locally, in side-by-side mode!

dandavison:
And @Kr1ss-XD, thanks for reviewing and testing also.

th1000s:
@Kr1ss-XD Thank you for testing, I somehow managed to overlook that. It is fixed now.

@dandavison Yes, the other PR is a stand-alone bug fix, though I noticed the issue while working on this one.

dandavison:
@th1000s thanks again for the very significant contribution you've made. This question applies to both #512 and this PR, but maybe it'll help to keep the discussion here in the big PR.

------------------------------

I was thinking that right-filling with a terminal emulator instruction was an improvement because the user can resize the screen width and the color will expand/contract automatically. If I'm understanding correctly, in #512 you switch to hard-padding with spaces in the right panel, and `side-by-side-wrapped` also hard-pads on the right with spaces. This of course results in effects like below (where I issued the command with a larger width and then afterwards made my terminal narrower).

I may not have fully understood the nature of the problem you're addressing in #512 but here's my question with my current level of understanding: is it possible to have the best of both worlds, i.e. fill right with terminal emulator colors instead of hard spaces, and yet avoid the problem you're addressing in #512? (Would explicitly painting the background color of the continuation character do the trick?)

**master**

<table>
<tr><td>image</td></tr>
</table>

**this branch**

<table>
<tr><td>image</td></tr>
</table>

dandavison:
By the way, did you consider `↴` as an alternative to `↵` for the continuation character? I like the longer horizontal stem of `↵` but on the other hand I find that it sort of "points backwards" whereas `↴` perhaps points more in the direction of where the text is going?  Anyone got any views on this?

th1000s:
> #512

Ah, thank you, now I understand the reasoning behind the ANSI sequence.

Only emitting color-filling when the line is not wider than the panel would help with the green background highlighting extending beyond it.

However this would result in shorter lines still having their highlighting extend beyond the panel, resulting in a jagged right edge:

end

But since these +/- lines are processed as one block, padding with spaces or with color-filling could depend on there being one or more lines which touch the right edge. Any plus-block containing a right-edge-touching line would stand out a bit (feature!) and not wrap nicely on terminal resizes, but otherwise have a consistent look.

> ↵ / ↴

I haven't given this detail much thought, I just picked the first angled arrow I found :)

But this might combine nicely with adding alignment to the wrapped line. Most of the time just a few characters do not fit onto the line (as visible in Kr1ss' screenshot). These could be right aligned and be marked with `↴` - as in "the rest is right down here".

When the wrapped line is wider, then it starts all the way on the left again, indicated by e.g. `↵`.

Two different wrapping symbols are a less obtrusive way to indicate a different alignment than writing `...`/`>>` or so into the line itself.

Also and FYI, the truncation symbol is inverted to indicate that some information was lost, which wrapping should actually prevent.

dandavison:
Hi @th1000s, I just want to check that you're not waiting for anything from me. I'm thinking that the next step here is that you're planning to rework the #512 changes to use the ANSI sequence, does that sound right? And possibly also some work on the alignment details and continuation characters as you describe above. For my part, I will review the code more, and at some point start a discussion in this thread about what the defaults should be and whether the new code can basically replace the old (but that needn't block the ANSI sequence and alignment details work).

src/paint.rs
```diff
@@ -140,13 +142,100 @@ impl<'a> Painter<'a> {
...
+            // Only set `wrap_lines` to true if wrapping is wanted and lines which are
+            // too long are found.
+            // If so, remember the calculated line width and which of the lines are too
+            // long for later re-use.
+            let (wrap_lines, line_width, syntax_too_long, diff_too_long) =
+                if self.config.side_by_side_wrapped {
+                    let line_width =
+                        wrap::available_line_width(&self.config, &self.line_numbers_data);
+
+                    let (wrap_syntax, syntax_style_lines_too_long) = wrap::minus_plus_too_long(
```
dandavison, line 154: Suggesting:

```rust
let (should_wrap, long_lines) = wrap::find_long_lines(...
```

src/paint.rs
```diff
@@ -140,13 +142,100 @@ impl<'a> Painter<'a> {
             Self::get_diff_style_sections(&self.minus_lines, &self.plus_lines, self.config);
 
         if self.config.side_by_side {
+            // Only set `wrap_lines` to true if wrapping is wanted and lines which are
+            // too long are found.
+            // If so, remember the calculated line width and which of the lines are too
+            // long for later re-use.
```
dandavison, line 148: I believe that the two calls below to `wrap::minus_plus_too_long` will always return the same values. That is, the following assertions hold (after adding a `PartialEq` implementation to `PlusMinus`):

```rust
assert!(wrap_syntax == wrap_diff);
assert!(syntax_style_lines_too_long.minus == diff_style_lines_too_long.minus);
assert!(syntax_style_lines_too_long.plus == diff_style_lines_too_long.plus);
```

The reasons these are the same is that the `*_style_sections` data structures are arrays of `(style, substring)` tuples, and the concatenation of the `substring`s is the same in both cases. I.e. `*_syntax_style_sections` and `*_diff_style_sections` represent the same line, broken up in different ways.

Since these are (I believe) guaranteed to be the same at runtime, I think it might make the code easier to understand if we just have one variable name for each quantity.

dandavison (commented):
I haven't reviewed all the code yet but here are some thoughts on the bit that I've been through so far. It's really nice work, and I really appreciate you having grappled with one of the most complex parts of the codebase.

dandavison:
I'd vote for renaming `features/wrap.rs` => `features/side_by_side_wrap.rs`.

dandavison:
I think that users would prefer the new behavior as the default, rather than the old.  If we expose `wrap_max_lines` as a command line option `side-by-side-max-wrapped-lines`, then am I right in thinking that we can have `side-by-side` invoke the new code, and a user who wants the old behavior can get that with `side-by-side-max-wrapped-lines = 0`?

src/features/wrap.rs
```diff
@@ -0,0 +1,782 @@
...
+    // Process blocks according to alignment and build a new alignment.
+    // Wrapped lines get assigned the state HunkMinusWrapped/HunkPlusWrapped.
+    for (minus, plus) in alignment {
+        let (minus_extended, plus_extended) = match (minus, plus) {
+            (Some(m), None) => {
+                assert_alignment!(m, m_expected, "l (-)");
+
+                let (minus_start, extended_to) = wrap_text_and_style(
+                    &mut new_minus,
+                    minus_iter.next().expect("bad alignment l (-)"),
```
dandavison, line 312: I think this is going to sound a bit odd, but would you mind using `unwrap_or_else(|| panic!(...))` instead of `expect()`? It's a local style rule that I've been following.

I'm prepared to be shouted down  :), but the reason is that I personally really dislike the idiom `.expect("something is wrong")`. It seems to me like the only bad design decision I've encountered so far in Rust -- that clearly reads as "I expect something is wrong", when apparently the Rust language designers thought it would be understood as "I expect this operation to succeed but, if not, then something is wrong".

th1000s:
Thank you for the kind words :)
I worked off all your review notes, and regarding:

> I believe that the two calls below to wrap::minus_plus_too_long will always return the same values.

I thought so, too, but wanted to be on the safe side. So now the raw string is used to calculate the length, not one of the Style vectors derived from it.

The config options are unchanged so far, however I agree that `wrap_max_lines` should be exposed. Maybe `1` here would call the old behavior, and `0` means unlimited wrapping (and also set `max-line-length` to `0`)?

The right-align wrapping has been added. As such a line now does not start on the left I added an unstyled cell before it so it can be distinguished from wrapped whitespace. However this is not visible in the context/zero lines and a bit jarring in changed lines and I am open to suggestions.

> `expect()`

Yes, it is weird, it took me a white to finally think of this as **.expect**__success_otherwise_panic_with_**("message")**

Also, more arrows, the first in the rows are ones currently used:

↵ ↲ ⮨ ⮰ ⤶ ⮐ ↩ /  `↵ ↲ ⮨ ⮰ ⤶ ⮐ ↩ `

↴ ⮷ ⮯ 🢱 ⤵ ⤸ ⬎ /  `↴ ⮷ ⮯ 🢱 ⤵ ⤸ ⬎`

(oh, some are blue here, in the terminal they are all monochrome as on the right)

dandavison:
Hi @th1000s, I'm still very excited about the work here of course. Can you tell me where things stand and how I can help? By the way, using delta for its side-by-side mode was recently suggested in the GitHub blog here: https://github.blog/2021-03-11-scripting-with-github-cli/#make-ghyour-own so that's a nice additional context for your improvements to land in.

dandavison:
I need to spend more time with the diff here but I've taken another look and it's looking fantastic to me. It's very impressive; there's a lot for me to learn from it, it improves on the previous implementation, and I know it involved grappling with a lot of details of the codebase!

I think I can see a couple of bugs. You're probably aware of this one (the single space with no background colour before the `ol`):

Also, I can crash delta on this branch by resizing my terminal to (e.g.) 80 characters and issuing `git log -p` in the delta repo at 8ec99b868656f259b177f71c89c2c0ddea2f98d0.

```
:thread 'main' panicked at 'String mismatch encountered while superimposing style sections: '↵' vs '↴'', src/paint.rs:790:17
stack backtrace:
   0: _rust_begin_unwind
   1: std::panicking::begin_panic_fmt
   4: delta::paint::Painter::paint_zero_line
   5: delta::delta::StateMachine::handle_hunk_line
   6: delta::delta::delta
   7: delta::run_app
   8: delta::main
```

th1000s:
The space in front of the the right-aligned text was actually on purpose, a first draft to indicate where the line continues. I replaced it with `…` and keep the background. For now these inserted symbols are not following any theme and are just blue. But for some reason later on the color is removed in the the zero hunks, but kept correctly in the diff hunks. Any idea why that might be?

And the other obvious improvement would be to not wrap lines by cutting up words so eagerly (see: `bo↴ol`) , but let's leave that for later :)

dandavison:
Hi @th1000s,

> The space in front of the the right-aligned text was actually on purpose, a first draft to indicate where the line continues. 

Ah, OK great, I thought it was unlikely that you'd be unaware of it.

> For now these inserted symbols are not following any theme and are just blue. But for some reason later on the color is removed in the the zero hunks, but kept correctly in the diff hunks. Any idea why that might be?

This is just a quick guess but this code comes to mind:
https://github.com/dandavison/delta/blob/e7d1e281ea539adb91a57c6945dea123fff1bd16/src/paint.rs#L536-L572

dandavison:
@th1000s I've pushed a commit coalescing the two features into a single feature named `side-by-side` (thus making the new wrap code paths take over), and adding the following (provisionally named) options to the CLI:

```
--side-by-side-wrap-max-lines <side-by-side-wrap-max-lines>
    Maximum number of wrapped lines to display in side-by-side mode. Any content that still does not fit will be
    truncated [default: 3]
--side-by-side-wrap-symbol <side-by-side-wrap-symbol>
    Symbol indicating that a line has been wrapped in side-by-side mode [default: ↵]

--side-by-side-wrap-right-percent <side-by-side-wrap-right-percent>
    Threshold for right-aligning wrapped content in side-by-side mode. If the length of remaining wrapped
    content, as a percentage of the panel width, is less than this quantity then it will be displayed right-
    aligned [default: 37.0]
--side-by-side-wrap-right-symbol <side-by-side-wrap-right-symbol>
    Symbol indicating that a line has been wrapped and that the subsequent content is displayed right-aligned
    [default: ↴]
--side-by-side-wrap-right-align-symbol <side-by-side-wrap-right-align-symbol>
    Symbol displayed in front of right-aligned wrapped content [default: …]
```

dandavison:
Any thoughts on the option names and defaults? On reflection `side-by-side-wrap-right-align-symbol` maybe isn't the right name. Maybe something like `side-by-side-wrap-right-prefix-symbol`?

th1000s:
Here are my suggestions, using the same prefix for all 3 right-align arguments:

`--side-by-side-wrap-max-lines` (keep)
`--side-by-side-wrap-symbol`  (keep)
`--side-by-side-wrap-right-align-percent`  (was `--side-by-side-wrap-right-percent`)
`--side-by-side-wrap-right-align-wrap-symbol` to mirror the name of the 2nd argument  (was `--side-by-side-wrap-right-symbol`)
`--side-by-side-wrap-right-align-prefix-symbol` (was `--side-by-side-wrap-right-align-symbol`)

src/config.rs
```diff
@@ -244,7 +254,19 @@ impl From<cli::Opt> for Config {
...
-            truncation_symbol: "→".to_string(),
+            truncation_symbol: format!("{}→{}", ansi::ANSI_SGR_REVERSE, ansi::ANSI_SGR_RESET),
+            wrap_config: side_by_side_wrap::WrapConfig {
+                wrap_symbol: opt.side_by_side_wrap_symbol,
+                wrap_right_symbol: opt.side_by_side_wrap_right_symbol,
+                right_align_symbol: opt.side_by_side_wrap_right_align_symbol,
+                // TODO, support multi-character symbols, and thus store
+                // right_align_symbol_len here?
+                use_wrap_right_permille: {
+                    let percent = opt.side_by_side_wrap_right_percent.clamp(0.0, 100.0);
```
dandavison, line 265: I wanted to use `clamp` but wouldn't it mean changing the delta build environment to use nightly rust? I thought that that might cause problems with CI/CD and package manager build systems.
```
error[E0658]: use of unstable library feature 'clamp'
```

dandavison:
I like your consistency improvements to the option names. One thought -- do you think perhaps the word "align" is superfluous? We could still have a shared prefix but keep it shorter as

`--side-by-side-wrap-max-lines`
`--side-by-side-wrap-symbol`
`--side-by-side-wrap-right-percent`
`--side-by-side-wrap-right-wrap-symbol`
`--side-by-side-wrap-right-prefix-symbol`

src/config.rs
th1000s: `clamp()` became stable with 1.50, or do you want to target a specific rustc version?

dandavison: Ah sorry, I was confused, my local compiler just needed updating.

src/features/side_by_side.rs
```diff
@@ -534,7 +534,7 @@ pub mod tests {
         let config = make_config_from_args(&[
             "--side-by-side",
             "--side-by-side-wrap-max-lines",
-            "0",
+            "1",
```
dandavison, line 537: 👍

dandavison:
Just to confirm, you've fixed this, correct? (I wasn't able to reproduce it on the latest commit).

> Also, I can crash delta on this branch by resizing my terminal to (e.g.) 80 characters and issuing git log -p in the delta repo at 8ec99b8.

th1000s:
Yes, the crash is fixed now. And apart from ensuring the user-provided characters always have a display width of 1 this fixes another bug with the n:m alignment.

joshdholtz (approved):
I don’t know anything about rust but I build this from source and I’m using this now with the GitHub CLi tool and this is 🔥 

```sh
gh config set pager '/Users/josh/Projects/delta/target/release/delta -s --side-by-side-wrap-max-lines=100'
```

Really appreciate the work put into this PR and to delta 🙌  Again, not sure about the code quality but this gets a 👍 from me as a user

hwo411:
Hi everyone, sorry, maybe not the best place for this question, but it's kinda related to the situation with the long lines and people in this MR are familiar with the related parts of the code.

Is there any way right now to scroll the long lines left and right in the side-by-side view? I searched docs and issues and didn't find this question and the answer to it. To me it'd be a preferable way even over this nice wrapping solution. If there is no such option, I'll then put a feature request then. Thank you in advance.

dandavison:
> Is there any way right now to scroll the long lines left and right in the side-by-side view? 

Hi @hwo411, yes there is. It's described like this in the README:

> If the truncation is a problem, one approach is to set the width of Delta's output to be larger than your terminal (e.g. `delta --width 250`) and ensure that less doesn't wrap long lines (e.g. `export LESS=-RS`); then one can scroll right to view the full content. (Another approach is to decrease font size in your terminal.)

hwo411:
@dandavison thank you! Unfortunately, it's not a solution to my problem. I'd like to keep the side-by-side delimiter in the middle of the screen and be able to scroll right and left in each view (either together or separately). The solution in README allows only to move the side-by-side delimiter to the right side of the screen.

In other words, given the files with the contents `ABCD` and `QWER` and 3 character-wide panes, I want to be able to do the following:

From the screen `|AB→|QW→|` I want to press some key and then see `|←CD|←ER|`. Then I want to press another key and get back to  `|AB→|QW→|`. That would be the perfect solution for me. Is it possible now (and actually is it even possible at all?).

ri-aje:
any chance this can be merged soon? not being able to view full diffs is a major pain point. thanks.

dandavison:
This is very exciting! Let me know when you're not aware of anything more that needs addressing.

dandavison:

I think you've mentioned this elsewhere -- is respecting word boundaries in wrapping computations something that you feel should be a goal for future PRs (worth creating an issue for)?

dandavison:
Is it intended that when I deliberately specify an invalid two-character wrap symbol, I don't get the default but rather a different sort of arrow on a reverse-color background? One might expect `try_parse` to fall back to the default?

```
git -c delta.side-by-side=true -c delta.wrap-right-symbol='<<' diff master...
```

src/cli.rs
```diff
@@ -7,12 +7,25 @@ use structopt::clap::AppSettings::{ColorAlways, ColoredHelp, DeriveDisplayOrder}
...
+use unicode_segmentation::UnicodeSegmentation;
 
 use crate::bat_utils::assets::HighlightingAssets;
 use crate::bat_utils::output::PagingMode;
 use crate::git_config::{GitConfig, GitConfigEntry};
 use crate::options;
 
+pub const INLINE_SYMBOL_WIDTH_1: usize = 1;
+
+fn ensure_display_width(arg: &str) -> Result<String, String> {
```
dandavison, line 19: This could be called `ensure_inline_symbol_display_width`, and it would be nice to have a comment indicating what an "inline symbol" is.

src/cli.rs
```diff
@@ -497,6 +516,31 @@ pub struct Opt {
     #[structopt(long = "line-numbers-right-style", default_value = "auto")]
     pub line_numbers_right_style: String,
 
+    /// Maximum number of wrapped lines to display. Any content that still does not fit will be truncated.
+    /// A value of 0 means no limit. Only used in side-by-side mode.
+    #[structopt(long = "wrap-max-lines", default_value = "3")]
+    pub wrap_max_lines: usize,
```
dandavison, line 522: I still have a slight concern that some users will experience a mental off-by-one here. I'm not expressing a hard opinion here, just opening discussion. My brain at least partly expects

- `wrap-max-lines=1`: one normal line and one "wrapped" line, so 2 lines total
- `wrap-max-lines=0`: just one normal line, no wrapped lines. In other words: disable wrapping
- `wrap-max-lines=-1`: there is no maximum to the number of wrapped lines

src/features/side_by_side.rs
```diff
@@ -118,15 +169,10 @@ pub fn paint_minus_and_plus_lines_side_by_side<'a>(
...
-                Some(i) => states_left_right[Right][i],
+                Some(i) => &states_left_right[Right][i],
                 None => &State::HunkPlus(None),
             },
             line_numbers_data,
-            if config.keep_plus_minus_markers {
-                Some(config.plus_style.paint("+"))
-            } else {
-                None
-            },
```
dandavison, line 129: 👍 I see there's some nice cleanup related to `painted_prefix` in this module.

src/features/side_by_side_wrap.rs
```diff
@@ -0,0 +1,1115 @@
...
+                    // Do not count the '\n': + 0
+                    curr_line.push_and_set_len((*next_style, *nl), new_len + 0);
+                    stack.pop();
+                    false
+                }
+                _ => true,
+            }
+        } else if new_len == max_len + 1 && stack.is_empty() {
+            // If the one overhanging char is '\n' then keep it on the current line.
+            if !text.is_empty() && *text.as_bytes().last().unwrap() == b'\n' {
```
dandavison, line 156: Is `text.ends_with('\n')` a good alternative here?

src/features/side_by_side_wrap.rs
```diff
@@ -0,0 +1,1115 @@
...
+    // Done if wrapping adds exactly one line and this line is less than the given
+    // permille wide. Also change the wrap symbol at the end of the previous (first) line.
+    if result.len() == 1 && curr_line.has_text() {
+        let current_permille = (curr_line.text_len() * 1000) / max_len;
+
+        let pad_len = max_len.saturating_sub(curr_line.text_len() + INLINE_SYMBOL_WIDTH_1);
+
+        if wrap_config.use_wrap_right_permille > current_permille && pad_len > INLINE_SYMBOL_WIDTH_1
+        {
+            const SPACES: &str = "        ";
```
dandavison, line 203: Maybe clarify what the significance of 8 spaces is here?

src/features/side_by_side_wrap.rs
```diff
@@ -0,0 +1,1115 @@
...
+
+    let max_len = line_width + LINEPREFIX.len();
+
+    // The current line being assembled from the input to fit exactly into the given width.
+    // A somewhat leaky abstraction as the fields are also accessed directly.
+    struct CurrLine<'a, S: Default> {
+        line_segments: Vec<(S, &'a str)>,
+        len: usize,
+    }
+    impl<'a, S: Default> CurrLine<'a, S> {
```
dandavison, line 77: For my own education -- is the reason this type and its associated methods are declared inside the function definition that we want to ensure that the type parameters supplied to the generic types are "tied together", i.e. that `CurrLine<S>` is only defined for `S` such that  `wrap_line<S>` is defined? Is declaring `CurrLine` at the top level (along with `LINEPREFIX`) attractive to you at all? (the advantage I see would be to keep the function body shorter and simpler)

(By the way, I don't use Rust professionally, and I've never really used C++ or another language with generics, so you are going to be much better informed about the facts and what's standard style than I am.)

src/features/side_by_side_wrap.rs
```diff
@@ -0,0 +1,1115 @@
...
+            if !text.is_empty() && *text.as_bytes().last().unwrap() == b'\n' {
+                // Do not count the included '\n': - 1
+                curr_line.push_and_set_len((style, text), new_len - 1);
+                false
+            } else {
+                true
+            }
+        } else {
+            true
+        };
```
dandavison, line 165: `wrap_line` is a long, complex function. I think that, personally, I would factor out some helper functions such as `get_must_split` here (we spend 30 lines computing this boolean), and I think `split()` and `align()` methods below.

In conjunction with my comment above about the struct definition inside the function body, perhaps what I'm wondering is -- could it make sense to have a top level `LineWrapper` struct, holding the mutable data used in the wrapping algorithm, with `wrap_line` an associated function and additional helper functions `get_must_split()`, `split()`, and `align()`. I'm thinking that the actual wrapping algorithm would be much more apparent with a refactoring like that.

src/features/side_by_side_wrap.rs
```diff
@@ -0,0 +1,1115 @@
...
+                let this_line = &text[..byte_split_pos];
+                line_segments.push((style, this_line));
+                &text[byte_split_pos..]
+            };
+            stack.push((style, next_line));
+
+            line_segments.push((symbol_style, &wrap_config.left_symbol));
+            result.push(line_segments);
+
+            curr_line = CurrLine::reset();
```
dandavison, line 189: This is what I'm suggesting could be `split()`

src/features/side_by_side_wrap.rs
```diff
@@ -0,0 +1,1115 @@
...
+            }
+
+            right_aligned_line.push((symbol_style, &wrap_config.right_prefix_symbol));
+
+            // skip LINEPREFIX which `CurrLine::reset()` adds
+            right_aligned_line.extend(curr_line.line_segments.into_iter().skip(1));
+
+            curr_line.line_segments = right_aligned_line;
+
+            // curr_line.len not updated, as only 0 / 1 / > 1 is required now
```
dandavison, line 230: And this is what I was thinking might be `align()`.

dandavison:
Ok, I've done a review pass and I'll hand over to you now. Really exciting! I think delta might be about to possess humanity's most sophisticated side-by-side diff view!

th1000s:
> image
> 
> I think you've mentioned this elsewhere -- is respecting word boundaries in wrapping computations something that you feel should be a goal for future PRs (worth creating an issue for)?

Absolutely, `--wrap-at-chars` is a must. Plus a limit as with right align and other possible customization options.

> `git -c delta.side-by-side=true -c delta.wrap-right-symbol='<<' diff master...`
> 
> image

It works on the command line, it seems as though `parse(try_from_str ..)` is ignored when reading values from the git config, is there some `make_feature()` magic missing? I'll leave that for you do debug :)

th1000s:
And more arg parsing issues: Adding support for signed integer / `isize` leads to an error inside the `builtin_feature!` macro where an `Into<OptionValue> for i32` is missing, see my [isize branch](https://github.com/th1000s/delta/tree/isize_arg) - note that `i32s` are practically unused in delta!

Or, instead of a magic -1 value for `--wrap-max-lines` this could become a string argument and "unlimited" the explicit name for never ending wrapping.

dandavison:
> It works on the command line, it seems as though parse(try_from_str ..) is ignored when reading values from the git config, is there some make_feature() magic missing? I'll leave that for you do debug :)

Ok! That's fine then, we can ignore that.

> Or, instead of a magic -1 value for --wrap-max-lines this could become a string argument and "unlimited" the explicit name for never ending wrapping.

That would be fine with me, I'm certainly not attached to the magic -1.

th1000s:
A few touch-ups, simplified Left/Right more, hit a nightly rustc crash, and it looks good to me as well! :)

th1000s:
Very nice, and thank you for your help!

Before the next release the background color should be extended via ANSI codes again, see #727 and  #740.

fakhrulhilal:
Any idea when will it be released? This the feature that I've been waiting for. Delta is nice tool to compare many files in a scrollable screen. At least tell me how to build the cli tool.

Cheers.

dandavison:
Hi @fakhrulhilal, I agree, this is a very exciting addition to delta. We're pretty close to release I think. If you'd like to try out the unreleased master branch that would be great. The [instructions to build](https://github.com/dandavison/delta#build-delta-from-source) are in the README; it's just the standard process for building a Rust executable. If you have any feedback on released or unreleased features, please do let us know.

PR #523, opened 2021-02-12
Allow user to specify tab replacement string

### LINKED ISSUE #522: 🚀 make whitespace changes visible as such
Hi, at first glance I couldn't find any open or closed issues on this topic. Please excuse me if I missed one.

Recently, while viewing a particular diff, I noticed that every indented line was highlighted as a difference, even though the code itself was not changed, other than being indented by _spaces_ vs. _tabs_. (PKGBUILD's are interpreted by bash, so indentation doesn't matter).

delta_whitespace1

Therefore, I think it might be useful to have an option which allows displaying whitespace characters using one or more user-defined glyphs, and/or applying the `delta.{plus,minus}-emph-style` settings to whitespace changes.

In the case that the characters to display tabs were be defined as `'‣·'` and tab size as `4`, I would imagine something like :

delta_whitespace2

dandavison:
Fixes #522 cc @Kr1ss-XD

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -470,10 +470,16 @@ pub struct Opt {
     #[structopt(short = "w", long = "width")]
     pub width: Option<String>,
 
-    /// The number of spaces to replace tab characters with. Use --tabs=0 to pass tab characters
-    /// through directly, but note that in that case delta will calculate line widths assuming tabs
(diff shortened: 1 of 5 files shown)


dandavison:
Stale

Issue #237, opened 2020-07-05
🐛 `--navigate` breaks `git log --oneline` and empty `git diff`


SimenB:
Clean repo and running `git diff`

![image](https://user-images.githubusercontent.com/1404810/86542696-3a82ea80-bf18-11ea-9f26-0f094ad3a92b.png)

Running `git log --oneline` gives same error.

Note that in both cases hitting `Enter` will make it print out the correct stuff (which makes #234 really obvious)

dandavison:
Hi @SimenB, thanks for this report also. I am not sure yet what to do about the way `less --pattern` behaves, other than to recommend using `--navigate` only when needed -- ideas welcome!

SimenB:
For the `diff` case, could you detect that it's empty and skip the pattern? For the log case, somehow detect you did no replacements (or whatever you do), so you know the pattern doesn't exist?

dandavison:
The latest release of delta allows `navigate` to be activated via an environment variable, e.g.:

```
DELTA_NAVIGATE=1 git diff
```

This is motivated by the fact that leaving `navigate` active all the time is problematic; please see the README section https://github.com/dandavison/delta#navigation-keybindings-for-large-diffs

cben:
I've found some less tricks: https://superuser.com/a/1626732/33415
If instead of `less --pattern=foo`, you do `less '+r/foo|$', that behaves a bit better, and at least _shows the text_ before the annoying "Pattern not found (Enter)".
It's still not perfect, so :+1: to the env var solution.

Another thing aggravating this is the navigate regexp only matches git mode output.
It doesn't match `comparing:` from regular `delta FILE1 FILE2`, nor plain unified-diff `^---` that you see in `--color-only` mode.

dandavison:
@cben thanks very much for posting this. So do you think a change to delta code is warranted? Can you help me make sure I'm getting this right? I've made a very quick PR: https://github.com/dandavison/delta/pull/526 The diff there is

```diff
-                    process.args(&["--pattern", &navigate::make_navigate_regexp(&config)]);
+                    process.args(&[format!("+r/{}", navigate::make_navigate_regexp(&config))]);
```

Or did you intend for delta to use the `|$` part also? That's not behaving as desired for me, e.g. this does not seem to navigate on pressing `n` and `N`:

```
git log -p | less '+r/^commit|$'
```

I have some tangential questions that you might be able to help with:

[...]

PR #526, opened 2021-02-17
237 navigate improvement

dandavison:
cc @cben

### DIFF
--- a/src/bat_utils/output.rs
+++ b/src/bat_utils/output.rs
@@ -123,7 +123,9 @@ impl OutputType {
                     p
                 };
                 if config.navigate {
-                    process.args(&["--pattern", &navigate::make_navigate_regexp(&config)]);
+                    // "+" behaves approximately as if the subsequent keys were entered interactively.
+                    // "r" (repaint the screen) before "/" (search) shows the diff before "Pattern not found" in the case of no match.
+                    process.args(&[format!("+r/{}", navigate::make_navigate_regexp(&config))]);
                 }
                 Ok(process
                     .env("LESSANSIENDCHARS", "mK")
--- a/src/bat_utils/output.rs
+++ b/src/bat_utils/output.rs
@@ -10,7 +10,6 @@ use std::process::{Child, Command, Stdio};
 use super::less::retrieve_less_version;
 
 use crate::config;
-use crate::features::navigate;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[allow(dead_code)]
@@ -44,7 +43,7 @@ impl OutputType {
     fn try_pager(
         quit_if_one_screen: bool,
         pager_from_config: Option<&str>,
-        config: &config::Config,
+        _config: &config::Config,
     ) -> Result<Self> {
         let mut replace_arguments_to_less = false;
 
@@ -122,11 +121,6 @@ impl OutputType {
                     p.args(args);
                     p
                 };
-                if config.navigate {
-                    // "+" behaves approximately as if the subsequent keys were entered interactively.
-                    // "r" (repaint the screen) before "/" (search) shows the diff before "Pattern not found" in the case of no match.
-                    process.args(&[format!("+r/{}", navigate::make_navigate_regexp(&config))]);
-                }
                 Ok(process
                     .env("LESSANSIENDCHARS", "mK")
                     .stdin(Stdio::piped())
--- a/src/bat_utils/output.rs
+++ b/src/bat_utils/output.rs
@@ -10,6 +10,7 @@ use std::process::{Child, Command, Stdio};
 use super::less::retrieve_less_version;
 
 use crate::config;
+use crate::features::navigate;
 
 #[derive(Debug, Clone, Copy, PartialEq)]
 #[allow(dead_code)]
@@ -43,7 +44,7 @@ impl OutputType {
     fn try_pager(
         quit_if_one_screen: bool,
         pager_from_config: Option<&str>,
-        _config: &config::Config,
+        config: &config::Config,
     ) -> Result<Self> {
         let mut replace_arguments_to_less = false;
 
@@ -121,6 +122,13 @@ impl OutputType {
                     p.args(args);
                     p
                 };
+                if config.navigate {
+                    if let Ok(hist_file) =
+                        navigate::copy_less_hist_file_and_append_navigate_regexp(config)
+                    {
+                        process.env("LESSHISTFILE", hist_file);
+                    }
+                }
                 Ok(process
                     .env("LESSANSIENDCHARS", "mK")
                     .stdin(Stdio::piped())
--- a/src/features/navigate.rs
+++ b/src/features/navigate.rs
@@ -1,4 +1,4 @@
-use std::io::Write;
+use std::io::{Error, ErrorKind, Write};
 use std::path::PathBuf;
 
 use crate::config::Config;
@@ -41,8 +41,14 @@ fn make_navigate_regexp(config: &Config) -> String {
 // history file so, for example, a (non-navigate) search performed in the delta less process will
 // not be stored in history.
 pub fn copy_less_hist_file_and_append_navigate_regexp(config: &Config) -> std::io::Result<PathBuf> {
-    let xdg_dirs = xdg::BaseDirectories::with_prefix("delta")?;
-    let delta_less_hist = xdg_dirs.place_data_file("lesshst")?;
+    let delta_less_hist_file = if !cfg!(windows) {
+        xdg::BaseDirectories::with_prefix("delta")?.place_data_file("lesshst")?
+    } else {
+        let mut path = dirs_next::home_dir()
+            .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't determine home dir"))?;
+        path.push(".delta.lesshst");
+        path
+    };
     let initial_contents = "\
 .less-history-file:
 .search
@@ -54,12 +60,12 @@ pub fn copy_less_hist_file_and_append_navigate_regexp(config: &Config) -> std::i
         initial_contents
     };
     writeln!(
-        std::fs::File::create(&delta_less_hist)?,
+        std::fs::File::create(&delta_less_hist_file)?,
         "{}\"{}",
         contents,
         make_navigate_regexp(config)
     )?;
-    Ok(delta_less_hist)
+    Ok(delta_less_hist_file)
 }
 
 // LESSHISTFILE
--- a/src/features/navigate.rs
+++ b/src/features/navigate.rs
@@ -1,3 +1,8 @@
+use std::io::{Error, ErrorKind};
+use std::io::Write;
+use std::path::PathBuf;
+
+use crate::config::Config;
 use crate::features::OptionValueFunction;
 
 pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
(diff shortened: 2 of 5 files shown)


src/bat_utils/output.rs
```diff
@@ -123,7 +123,7 @@ impl OutputType {
                     p
                 };
                 if config.navigate {
-                    process.args(&["--pattern", &navigate::make_navigate_regexp(&config)]);
+                    process.args(&[format!("+r/{}", navigate::make_navigate_regexp(&config))]);
```
cben, line 126: Since the flag is pretty obscure, consider adding a comment e.g.
> "+" behaves approximately as-if you pressed those keys interactively.
> "r" (repaint the screen) before "/" (search) is helpful to show the diff before "Pattern not found" in case of no match.

dandavison: Done, thanks!

dandavison:
@cben @Kr1ss-XD @ronjouch @kevinhwang @SimenB would any of you have time to test out this branch? It's a potential improvement to `navigate` (thanks to @cben https://github.com/dandavison/delta/issues/237#issuecomment-780654036). The hope is that on this branch, you can have `navigate = true` always set in your config and it should not have any bad consequences. To test you'd have to build delta from source on this branch: the instructions are 3 lines long and are [here](https://github.com/dandavison/delta#build-delta-from-source). Basically you install rust, run `cargo build --release` and then use the executable at `./target/release/delta`.

What @cben suggested is that, instead of delta invoking `less` with the `--pattern` argument (which causes the annoying `Pattern not found  (press RETURN)` error and can fill the screen with `~` lines), what we can do is append our desired `navigate` search regexp to the less history file before starting `less`. One way or another, this seems to give us desirable behaviour, e.g., you hit `n` or `N` and it will use the navigate regexp, and we don't get the pointless screen of `~`s.

Now it seems to me that one issue with that solution is that delta would be polluting the user's less history with a regexp that is an implementation detail of delta. Therefore, what I am currently doing in this branch is appending the regexp to a _copy_ of the user's less history file, and telling `less` to use that copy (via the `LESSHISTFILE` env var).

I think this is working, but I'd appreciate it it if you could try to break it. (The test suite does not currently perform integration tests with `less`, although that is certainly possible.)

A downside of the current solution in this branch is that anything you do in the `less` invoked delta will no longer be preserved in your less history! Not in the real less, nor the next time you run delta. So, I would also appreciate ideas there -- e.g. how much does this matter, is this version shippable as-is and that's something we think about in the future, or how can we solve all of this more perfectly.

Kr1ss-XD:
Sure ! Thanks for the ping. :+1:

src/features/navigate.rs
```diff
@@ -28,6 +32,68 @@ pub fn make_navigate_regexp(config: &Config) -> String {
...
+    let contents = if let Some(hist_file) = get_less_hist_file(&home_dir) {
+        std::fs::read_to_string(hist_file).unwrap_or(initial_contents)
+    } else {
+        initial_contents
+    };
+    let mut hist_file = home_dir;
+    hist_file.push(".delta.lesshst");
+    writeln!(
+        std::fs::File::create(&hist_file)?,
+        "{}\"{}",
```
dandavison, line 59: @cben at the moment this just does the naive append. Can you show how to break this with `--set-marks` or similar?

src/features/navigate.rs
```diff
@@ -28,6 +32,68 @@ pub fn make_navigate_regexp(config: &Config) -> String {
...
+            Some(PathBuf::from(path))
+        }
+        Err(_) => {
+            // The user is using the default less histfile location.
+            let mut hist_file = home_dir.clone();
+            hist_file.push(if cfg!(windows) {
+                "_lesshst"
+            } else {
+                ".lesshst"
+            });
```
dandavison, line 91: If anyone were able to test on Windows that would be awesome.

Kr1ss-XD:
@dandavison The search pattern seems to be wrong - at least it is on my setup; `n` jumps to arbitrary lines here, I can't spot a pattern though. No idea yet if that's to do with my `less` or `delta` config.

dandavison:
> EDIT/ Turns out it must be some setting in my lesskey file.

Right, OK. Personally I was scratching my head for a while until I realized I had the `-r` flag set in my `DELTA_PAGER`. Changing that to `-R` fixed it.

In general  if you encounter problems, temporarily getting rid of all less config (`lesskey`, and `unset LESS PAGER DELTA_PAGER BAT_PAGER`) might be a good idea.

And if problems persist, this might help debugging:

1. Move your current `~/.lesshst` out of the way
2. Run delta with `navigate` active and see if it works
3. Post the `~/.delta.lesshst` that delta will now have written

Kr1ss-XD:
Yeah it works fine if I set `LESSKEY= git ...` .

dandavison:
> A configurable location for `.delta.lesshst` would be nice, by the way.

Yes, sorry the current hard-coded `~/.delta.lesshst` is hacky. If we do go this route of having a delta version of the file, would following the [XDG spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) be appropriate here? I think this would this be an XDG runtime file? (It could contain sensitive data).

Kr1ss-XD:
Yeah, if I were to configure the location, I'd probaly put it in `$XDG_RUNTIME_DIR` (which is a tmpfs on my system, but that doesn't apply to all distros).
Otherwise, `$XDG_DATA_HOME` would also be okay, I think.

dandavison:
> It was actually the `LESSHISTFILE` setting in my `lesskey`. If I remove that, navigate works even with LESSKEY pointing to the correct file, and even with LESSHISTFILE set in the environment.

> I guess if there's no way for delta to undo/mask settings from lesskey files, it should clear the LESSKEY environment variable ? This would, on the other hand, also undo configuration like key bindings.

I'm tempted to say that `lesskey` has no business overriding environment variables! But I also have not considered the matter for more than a minute :) Do you know if there are good reasons to configure the env var in `lesskey` rather than in the process environment?

Kr1ss-XD:
> Do you know if there are good reasons to configure the env var in lesskey rather than in the process environment?

Well, the reason why I did it was to unclutter my shell startup. Anyways, I guess since it's a valid constellation, delta should probably be able to handle it, or at least the documentation should mention that it's an issue.

dandavison:
> since it's a valid constellation, delta should probably be able to handle it, or at least the documentation should mention that it's an issue.

Yes agreed. I just experimented also and can see the behavior you describe, i.e. that you can't override the value set in the `lesskey` file. And I see from the docs

> If a variable is specified in the system environment and also in a lesskey file, the value in the lesskey file takes precedence. Although the lesskey file can be used to override variables set in the environment, the main purpose of assigning variables in the lesskey file is simply to have all less configuration information stored in one file.

I think it could be annoying and mysterious to have less not honor the `lesskey` file under `delta --navigate`, so I think I'd prefer to say in  the docs that `delta --navigate` users who are setting `LESSHISTFILE` will need to set it as an env var rather than in the `lesskey` file.

Kr1ss-XD:
That makes sense. Thx for investigating !

dandavison:
Thanks very much for testing @Kr1ss-XD, and @cben for the idea and research. I've merged this to master. If anyone's able to test on Windows that would be great.

Kr1ss-XD:
I'm sorry @dandavison that I didn't realize this earlier.

Tonight I've built a new binary from current master, and I am noticing that `navigate` doesn't work any more with a custom `$LESSHISTFILE` defined environment (note i.e. _not_ via `LESSKEY`).

Unsetting that variable or setting it to an empty string fixes `navigate`.

dandavison:
Hi @Kr1ss-XD, I'm failing to reproduce that. When you say `navigate` doesn't work, what behavior are you seeing? Can you give a minimal example of steps leading to the bug? Here's an example of what I'm doing:

```
$ rm -f ~/.lesshst ~/.local/share/delta/lesshst /tmp/delta.tmp.lesshst

$ export LESSHISTFILE=/tmp/delta.tmp.lesshst

$ less Cargo.toml
# Now search for "aaa" with / + aaa
# q to quit

$ cat /tmp/delta.tmp.lesshst
.less-history-file:
.search
"aaa

$ cat ~/.local/share/delta/lesshst
[bat error]: '/Users/dan/.local/share/delta/lesshst': No such file or directory (os error 2)

$ git -c delta.navigate=true log -p
# the Δ symbols are visible and navigate (n) works
# enter / + <up arrow> twice reveals last searches to be '^(commit|∆|added:|removed:|renamed:)' and before that 'aaa'

$ cat ~/.local/share/delta/lesshst
.less-history-file:
.search
"aaa
"^(commit|Δ|added:|removed:|renamed:)
.search
"aaa
```

Kr1ss-XD:
Thanks. The behavior I'm seeing is basically the same I had earlier, when I have had `LESSHISTFILE` defined in my `lesskey` file.

That means there was no search regex in the less history when I did a `git log -p` or similar, therefore `n` gave me an error message most of the times, saying "no previous regular expression (press RETURN)". Depending on the diff/log, pressing `n` sometimes jumped to arbitrary locations instead (without an error message).

Funny thing is: when following your steps from above, it _works_, and I get the same output like you did, from each of the commands.

**But** - if I do the same, except `export LESSHISTFILE=/home/kriss/.config/less/history`, and do **not** rm the file, I get the error about no previuous regex.

If I `export LESSHISTFILE=/home/kriss/.config/less/lesshst` instead, it works. Maybe the history filename needs to end with (or be) `lesshst` ?  Otherwise I've no idea, since permissions on that file aren't the issue.

dandavison:
> if I do the same, except export LESSHISTFILE=/home/kriss/.config/less/history (which exists),

Do you use `--save-marks`? Does your history file have a `.mark` section?

If you have time to experiment a bit, does the bug go away if you use a smaller history file without any `.mark` section? Can you create a minimal version of the file that causes the bug? And if you _really_ have time, I'm experimenting with some shell-script based integration tests in this PR: https://github.com/dandavison/delta/pull/536/ so if you can create a failing test by adding to that shell script that would of course be awesome.

If the `.mark` sections can't be handled sanely, one option would be to not have less history in the less that delta uses; how bad would that be for the more serious less users here?

Kr1ss-XD:
Guess I've finally found the cause. My less history contains the following :

```
.less-history-file:
.search
"depth
"log
"-z
"verify
"script
"rsync
"rsync
"ssh
.shell
"pwd
```

So there's something after the search patterns. If I remove that, the error doesn't appear. :)

dandavison:
OK, excellent, thanks very much for the example file. I don't think the format is documented and I haven't studied the source code, but it's tempting to think that the answer is for delta to append a `.search` line before its regexp. I've pushed a commit to master that does that. That seems to fix things with your example file.

src/features/navigate.rs
cben:
```
$ export LESS='--save-marks --status-column'
$ less README.md 
```
Inside less, press <kbd>d</kbd><kbd>M</kbd><kbd>a</kbd> to set a mark called `a` at end of screen.
Quit.
```
$ cat ~/.lesshst
.less-history-file:
.search
"foo
"bar
...
"various search strings...
.mark
m a 73 3692 /home/cben/term/delta/README.md
```
We can confirm the saved mark works:
```
$ less README.md 
```
Inside less press <kbd>'</kbd><kbd>a</kbd> to jump to mark `a`.  (and can see it highlighted in status column on left.)

Now use `delta --navigate`, e.g. by `git show`.  Navigation doesn't work.

cben:
yeah, it looks like sometimes `less` itself creates a 2nd `.search` section, and can recall search history from both of them.

src/features/navigate.rs
cben: ^ that was tested with the version when you merged this PR.  With current master it works :heavy_check_mark:.

Kr1ss-XD:
> if you can create a failing test by adding to that shell script that would of course be awesome.

So that's what I've got so far :  [tests/test_navigate_less_history_file](https://gist.github.com/91903c7792bb9592c76fa2101ce3c165)

This adds two more tests which fail on current release, but pass on master (since https://github.com/dandavison/delta/commit/693d71d0bb6570705ae30bb732c20fe99fa49f14).

dandavison:
And thanks very much for the instructions for manual reproduction of the `.mark` bug and testing @cben. I think this is getting more robust now.

PR #532, opened 2021-03-05
replace deprecated config options in README

KristobalJunta:
Remove deprecated *-color options in configuration examples
and replace them with new *-style variants


dandavison (approved):
Argh, was I really doing that in the first example for all that time?? Thanks very much for fixing this!

PR #536, opened 2021-03-09
Add navigate/less integration test

dandavison:


tests/test_navigate_less_history_file
```diff
@@ -18,10 +18,11 @@ DELTA="${1:-./target/release/delta} --no-gitconfig --navigate"
 unset DELTA_PAGER
 TEMPDIR="$(mktemp -d)"
 export PAGER="$TEMPDIR/less"
-install -m755 /dev/stdin "$PAGER" <<-EOF
```
Kr1ss-XD: Oh right, sorry. I've fallen into that trap before, this seems to be a GNU extension, which isn't implemented on BSD and the likes.

dandavison: Hi @Kr1ss-XD total aside / rabbit-hole but,

> this seems to be a GNU extension, which isn't implemented on BSD and the likes.

That's what I assumed, however if that were the case wouldn't it work using the GNU coreutils on BSD (MacOS):

```
$ bash -c 'echo hello | ginstall -m755 /dev/stdin /tmp/file && ls -l /tmp/file'
ginstall: skipping file '/dev/stdin', as it was replaced while being copied
$ ls -l $(which ginstall)
lrwxr-xr-x 37 dan 10 Jul  2018 /usr/local/bin/ginstall -> ../Cellar/coreutils/8.30/bin/ginstall
```

whereas the same command works in ubuntu:

```
$ docker run ubuntu bash -c 'echo hello | install -m755 /dev/stdin /tmp/file && ls -l /tmp/file'
-rwxr-xr-x 1 root root 6 Mar 10 14:56 /tmp/file
```

Perhaps the problem

> ginstall: skipping file '/dev/stdin', as it was replaced while being copied

is more to do with the behavior of the `/dev/stdin` pseudo-file on MacOS?

dandavison:
OK, let's merge this seeing as they seem to be working. I admit that this style of test could well start to give spurious failures for some reason or other, but while they are passing at least I think they're being helpful!

tests/test_navigate_less_history_file
Kr1ss-XD: That's interesting !

If I'll have the time I'm gonna spin up a quick FreeBSD VM and see how behaves on that.
I'd really like to learn about scripting in a more compatible way.

Issue #501, opened 2021-01-11
🚀 Update completions to include local files/directories


aneeshusa:
Hi, thanks for contributing to Delta, please go ahead and describe your proposal!

I discovered delta maybe a month ago and it's awesome! Thank you so much for making it, have already been spreading the word about it at work :)

After initially setting delta up for usage with git, I learned it can also be used as a `diff` replacement, e.g. `delta <minus-file> <plus-file>` and have been using that functionality as well. However, the completions for delta only appear to complete options like `--color-only`. Normally in the lack of completions the shell would allow completing filenames, but because delta does ship a completion, the result is there is no autocompletion for file/directory names for the `<minus-file>` and `<plus-file>` arguments.

So, I'd like to request that the completions be extended to also support completing file/directory names.

jonathanpoelen:
On my system with zsh, there is an auto-completion for `delta` that has nothing to do with (`/usr/share/zsh/functions/Completion/Unix/_sccs`). Maybe you are in a similar situation? I made my own auto-completion: [_delta](https://github.com/jonathanpoelen/dotfiles/blob/master/.zshcompletions/_delta) and with `_comps[delta]=_ delta` after `compinit` to replace the original completion.

dandavison:
Hi @aneeshusa @jonathanpoelen, agreed I also like to have completions working well but I've been neglecting the file that is in the delta repo.

I wanted to mention that in the next release of delta, we'll be able to  use `git -c`, e.g. something like `git -c delta.side-by-side=false diff`. It has occurred to me that it might be nice to somehow extend git completion to complete delta options after `-c`?

Kr1ss-XD:
>It has occurred to me that it might be nice to somehow extend git completion to complete delta options after -c?

:rocket: Absolutely, this would be great !

PR #540, opened 2021-03-13
Fix bash completion to include local files/dirs

foamkeen:
This PR fixes the issue #501.


dandavison (approved):
Thanks very much @foamkeen. Confirmed that this fixes the problem on bash with `completions.bash` (and that it is already working with zsh and `completions.zsh`).

PR #547, opened 2021-03-22
Silently change pager to less if $PAGER is delta

dandavison:


alerque:
I would prefer to be warned of h nonsensical config rather than having it silently fixed to a best guess that may or may not be what I want. This looks like a regression not a feature to me.

dandavison:
> I would prefer to be warned of h nonsensical config rather than having it silently fixed to a best guess that may or may not be what I want. This looks like a regression not a feature to me.

Hi @alerque! Thanks for that I very much appreciate having input / someone to discuss this with. Here's what motivated me to open this PR: https://github.blog/2021-03-11-scripting-with-github-cli/#make-ghyour-own. There, Github are suggesting to use delta with their CLI tool, which is great. Now, their tool uses `PAGER` to mean the pager for any significant output generated by their app, including diff output. So my concern is that delta needs to handle users who have set `PAGER=delta`.

alerque:
That example is a bug in `gh` and the relevant docs. It should be fixed at the source so it uses and directs people to the proper variable not papered over where symptoms crop up.

dandavison:
> That example is a bug in gh and the relevant docs.

Yes, I've been considering whether to suggest a change in `gh`. I need to think a bit more about what is correct.  I think I have been ignoring some of the implications of `delta` putting itself forwards as a value for git's `core.pager`. Some observations:

- `git` respects both `PAGER` and `GIT_PAGER`, with the latter taking priority. 
- AIUI There is no way to specify an executable to be used for handling just some sections of git's output: if delta is to handle the diff sections of git output, then it must handle everything, e.g. stuff like `git grep`. Although I suppose you could write something that registers itself as a git [difftool](https://git-scm.com/docs/git-config#Documentation/git-config.txt-difftool), calls out to _something else_ to actually compute the diff, and then colorizes it.
- So given that delta is (AFAICS) obliged to handle basically anything, then I guess it _is_ a somewhat reasonable value for `PAGER`. 
- This is how delta works today: it reads lines of stdin, does nothing to lines it doesn't think it is responsible for, does its thing with the other lines, and sends all the lines to a specialist traditional pager executable. 
- So if this is correct, and if delta is to be `PAGER`, then it must find an appropriate specialist traditional pager.

alerque:
I haven't had time to really think through the implications of each option, but at first blush regearing to be a difftool and not a pager makes much more sense to me than anything else. If it does stay as a pager then I guess it needs some configuration to not only discover `less` if it's in the path but allow other pagers to be configured as well.

dandavison:
Thanks @tpoliaw Related, above I wrote

> AIUI There is no way to specify an executable to be used for handling just some sections of git's output: if delta is to handle the diff sections of git output, then it must handle everything, e.g. stuff like git grep. 

But that's actually just not true. Git allow's us to configure the pager for individual subcommands. I've opened https://github.com/dandavison/delta/pull/610 to discuss changing the README's advice, but it's relevant to the discussion here also.  

> at first blush regearing to be a difftool and not a pager makes much more sense to me than anything else. I

@alerque, as @tpoliaw says, I think that horse has already bolted: delta adds decorations to various sections of git log output, in addition to the special handling of diffs which is of course its main purpose as you are pointing out. _But_, one could just set `pager.diff=delta` and leave `core.pager` alone, and then delta will only handle diffs.

dandavison:
Looks like we did this at the time: https://github.com/dandavison/delta/pull/531

alerque:
That looks like a good solution to me.

PR #550, opened 2021-03-28
Provide an option to cycle through delta themes using --show-themes

clnoll:
When delta is invoked with `--show-themes`, cycles through all themes in the git config, where a theme is identified as a section with a "dark" or a "light" entry. Uses the computed `is_light_mode` value to determine whether to show light themes or dark themes by default.

When either the `--dark` or `--light` flag is passed in with `--show-themes`, will additionally display the dark or light themes, respectively. When `--light` and `--dark` are passed in, shows all themes.

The user can navigate from theme to theme using the `n` and `N` navigation keys.

In addition to viewing the themes using a sample diff, users can display themes on their own diff by piping it into delta, e.g. `git show | delta --show-themes`.

That being said, I might still swap out the sample commits used in the display, if I find something that seems to exhibit the themes' characteristics more clearly.

I'd also like to add a light theme or two to themes.gitconfig, because right now it only has dark themes to display.


src/cli.rs
```diff
@@ -270,6 +271,14 @@ pub struct Opt {
     #[structopt(long = "show-syntax-themes")]
     pub show_syntax_themes: bool,
 
+    /// Show all available delta themes, each with an example of highlighted diff output.
+    /// If diff output is supplied on standard input then this will be used for the demo. For
+    /// example: `git show --color=always | delta --show-themes`.
```
dandavison, line 274: I don't think `--color=always` is needed here.

src/options/get.rs
```diff
@@ -275,4 +299,38 @@ pub mod tests {
...
+    max-line-distance = 0.6
+    light = true
+
+[delta \"light-and-dark-theme\"]
+    max-line-distance = 0.6
+    light = true
+    dark = true
+
+[delta \"not-a-theme\"]
+    max-line-distance = 0.6
```
dandavison, line 320: Tiny thing but can you use the `b#"..."#` string literal syntax that allows you not to escape the double quotes inside the string?

src/options/get.rs
```diff
@@ -38,6 +40,27 @@ where
...
+    static ref GIT_CONFIG_THEME_REGEX: Regex = Regex::new(r"^delta\.(.+)\.(light|dark)$").unwrap();
+}
+
+pub fn get_themes(git_config: Option<git_config::GitConfig>) -> Vec<String> {
+    let mut themes: Vec<String> = Vec::new();
+    for e in &git_config.unwrap().config.entries(None).unwrap() {
+        let entry = e.unwrap();
+        let entry_name = entry.name().unwrap();
+        let entry_value = entry.value().unwrap();
+        if entry_value == "true" {
```
dandavison, line 53: Do we actually need to look at the value here? (Also, if someone has somehow got both `light` and `dark` set in a theme would that theme be demo'd twice?)

dandavison (commented):
This is fantastic! Delta should definitely have this. Some tiny comments but I can see this is pretty close.

src/cli.rs
clnoll: 👍 removed.

src/options/get.rs
clnoll: Good call, it's not needed. Removed.

(And just pushed a test update and a fix for the bug that would have allowed it to be demoed twice.)

themes.gitconfig
```diff
@@ -94,6 +96,7 @@
 [delta "tangara-chilensis"]
     # author: https://github.com/clnoll
     commit-decoration-style = bold box ul "#34fd50"
+    dark = true
     file-style = omit
     file-decoration-style = none
     hunk-header-style = file line-number syntax
```
dandavison, line 102: Would it make sense to add a sentence to the instructions at the top of this file explaining that authors should include (exactly) one of the light or dark bool flags?

src/options/get.rs
clnoll: 👍 done.

src/main.rs
```diff
@@ -302,6 +312,56 @@ where
...
+    let opt = cli::Opt::from_iter_and_git_config(
+        &["", "", "--navigate", "--show-themes"],
+        &mut git_config,
+    );
+    let mut output_type =
+        OutputType::from_mode(PagingMode::Always, None, &config::Config::from(opt)).unwrap();
+    let title_style = ansi_term::Style::new().bold();
+    let writer = output_type.handle().unwrap();
+
+    for theme in &get_themes(git_config::GitConfig::try_create()) {
```
dandavison, line 339: Let's display them in alphabetical order.

src/sample_diff.rs
```diff
@@ -0,0 +1,224 @@
+pub const DIFF: &[u8; 8715] = b"\
```
dandavison, line 1: Shall we use one of the example diffs used in the README, or something else? I was going to say we should choose a shortish one, but I can see the attraction in using a diff long enough to not fit on the screen -- i.e. that it prevents the user from seeing two themes on the screen at the same time.

themes.gitconfig
dandavison: Don't worry, I've pushed a commit doing this.

dandavison (commented):
Almost there! I think just selecting the default diff is all that remains.

src/sample_diff.rs
clnoll: 👍 I switched to https://github.com/vuejs/vue/commit/7ec4627902020cccd7b3f4fbc63e1b0d6b9798cd, from the README.

src/main.rs
clnoll: Done.

dandavison (approved):
Excellent, thank you! Everything looks good. Hopefully your work here will lead to some really good themes being surfaced from delta users. (I added a link to the README section on themes to your `--show-themes` help text to try to help discoverability a bit more.)

michaelblyons:
```
[09:37:44 ~]$ delta --version
delta 0.7.1
[09:37:53 ~]$ delta --show-themes
Pattern not found  (press RETURN)
```
Am I doing something wrong?

dandavison:
Hi @michaelblyons, thank you, I don't think this is your fault. This is likely to be a bug, related to https://github.com/dandavison/delta/issues/237. Could you tell us the following information if you don't mind:

- operating system
- output of `env | grep -i less`
- output of `env | grep -i pager`
- output of `less --version`
- Contents of `~/.local/share/delta/lesshst` if it exists

dandavison:
Would you mind trying a more recent less? I'm assuming `which less` gives you `/usr/bin/less`. If you use homebrew could you do `brew install less`? (I'm not sure this is the problem.)

dandavison:
Oh! I'm sorry, I know what the problem is. You need to tell delta where the themes are, like this: https://github.com/dandavison/delta#custom-color-themes

And for our part, we need to fix it so that it gives a more helpful error message!

michaelblyons:
Oh, good. I was having some trouble with `less`, wherein `which less` suggested I was using a newly `brew install`ed 563, but `less --version` still showed 487 (vs `/usr/local/bin/less --version` showing 563).

dandavison:
Great, I'm glad it was nothing more serious than that. I wonder whether, in addition to that error message needing to be improved, this is pointing to moving some of the theme definitions into the executable itself. [The delta repo can continue to host a faster-evolving expanded collection, and sers would still be able to create custom themes using `[delta "my-theme"]` either in `~/.gitconfig` or in an include'd file.]

PR #555, opened 2021-03-30
Remove unnecessary trailing semicolon after enum

aragon999:
Thank you for creating this software :-)

I just build the latest release and got a warning, that this trailing semicolon wants to be removed.

I did a quick check as far as I understand that semicolon can (and should) be removed.
Since I do not know Rust very well (my first contribution to a Rust project :-)), please correct me if I am wrong.

### DIFF
--- a/src/options/theme.rs
+++ b/src/options/theme.rs
@@ -132,7 +132,7 @@ mod tests {
         enum Mode {
             Light,
             Dark,
-        };
+        }
         for (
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks @aragon999, I think you're right. I wonder why clippy (the rust linter) didn't complain. In any case, I'll merge when tests pass -- thanks!

PR #559, opened 2021-04-15
CI: check docs

MarcoIeni:
working example: https://github.com/LukeMathWalker/wiremock-rs/pull/66


dandavison:
Thanks very much @MarcoIeni! I've removed the triple-slash comments for now since they were just used randomly in a couple of places.

PR #563, opened 2021-04-19
Support matching GitHub remotes that don't end in `.git`

rouge8:
`.git` is optional in clones from GitHub.com.

### DIFF
--- a/src/git_config/git_config_entry.rs
+++ b/src/git_config/git_config_entry.rs
@@ -20,7 +20,7 @@ pub enum GitRemoteRepo {
 }
 
 lazy_static! {
-    static ref GITHUB_REMOTE_URL: Regex = Regex::new(r"github\.com[:/]([^/]+)/(.+)\.git").unwrap();
+    static ref GITHUB_REMOTE_URL: Regex = Regex::new(r"github\.com[:/]([^/]+)/(.+)").unwrap();
 }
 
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks @rouge8. I really appreciate you taking the time to open a PR, and don't want to look a gift horse in the mouth! But would you mind describing the bug that this fixes (in this case as there is no issue). That would help me be able to immediately assess the PR and whether or not the test suite should be catching it.

rouge8:
Sure. Delta hyperlinks commit hashes on repos cloned from github.com. Currently, that only works if the remote repository ends in `.git`, but that’s no longer necessary on GitHub; e.g. you can `git clone git@github.com:rouge8/delta`. This PR updates the regex to match those remotes as well.

dandavison:
Right, thanks very much for explaining. I wanted to be sure that it was hyperlinks that had motivated you to make this change. I didn't know about the permissiveness of github URLs but as you say, I can see that this fixes it for ` git clone git@github.com:rouge8/delta`.

dandavison:
Actually, I moved too fast there and should have followed through on my claim that I would assess the need for test coverage :) This change breaks parsing of URLs without the `.git` suffix. Test coverage and a fix added in https://github.com/dandavison/delta/pull/572

rouge8:
Oof, good catch

PR #565, opened 2021-04-21
Output paths relative to current working directory

### LINKED ISSUE #552: 🚀 Transform file paths to be relative to current directory
Git outputs file paths in various contexts (e.g. hunk headers, `--stat` output) but if you are not in the root directory of your repository these paths will not be valid as relative paths from your location. This can be annoying, for example the following workflows fail:

- Copy the relative path and try to open it from the command line using something like `vim`, `open`, `bat`, `code`, `emacsclient`, etc
- Command-click on the relative path in something like VS Code terminal

Since every line of git output is passing through delta, delta could relativize these paths thus making those workflows succeed.

dandavison:
Fixes #552

TODO:

- [ ] Make things correct when git's `--relative` is in effect
- [x] Add option to control this: Done: `--relative-paths`
- [x] Add option to control padding width when this is active: Done: this is `--diff-stat-align-width`
- [ ] Would `--max-relative-path-length` be a better name?

Out of scope for this PR:

- [ ] Process all lines in the diff stat before emitting any of them so that the max path length can be determined.


anordal:
OOOOOOOOMFG!

```
[delta]
    relative-paths = true
```

I can't believe this is the day I'm no longer using Git in pain.

Thank you for this, and thank you for Delta, @dandavison!

PR #569, opened 2021-04-22
Fix --show-syntax-themes exhausting stdin

th1000s:
If a diff is provided via stdin then the first _show_syntax_themes call
reads from it directly and uses that, but the second one falls back to
the default. Fixed by reading from stdin earlier.

--------
Also the "gruvbox-white" theme does not show up here, so I used the -light variant


dandavison (approved):
Thanks!

PR #573, opened 2021-04-24
Support new GIT_CONFIG_PARAMETERS format

dandavison:
`git -c delta.option=val ...` is currently broken for git ≥2.31.0 due to some recent changes in the format used by git for the `GIT_CONFIG_PARAMETERS` env var. See https://github.com/git/git/blob/311531c9de557d25ac087c1637818bd2aad6eb3a/Documentation/RelNotes/2.31.0.txt#L127-L130. This PR makes delta work with git < 2.31.0 and ≥2.31.0.

### DIFF
--- a/src/git_config/mod.rs
+++ b/src/git_config/mod.rs
@@ -167,32 +167,41 @@ mod tests {
         // [core]
         //     pager = env | grep GIT_CONFIG_PARAMETERS
 
-        let config = parse_config_from_env_var_value("'user.name=xxx'");
-        assert!(config.is_empty());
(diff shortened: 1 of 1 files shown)


epage:
Thanks for laying the grounwork on this!  I've gone ahead and generalized this in [git-config-env](https://github.com/crate-ci/git-config-env), based on the logic from git's source.

dandavison:
@epage ah I'm glad you came across this and thanks very much for doing that. As you see, I didn't attempt to do this very robustly here, e.g. did not tackle embedded quotes.

PR #585, opened 2021-05-03
docs: wrap link to ubuntu issue in link tag

veryspry:
### Summary

Makes this issue link a little easier to get to! :smile:


dandavison:
Ah right! Thanks, I probably thought Github was doing its link rendering magic, but forgot it's in a table.

PR #606, opened 2021-05-20
Only read major version when parsing less version

tpoliaw:
Not sure if this is the best fix for this but it does work locally and returns delta to its previous behaviour. It might be nicer to parse the whole version and then explicitly only match on the major version? Seems a bit of an overkill unless less intend to continue with point releases in future instead 581.2 being a one off special bug fix.

### DIFF
--- a/src/bat_utils/less.rs
+++ b/src/bat_utils/less.rs
@@ -8,7 +8,7 @@ pub fn retrieve_less_version() -> Option<usize> {
 fn parse_less_version(output: &[u8]) -> Option<usize> {
     if output.starts_with(b"less ") {
         let version = std::str::from_utf8(&output[5..]).ok()?;
-        let end = version.find(' ')?;
+        let end = version.find(|c: char| !c.is_ascii_digit())?;
         version[..end].parse::<usize>().ok()
     } else {
         None


tpoliaw:
Just noticed bat has made the same change in https://github.com/sharkdp/bat/commit/91a347bf6d549ba543f1b4bfdc2a05f269c975c8

dandavison:
Thanks @tpoliaw! I'll maybe copy over bat's unit test as well after merging this.

I'd really like not to have to do this less version parsing! I wonder if there's an argument that for delta it's not necessary as users really need to be using a recent version of less for other reasons -- 24-bit colors and unicode rendering bugs. If anyone really has to use an old less then the onus would be on them to add this `-X` argument.

tpoliaw:
I can add the test to this PR if you want. I hadn't seen the bat version when I did this but probably ought to have added a test anyway.

dandavison:
> I can add the test to this PR if you want.

Thanks, but no need to block this PR, we've got a green CI build so I'll merge this now. Feel absolutely free to copy over the updated bat code in a subsequent PR but if you have energies for delta work there might also be more exciting and valuable tasks :)

smac89:
Thanks for the fix guys. Looking forward to the next release!

PR #610, opened 2021-05-25
Recommend setting delta as the pager for specific git commands

dandavison:
This PR changes the way in which users are recommended to use delta: instead of recommending users to set `core.pager`, we recommend them to set delta as the pager for specific git commands (`diff`, `log`, `reflog`, `show`).

Does anyone have opinions on whether this is the right thing to do? cc @tpoliaw @th1000s @Kr1ss-XD @alerque) ref https://github.com/dandavison/delta/pull/547

The motivation for doing this is that setting `core.pager=delta` is instructing git to send all its output, from any command, even if it has nothing to do with diffs, to delta. But delta isn't a general-purpose pager, and indeed behaves very strangely if it is used as one, e.g.

```
~ echo 'hello\ncommit \nbye' | delta
hello
────────┐
commit  │
────────┘
bye
```

I don't yet have a realistic pathological example, but currently, if we have `core.pager=delta`, then delta is handling git's `help` output, and it definitely seems like `^commit ` could occur there.

Another example is that `delta` is handling `git grep` output which, again, it was never intended for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,11 @@ The most convenient way to configure delta is with a `[delta]` section in `~/.gi
 <sub>
 
 ```gitconfig
-[core]
-    pager = delta
+[pager]
+    diff = delta
+    log = delta
+    reflog = delta
+    show = delta
 
 [interactive]
     diffFilter = delta --color-only
@@ -278,13 +281,16 @@ Behind the scenes, delta uses `less` for paging. The version of `less` that come
 
 ## Configuration
 
-#### Git config files
+#### Git config file
 
-Set delta to be git's pager in your `.gitconfig`. Delta has many options to alter colors and other details of the output; `delta --help` shows them all. An example is
+Set delta to be the pager for git commands in your `.gitconfig`. Delta has many options to alter colors and other details of the output; `delta --help` shows them all. An example is
 
 ```gitconfig
-[core]
-    pager = delta
+[pager]
+    diff = delta
+    log = delta
+    reflog = delta
+    show = delta
 
 [delta]
     plus-style = "syntax #012800"


README.md
```diff
@@ -55,8 +55,11 @@ The most convenient way to configure delta is with a `[delta]` section in `~/.gi
 <sub>
 
 ```gitconfig
-[core]
-    pager = delta
+[pager]
+    diff = delta
+    log = delta
+    reflog = delta
+    show = delta
```
dandavison, line 62: Am I missing any?

alerque:
This sounds much saner approach than the previous arrangement which was easy but circular and involved silent hardcoded fallbacks. More work for the user to setup but much more likely to do the right thing in the end.

dandavison:
> This sounds much saner approach than the previous arrangement which was easy but circular and involved silent hardcoded fallbacks. More work for the user to setup but much more likely to do the right thing in the end.

Hi @alerque, I don't think this is quite the same issue as #547, but they are related. There's no circularity in the current delta recommendations: it's just that `core.pager=delta` is assigning inappropriate work to delta. A circularity arises if a user sets `PAGER=delta`, since then `delta` will invoke itself as its own pager. Nowhere does delta encourage users to do such a thing, but it is easy for people to confuse `GIT_PAGER` with `PAGER`, and furthermore Github's `gh` tool [does](https://github.blog/2021-03-11-scripting-with-github-cli/#make-ghyour-own) encourage using `PAGER=delta`. But the changes in this PR don't affect that issue. Currently, delta solves it by exiting with

```
It looks like you have set delta as the value of $PAGER.
This would result in a non-terminating recursion. delta is
not an appropriate value for $PAGER (but it is an appropriate value for $GIT_PAGER).
```

dandavison:
OK, here's a realistic pathological example:  `git grep -h commit` I think this settles the matter -- as @alerque said, it is not sane for delta to be handling arbitrary output such as `git grep`!

(This consideration applies equally to diff-so-fancy incidentally; I'm not sure when the per-command pager options got introduced to git.).

dandavison:
@scottchiefbaker 👋  is this something that diff-so-fancy has ever worried about? I think that with the d-s-f config recommended in the README I can trigger the same sort of inappropriate behavior that delta displays above, e.g.

```
git -c core.pager='diff-so-fancy | less --tabs=4 -RFX' grep -h diff
```


scottchiefbaker:
Hmmmmm... this is an interesting problem. I don't think diff-so-fancy suffers from this because we just rewrite (on-the-fly) a handful of diff specific lines to be human readable. Anything we don't recognize gets passed straight through. 

In the above example with `git grep` those lines are just ignored (passed straight through) because the filter doesn't recognize them. The output is as expected.

dandavison:
Hi @scottchiefbaker, thanks! I may not have explained myself very well. The screenshot there is output from diff-so-fancy and illustrates what seems to me to be a bug: diff-so-fancy is inserting purple horizontal lines and the string "modified: " in the middle of the user's `git grep` output. Delta also had an analogous bug until this PR. Seeing as (as far as I can see) the same bug applies to delta and diff-so-fancy, I wanted to check with you whether you agreed that it's a bug, and with the remedy. Let me know if you'd like me to open an issue in the diff-so-fancy repo.

dandavison:
> At first glance it looks like the git grep output just isn't suitable for d-s-f?

Yes, I think that's right. It's a bit of an edge case in d-s-f as far as I'm aware so I won't bother opening an issue. Perhaps a solution is for users who care to set `pager.grep=less -R` or similar.

So, example reproductions are a bit contrived, but here are a couple, based on doing `git grep -h` searches in the `diff-so-fancy` repo on current master (8ecd68f8569c4a85fc02ad12f8af3857c5afe792):

(The `-h` flag causes the file name not to be included in the `git grep` output, which has the result that the search term occurs at the beginning of some lines of output.)

## 1
Looks like this one can trigger a crash:
```
$ cd diff-so-fancy
$ git checkout 8ecd68f8569c4a85fc02ad12f8af3857c5afe792
$ git grep -h '^similarity index' | diff-so-fancy       
Use of uninitialized value $next in pattern match (m//) at /usr/local/bin/diff-so-fancy line 324, <STDIN> line 2.
```

## 2
```
$ cd diff-so-fancy
$ git checkout 8ecd68f8569c4a85fc02ad12f8af3857c5afe792
$ git grep -h -- --- | diff-so-fancy | less
```

Results in lots of output like

.

And as in the screenshot above, a `git grep -h` search for anything (e.g. `git grep -h '.' | diff-so-fancy | less`) can trigger the diff-so-fancy decorations whereas the user would perhaps not expect the grep output to be decorated even if it resembles diff output.

README.md
Kr1ss-XD: Sorry for the late response @dandavison - maybe you could add `stash` ?

PR #627, opened 2021-06-08
Add pager option for customizing pager without environment variables

pbzweihander:
It would be great if we can configure `delta` with git config (or CLI args) only, without using environment variables.


dandavison:
Hi @pbzweihander, thanks very much for this! I agree with you that this should exist; I believe I've wanted to add it several times but never quite got around to it.

src/main.rs
```diff
@@ -82,7 +82,8 @@ fn run_app() -> std::io::Result<i32> {
         return Ok(0);
     }
 
-    let mut output_type = OutputType::from_mode(config.paging_mode, None, &config).unwrap();
+    let mut output_type =
+        OutputType::from_mode(config.paging_mode, config.pager.clone(), &config).unwrap();
```
dandavison, line 86: Rust question (I've already merged the PR so this is not blocking, I'm just curious): so here we're `clone`ing an `Option<String>`. Do you have a view on whether that is preferable to passing `Option<&str>` and creating a `String` later as the code did previously? For example I think https://github.com/dandavison/delta/commit/bf277caa65bfee3dc51fdcfd3866c5c6afb3a560 is a commit which does that.

pbzweihander: I thought that `to_string` (in https://github.com/dandavison/delta/blob/pbzweihander-configable-pager-str/src/bat_utils/output.rs#L74) will always clone the string. So I thought if `from_mode` receives an (already) owned string, the caller can decide whether clone or give ownership (and not clone).

dandavison: I see, thanks! Yes, maybe, if we are really trying to optimize the code, there's a way to use `&str` internally in `from_mode` without converting to an owned `String`. But that's of course all stuff that existed before your changes. Thanks again!

Issue #603, opened 2021-05-16
🚀 Add git hash hyperlinks on custom log formats


ricobl:
It would be nice to have git hash hyperlinks when using alternative log formats (eg `git log --format=...` or `git log --oneline`).

PR #633, opened 2021-06-18
Use more permissive commit regex to support custom log formats

ricobl:
This is an attempt to fix [🚀 Add git hash hyperlinks on custom log formats](https://github.com/dandavison/delta/issues/603).

I use a custom log format with abbreviated commit hashes (`%h`):

```ini
[alias]
	lg = log --format='%C(green)%h %C(red)%cd %C(reset)%s %C(blue)%an/%C(cyan)%cn%C(yellow)%d%C(reset)' --date=relative
```

I believe the current regular expression doesn't work for a couple of reasons.

The first capture group requires a space before the hash and my custom format has the commit hash right after a color escape sequence.

On the second group, the commit hash length is fixed at 40 chars while abbreviated hashes can be as short as 7 chars, so I'm proposing making it a bit more flexible.

### DIFF
--- a/src/features/hyperlinks.rs
+++ b/src/features/hyperlinks.rs
@@ -73,7 +73,15 @@ fn format_osc8_hyperlink(url: &str, text: &str) -> String {
 }
 
 lazy_static! {
-    static ref COMMIT_LINE_REGEX: Regex = Regex::new(r"(.* ?)(\b[0-9a-f]{7,40}\b)(.*)").unwrap();
+    // The first group can be empty and needs to end with either a word-boundary or an ANSI escape code.
+    // This ensures all commit characters are consumed (from 7 to 40 chars).
+    static ref COMMIT_LINE_REGEX: Regex = Regex::new(format!(
+        "(.*(?:\\b|{osc}{style}))({commit})({suffix})",
+        osc = "\x1b\\[",
+        style = "[0-9;]*[mK]",
+        commit = "[0-9a-f]{7,40}",
+        suffix = ".*",
+    ).as_str()).unwrap();
 }
 
 fn format_commit_line_captures_with_osc8_commit_hyperlink(
--- a/src/features/hyperlinks.rs
+++ b/src/features/hyperlinks.rs
@@ -73,7 +73,7 @@ fn format_osc8_hyperlink(url: &str, text: &str) -> String {
 }
 
 lazy_static! {
-    static ref COMMIT_LINE_REGEX: Regex = Regex::new("(.* )([0-9a-f]{40})(.*)").unwrap();
+    static ref COMMIT_LINE_REGEX: Regex = Regex::new(r"(.* ?)(\b[0-9a-f]{7,40}\b)(.*)").unwrap();
 }
 
 fn format_commit_line_captures_with_osc8_commit_hyperlink(
--- a/src/tests/test_example_diffs.rs
+++ b/src/tests/test_example_diffs.rs
@@ -1611,30 +1611,32 @@ src/align.rs:71: impl<'a> Alignment<'a> { │
     #[test]
     fn test_hyperlinks_commit_link_oneline_log() {
         let config = integration_test_utils::make_config_from_args(&[
-            // Use commit-style to ensure test to pass in Github Actions
+            // Use commit-style and commit-regex to ensure test to pass in Github Actions
             "--commit-style",
             "blue",
+            "--commit-regex",
+            "[0-9a-f]{7,40}",
             "--hyperlinks",
             "--hyperlinks-commit-link-format",
             "https://invent.kde.org/utilities/konsole/-/commit/{commit}",
         ]);
         let output = integration_test_utils::run_delta(GIT_LOG_ONELINE, &config);
-        println!("{}", output);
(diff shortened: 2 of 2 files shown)


ricobl:
I've fixed the failing test and included a couple more tests to support one-line logs.

ricobl:
Actually one-line log tests are failing in Github: https://github.com/ricobl/delta/runs/2884116372

dandavison:
Hi @ricobl, thanks for this! Regarding the test failure, I don't know exactly what's wrong yet. As you've probably seen, you can recreate the failure locally if you redirect standard output to a file, e.g.

```
cargo test tests::test_example_diffs::tests::test_hyperlinks_commit_link_oneline_log > /tmp/out
```

```
thread 'tests::test_example_diffs::tests::test_hyperlinks_commit_link_oneline_log' panicked at 'assertion failed: output.contains(r\"https://invent.kde.org/utilities/konsole/-/commit/fc5f383\")', src/tests/test_example_di>
```

It must be a similar problem to the one described in the code comment: https://github.com/dandavison/delta/blob/a75ca465060793148e81b3d5cc8d33207653ecbc/src/tests/test_example_diffs.rs#L1597-L1600

src/features/hyperlinks.rs
```diff
@@ -73,7 +73,15 @@ fn format_osc8_hyperlink(url: &str, text: &str) -> String {
...
-    static ref COMMIT_LINE_REGEX: Regex = Regex::new("(.* )([0-9a-f]{40})(.*)").unwrap();
+    // The first group can be empty and needs to end with either a word-boundary or an ANSI escape code.
+    // This ensures all commit characters are consumed (from 7 to 40 chars).
+    static ref COMMIT_LINE_REGEX: Regex = Regex::new(format!(
+        "(.*(?:\\b|{osc}{style}))({commit})({suffix})",
+        osc = "\x1b\\[",
+        style = "[0-9;]*[mK]",
+        commit = "[0-9a-f]{7,40}",
+        suffix = ".*",
+    ).as_str()).unwrap();
```
dandavison, line 84: What do you think about keeping this regex simple and applying the function `ansi::strip_ansi_codes` to the input before passing it to the regex? The possible advantage of that approach is that it keeps all the ANSI-parsing logic in one place. That function should be quite fast as the underlying parsing logic comes from the vte terminal emulator project.

src/tests/test_example_diffs.rs
```diff
@@ -1608,6 +1608,36 @@ src/align.rs:71: impl<'a> Alignment<'a> { │
...
+        let config = integration_test_utils::make_config_from_args(&[
+            // Use commit-style to ensure test to pass in Github Actions
+            "--commit-style",
+            "blue",
+            "--hyperlinks",
+            "--hyperlinks-commit-link-format",
+            "https://invent.kde.org/utilities/konsole/-/commit/{commit}",
+        ]);
+        let output = integration_test_utils::run_delta(GIT_LOG_ONELINE, &config);
+        println!("{}", output);
```
dandavison, line 1622: (These print statements can be deleted once the tests are passing with standard output redirected)

ricobl: Oh yeah, good pickup! I'll remove them once the Github build is sorted out.

src/features/hyperlinks.rs
ricobl: Keeping the regex simple sounds interesting, wouldn't it remove colors from a custom `git log --format` though?

ricobl (commented):
Thanks for the insight @dandavison! I noticed that comment on `test_hyperlinks_commit_link_format` and used the same approach. I'll try to figure it out.

I'm not sure what's happening with clippy though, I haven't touch those files.

src/features/hyperlinks.rs
dandavison: Ah right, I forgot the matched parts of the regex were being relied on for the colored output. Ignore me then :)

ricobl:
For some reason the `git log --oneline` lines were not getting picked up by the commit regex below when running in Github or directing output to a file.

https://github.com/dandavison/delta/blob/d4da66d0fc94ab2e62abd3cbb0d4221b422f7753/src/delta.rs#L118-L119

I verified that tests are not going through the commit regex match when running locally, which is why they were passing initially. Maybe it's related to the raw mode mentioned earlier.

I managed to fix the new tests by passing `[0-9a-f]{7,40}` as a the `commit-regex` and after that I tried to make that regex the new default and a lot of stuff failed. 

It's challenging to find commit hashes with custom log formats without also affecting other sections like filenames and diffs:

* hashes can be anywhere with a length of 7 to 40 chars
* it's very easy to mistake other sequences with a commit hash (e.g. a file named `aabbccdd.txt`)
  * it might be possible to run a command in parallel to check for real commit hashes, but this is probably sub-optimal
* hashes can be preceded by an ANSI code from a custom log format, which makes the commit regex complex and hard to maintain
  * unless the approach of stripping ANSI codes is taken as mentioned earlier

I'm starting to think that this not a viable approach and maybe it's easier to stick to the `commit <40_chars_hash>` pattern when using custom logs but offering a way to only truncate the link text to X chars.

Any thoughts @dandavison?

PR #642, opened 2021-06-24
Fix empty line highlighting

th1000s:


dandavison:
From offline discussion: consider the following diff

```
diff·--git·a/a·b/b␊
index·8cb7cb9..e69de29·100644␊
---·a/a␊
+++·b/b␊
@@·-1,2·+0,0·@@␊
-·␊
-␊
```

Prior to this PR, under `side-by-side`, the line that contained no whitespace is highlighted as a single removed block:

image


With this PR, it is styled the same as the line that did have some whitespace:

image

Issue #291, opened 2020-08-14
🚀 Display git blame commit hashes as hyperlinks 


dandavison:
With `hyperlinks = true`, commit hashes shown by `git blame` should be rendered as hyperlinks. By default, these should be http links to the commit page in the remote web-hosted git repo.

A new option `hyperlinks-commit-link-format` could be added to allow the nature of commit links to be customized. For example, one could use a custom link protocol and register a handler on your system to do something like

1. Display the commit using git and delta in a new terminal window

2. Display the commit in a different git client, such as magit.

 We already have `hyperlinks-file-link-format` which allows users to use custom protocols to open delta file hyperlinks in their editor/IDE. See https://github.com/dandavison/open-in-editor for some usage notes.

cc @navarroaxel @Kr1ss-XD

phil-blain:
Hi @dandavison ! searching "blame" in the issues brought me to this one, but here is one (crazy?) idea:

delta could syntax-highlight `git blame output` ! that would be really great. What do you think (I can open a new issue for that if you like) ?

dandavison:
> delta could syntax-highlight git blame output ! that would be really great. What do you think (I can open a new issue for that if you like) ?

I agree! Please do open an issue. I also think delta should highlight merge conflicts in a way that is helpful. So, there are definitely some major features that should  be added to delta but haven't been yet. Of course, they are open to anyone out there who'd like to have a go. (And having open issues for them is a very good start.)

PR #650, opened 2021-07-03
Do not generate hyperlinks for hunks with no file or line number

maximbaz:
Consider the following `delta` config:

```ini
[delta]
	; features = hyperlinks
	hunk-header-style = syntax bold red
	hunk-header-decoration-style = omit
```

`git diff` would look like this:

image

Now uncomment `hyperlinks` feature, git diff would now look like this:

image

Notice the wrong appearance of `:` in front of the hunk.

With this PR, when hunks contain no file or line number info, the url will not be generated (since it's not clear where to click anyway), and so the `:` symbol would not be added later on.

Let me know if this makes sense?

### DIFF
--- a/src/hunk_header.rs
+++ b/src/hunk_header.rs
@@ -117,10 +117,7 @@ fn get_painted_file_with_line_number(
         )
     }
     let file_with_line_number = ansi_term::ANSIStrings(&file_with_line_number).to_string();
-    if config.hyperlinks
-        && (config.hunk_header_style_include_file_path
-            || config.hunk_header_style_include_line_number)
-    {
+    if config.hyperlinks && !file_with_line_number.is_empty() {
         features::hyperlinks::format_osc8_file_hyperlink(
             plus_file,
             Some(plus_line_number),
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks @maximbaz! Yes, I agree this makes sense. Do you have time to add a unit test for this bug?

So, for the record, I think what we're saying is that `--hyperlinks` means something like

> Output hyperlinks wherever they make sense given other settings

Examples of things that might be formatted as hyperlinks include the commit hash, the file/line number in the hunk header, and git blame commit hashes when someone implements that! (#291).

src/hunk_header.rs
```diff
@@ -117,7 +117,10 @@ fn get_painted_file_with_line_number(
         )
     }
     let file_with_line_number = ansi_term::ANSIStrings(&file_with_line_number).to_string();
-    if config.hyperlinks {
+    if config.hyperlinks
+        && (config.hunk_header_style_include_file_path
+            || config.hunk_header_style_include_line_number)
+    {
```
dandavison, line 123: I wonder whether it would be slightly cleaner to test `file_with_line_number.is_empty()` here rather than re-using the two config flags, which have already been consulted above.

maximbaz:
Thanks for the review! Good suggestions. I added some tests, bottom two are most important (the last one would fail without the patch), top two just felt good to have, although they did produce some at-first-unexpected results, however looking at the code I understand now why these are the results 🙂 If you have other suggestions, please let me know!

dandavison:
Great, thanks very much for this work @maximbaz.

PR #654, opened 2021-07-08
Update README using Arch Linux official packages

navarroaxel:
Update the README because the `git-delta` package was moved from the Arch User Repository to the official community repository.


dandavison:
Great, that's good news! Thanks @navarroaxel.

Issue #557, opened 2021-04-07
🐛 git add -p added lines are shown empty


domrim:
since the last update `git add -p` does not show the content of new lines. see screenshot.

![image](https://user-images.githubusercontent.com/15873147/113883337-59bbea80-97be-11eb-8fbd-eae2a7d6d147.png)

my relevant `.gitconfig` parts, all pulled from the Readme:
```
[core]
    pager = delta

[interactive]
    diffFilter = delta --color-only

[delta]
    features = side-by-side line-numbers decorations
    whitespace-error-style = 22 reverse

[delta "decorations"]
    commit-decoration-style = bold yellow box ul
    file-style = bold yellow ul
     file-decoration-style = none
```

running delta v0.7.1 on macOS 11.2.3

domrim:
funny: i just checked it with another repo. there it is fine.

could it depend on the file content?

domrim:
yes, i've created a small minimal project which triggers the bug for me. you can find it here: https://github.com/domrim/example

dandavison:
Thanks! I'll take a look.

CaioCSdev:
I'm having the exact same issue. git diff works, but git add -p fails :thinking: 

![image](https://user-images.githubusercontent.com/1172532/118757121-c9ec7000-b842-11eb-886b-f443b65f6271.png)

CaioCSdev:
I discovered the issue!! :tada: 

Is the line break char.
Delta works perfectly with `LF`, but when the file uses `CRFL` this strange error happens, and only on `git add -p` (at least so far)

got a hint because I was using cshap

dandavison:
Hi @CaioCSdev, awesome, thanks! Now that you point this out I realise that the test suite has no coverage for alternate line endings. This sounds very fixable now.

michaelblyons:
Is this done with #664 or waiting for #754?

dandavison:
Thanks @michaelblyons. This should be fixed already by #664 (please someone shout out if that's not true). #754 raises a related question about that fix that we don't yet have an answer to, but the #664 fix still stands currently.

Issue #328, opened 2020-09-21
🐛 diff render issue on interactive patch `git add -p` with Windows Terminal


RemySphere:
Hi, 
Using the sample configuration, I have an unreadable output under Windows Terminal 1.2.2381.0
Only the background color seems to be rendered, i.e copy/pasting shows no content?

To make things more complicated the terminal I'm spawning calls `init.bat` from [cmder](https://github.com/cmderdev/cmder), so it might as well be a terminal emulator issue... 

- I'm using the patched version of less.exe, as recommended (although it doesn't make a difference for this particular issue)
- `git add -p` runs fine in a WSL instance within the same Windows Terminal window

~/.gitconfig
```
[interactive]
    diffFilter = delta --color-only

[delta]
    features = side-by-side line-numbers decorations
    whitespace-error-style = 22 reverse

[delta "decorations"]
    commit-decoration-style = bold yellow box ul
    file-style = bold yellow ul
    file-decoration-style = none
```

outputs
![image](https://user-images.githubusercontent.com/18334704/93766619-f2ba5800-fc16-11ea-9ea1-8b07ce13240e.png)

```
λ delta --version
delta 0.4.3
λ git --version
git version 2.28.0.windows.1
```

dandavison:
Hi @RemySphere, is delta working for you outside of the `git add -p` scenario? E.g. `git diff`, `git show`, etc.

If you pipe output from git to delta directly, does that look correct? `git diff | delta`

RemySphere:
Before I only had `core.pager = delta --dark` set in my `.gitconfig`, and there were no apparent issue with my day-to-day use, except the one.

But indeed, now that I am fiddling with more features there are a few glitches here and there...
![image](https://user-images.githubusercontent.com/18334704/93785615-93684200-fc2e-11ea-9eba-c5dd341448a1.png)

`side-by-side` is mostly ok, with occasional encoding errors
![image](https://user-images.githubusercontent.com/18334704/93785895-e8a45380-fc2e-11ea-8f5a-10705db9c865.png)

There is no visible difference when piping output from git to delta directly.  
Side note, those encoding artifacts don't seem to occur in a WSL container, but the main "background only" issue still does, somehow

voqv:
[...]

PR #664, opened 2021-07-22
Fix the empty result issue in add -p

norisio:
Related Issues: https://github.com/dandavison/delta/issues/328, https://github.com/dandavison/delta/issues/557

### Details of the issue

Current `delta` can't handle CRLF-ending diff hunks at least when used through `git add -p`. The reason here:

* `bytelines` library (used [here](https://github.com/dandavison/delta/blob/e5664693a955a81143a00c686e7c374f62692bc5/src/main.rs#L99)) strips CRs (`b'\r'`, 0x0D) as well as LFs (`b'\n'`, 0x0A) when it splits bytes into lines.
    * Find the operations [here](https://github.com/whitfin/bytelines/blob/6eccd7fb3c86e538397e39b60599b6c704e90ce3/src/lib.rs#L126-L132).
* So there shouldn't be any CRs in [`lines: ByteLines<BufRead>`](https://github.com/dandavison/delta/blob/e5664693a955a81143a00c686e7c374f62692bc5/src/delta.rs#L106).
* However, `git` inserts some ansi control characters between CR and LF: 
image
* So _bytelines_ library can't find and leaves the CR in `lines`.
* CR will be emitted directly into a terminal (at least when _delta_ used through `git add -p`). Thus the rendered result will be blank (CR erases the line).

### What I changed

When [`ingest_line`](https://github.com/dandavison/delta/blob/e5664693a955a81143a00c686e7c374f62692bc5/src/delta.rs#L175), delta strips the trailing CR.

Before:

image

After:

image

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -188,10 +188,8 @@ impl<'a> StateMachine<'a> {
         // Strip the neglected CR.
         // (CR-LF is unfortunately split by git because it adds ansi escapes between them.
         //  Thus byte_lines library can't remove the CR properly.)
-        if let Some(c) = self.line.bytes().nth_back(0) {
-            if c == b'\r' {
-                self.line.truncate(self.line.len() - 1);
-            }
+        if let Some(b'\r') = self.line.bytes().nth_back(0) {
+            self.line.truncate(self.line.len() - 1);
         }
     }
 
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -184,6 +184,15 @@ impl<'a> StateMachine<'a> {
             .to_string()
         };
         self.line = ansi::strip_ansi_codes(&self.raw_line);
+
+        // Strip the neglected CR.
+        // (CR-LF is unfortunately split by git because it adds ansi escapes between them.
+        //  Thus byte_lines library can't remove the CR properly.)
+        if let Some(c) = self.line.bytes().nth_back(0) {
+            if c == b'\r' {
+                self.line.truncate(self.line.len() - 1);
+            }
+        }
     }
 
     /// Should a handle_* function be called on this element?
--- a/src/tests/test_example_diffs.rs
+++ b/src/tests/test_example_diffs.rs
@@ -297,6 +297,16 @@ commit 94907c0f136f46dc46ffae2dc92dca9af7eb7c2e
     }
 
     #[test]
+    fn test_orphan_carriage_return_is_stripped() {
+        let config = integration_test_utils::make_config_from_args(&[]);
+        let output = integration_test_utils::run_delta(
+            GIT_DIFF_SINGLE_HUNK_WITH_SEQUENCE_OF_CR_ESCAPE_SEQUENCES_LF,
+            &config,
+        );
+        assert!(output.bytes().all(|b: u8| b != b'\r'));
+    }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


dandavison:
This (with the explanation) looks great @norisio, thank you! I've set the tests running on Github actions. Would you be able to add a test for the bug you've fixed? There are lots of examples in the codebase to follow, for example one possible style of test is the end-to-end type tests in [test_example_diffs.rs](https://github.com/dandavison/delta/blob/master/src/tests/test_example_diffs.rs), but if you prefer more of a unit test style that would be good too.

norisio:
OK, will try.

norisio:
I added a testcase. `GIT_DIFF_SINGLE_HUNK_WITH_SEQUENCE_OF_CR_ESCAPE_SEQUENCES_LF` contains the "harmful" pattern CR-_ansi_-LF.

@dandavison Could you check the content?

On master + testcase:

```shell-session
$ cargo test  orphan
    Finished test [unoptimized + debuginfo] target(s) in 0.05s
     Running target/debug/deps/delta-170397dd0efbb98e

running 1 test
test tests::test_example_diffs::tests::test_orphan_carriage_return_is_stripped ... FAILED

failures:
(snip)
```

On this branch:

```shell-session
$ cargo test  orphan
    Finished test [unoptimized + debuginfo] target(s) in 4.75s
     Running target/debug/deps/delta-170397dd0efbb98e

running 1 test
test tests::test_example_diffs::tests::test_orphan_carriage_return_is_stripped ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 233 filtered out; finished in 1.86s
```

src/delta.rs
```diff
@@ -184,6 +184,15 @@ impl<'a> StateMachine<'a> {
...
         self.line = ansi::strip_ansi_codes(&self.raw_line);
+
+        // Strip the neglected CR.
+        // (CR-LF is unfortunately split by git because it adds ansi escapes between them.
+        //  Thus byte_lines library can't remove the CR properly.)
+        if let Some(c) = self.line.bytes().nth_back(0) {
+            if c == b'\r' {
+                self.line.truncate(self.line.len() - 1);
+            }
+        }
```
dandavison, line 195: Minor tweak -- how about this?

```rust
        if let Some(b'\r') = self.line.bytes().nth_back(0) {
            self.line.truncate(self.line.len() - 1);
        }
```

src/tests/test_example_diffs.rs
```diff
@@ -1708,6 +1718,20 @@ Date:   Thu May 14 11:13:17 2020 -0400
...
+[1m+++ b/src/main.rs[m
+[36m@@ -1,5 +1,4 @@[m
+[31m-deleted line[m\r
+[31m-[m\r
+ fn main() {[m\r
+     println!(\"existing line\");[m\r
+[32m+[m[32m    println!(\"added line\");[m[41m\r[m
+ }[m\r
+";
+
```
dandavison, line 1734: Awesome, thank you!

src/delta.rs
norisio: I've never done like this. it's better! Committed.

dandavison:
Thanks very much for this work @norisio.

PR #668, opened 2021-07-26
Recognize GitHub SSH remote URLs that don't start with `git@` for hyperlinks

### LINKED ISSUE #667: 🚀 Show Github commit hyperlinks even if remote URL does not contain `git@`
Hi! I was recently a bit confused why hyperlinks for commits were working in some of my local clones but not others.
After a bit of investigation I've figured out that it's caused by the remote URL being simply `github.com:user/repo` instead of `git@github.com:user/repo`, which is not matched by the current regex.

Now, this remote URL only works in the first place because of a custom SSH config containing something like this:
```
Host github.com
    HostName github.com
    User git
    IdentityFile <some key>
```

Considering that I fully understand if you'd prefer to not support such configs out-of-the-box. (You can also put anything into the `Host` portion there and have it still end up going to GitHub with configs like this and I imagine that parsing the SSH config or anything similar to make everything here work is definitely out of scope 😄 ).

Still, I figured specifically allowing omitting the `git@` is a pretty small change so I might as well ask (and I suspect this is at least a *somewhat* common config; I don't remember where I stole it, but I definitely got the idea from someone else).

spaarmann:
Closes #667.

SSH remote URLs are usually formatted as `git@github.com:user/repo` but certain setups can allow using just `github.com:user/repo`, see issue #667 for an example.

This modifies the regex used to detect a GitHub remote for hyperlink generation to also allow the latter form.

### DIFF
--- a/src/git_config/git_config_entry.rs
+++ b/src/git_config/git_config_entry.rs
@@ -23,7 +23,7 @@ lazy_static! {
     static ref GITHUB_REMOTE_URL: Regex = Regex::new(
         r"(?x)
         ^
-        (?:https://|git@) # Support both HTTPS and SSH URLs
+        (?:https://|git@)? # Support both HTTPS and SSH URLs, SSH URLs optionally omitting the git@
         github\.com
         [:/]              # This separator differs between SSH and HTTPS URLs
         ([^/]+)           # Capture the user/org name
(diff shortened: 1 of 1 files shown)


dandavison (approved):
LGTM! Thanks for dealing with the unit test also.

PR #676, opened 2021-08-05
Use git-config-env crate instead of hacky GIT_CONFIG_PARAMETERS parsing

dandavison:
Thanks @epage. Ref #573


dandavison:
Build is currently failing to compile `bitvec` when running `cargo-tarpaulin`. Looks like it may be related to the `nom` dependency: https://github.com/xd009642/tarpaulin/issues/756

```
   Compiling bitvec v0.19.5
error: Broken pipe (os error 32)
warning: build failed, waiting for other jobs to finish...
error: Broken pipe (os error 32)
error: build failed
Aug 05 21:45:59.811 ERROR cargo_tarpaulin: Failed to compile tests! Error: bitvec: cannot find attribute `skip` in this scope
Error: "Failed to compile tests! Error: bitvec: cannot find attribute `skip` in this scope"
Error: The process '/usr/share/rust/.cargo/bin/cargo' failed with exit code 1
```

dandavison:
> I suspect this can be written as:
> 
> ```rust
> .map(|(k, v)| (k.into_owned(), v.map(Cow::into_owned).unwrap_or_default()))
> ```

Thanks! I believe you're right. 7d4c3c8b26d39a30fec34058492b29aa3eccce46. I guess if this were 100% clear clippy would have an opinion? (I'm not expert on this; I did just read https://users.rust-lang.org/t/what-is-the-idiomatic-way-to-convert-str-to-string/12160 and https://users.rust-lang.org/t/to-string-vs-to-owned-for-string-literals/1441/5)

epage:
Oh interesting, `String::to_string` uses specialization to avoid `Display` which isn't available outside the stdlib
https://doc.rust-lang.org/nightly/src/alloc/string.rs.html#2457

I've found myself settling on the same approach as dtolnay in the second thread and focusing on the conversion that most semantically describes what I'm trying to do.

PR #690, opened 2021-08-15
Fix computed values to be computed after all set_options

ulwlu:
## What this pr does

- Make "true-color = ***" from gitconfig works

## How it was

It was not working from original source.

ss 0003-08-15 at 21 38 49

So I make it works.

ss 0003-08-15 at 21 43 56

This still give prior to the config from cli. Also 24-bit-color can still work.

ss 0003-08-15 at 21 44 11

## Motivaiton

I'm trying to integrate delta inside tig.

https://github.com/dandavison/delta/discussions/689

However, if you allow user to use RGB, it's almost impossible inside tig since it uses ncurses. So I'm trying to limit only 8bit colors. Then I found this bug.

### DIFF
--- a/src/options/set.rs
+++ b/src/options/set.rs
@@ -90,12 +90,8 @@ pub fn set_options(
     let features = gather_features(opt, &builtin_features, git_config);
     opt.features = features.join(" ");
 
-    set_widths(opt, git_config, arg_matches, &option_names);
-
     // Set light, dark, and syntax-theme.
-    set_true_color(opt, git_config, arg_matches, &option_names);
     set__light__dark__syntax_theme__options(opt, git_config, arg_matches, &option_names);
-    theme::set__is_light_mode__syntax_theme__syntax_set(opt, assets);
 
     // HACK: make minus-line styles have syntax-highlighting iff side-by-side.
     if features.contains(&"side-by-side".to_string()) {
@@ -192,6 +188,10 @@ pub fn set_options(
         true
     );
 
+    // Setting ComputedValues
+    set_widths(opt);
+    set_true_color(opt);
+    theme::set__is_light_mode__syntax_theme__syntax_set(opt, assets);
     opt.computed.inspect_raw_lines =
         cli::InspectRawLines::from_str(&opt.inspect_raw_lines).unwrap();
     opt.computed.paging_mode = parse_paging_mode(&opt.paging_mode);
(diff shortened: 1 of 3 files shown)


dandavison:
Hi again @ulwlu! Great, thanks for this. I've pushed one commit to your branch adding a test that is intended to reproduce  the bug that you are fixing. But there's one thing I'm confused about:

My test fails on master and passes on your branch. However my test tests `true-color = never` whereas the new code you've added is inside a `if opt.true_color == "auto"` statement. I haven't had time yet to figure out why my test passes on your branch. Would you be able to look at the test, and check everything makes sense, and add any further test cases you think are needed?

ulwlu:
Hi @dandavison thank you for your review.

This is bacause cli.rs set opt.true-color to 'auto' if '--true-color' option is not given. That is the default value given by struct-opt.

I think it's little bit confusing, maybe we can remove the default value and change match condition inside set_true_color function.

src/options/set.rs
```diff
@@ -556,14 +556,31 @@ fn set_widths(
...
+            let empty_builtin_features = HashMap::new();
+            set_options!(
+                [true_color],
+                opt,
+                &empty_builtin_features,
+                git_config,
+                arg_matches,
+                option_names,
+                false
+            );
```
dandavison, line 580: `true_color` is present in the main call to `set_options!` on [l.182](https://github.com/dandavison/delta/blob/93e081985fe7192f7cb6f472ae8d6fb835cdc0e4/src/options/set.rs#L181-L182). I think we need to know why that call is not working -- it feels wrong to make two `set_options!` calls for `true_color`.

dandavison:
> This is bacause cli.rs set opt.true-color to 'auto' if '--true-color' option is not given. That is the default value given by struct-opt.

Thanks for explaining that!

src/options/set.rs
ulwlu: Oh I see, and sorry. I missed that and found `opt.true_color` is set correctly from gitconfig after set_options. I'm maybe going to fix.

ulwlu:
Thanks to your pointing out, I found some configurations which is to be computed depends on `opt.***` value must be after set_options which gets value from gitconfig. It is because gitconfig value will never be computed.

I'm going to think what way is good.

ulwlu:
@dandavison 

Hi I fixed and change this PR's purpose.

- Change ComputedValues's computing place after all set_options
  - It is because some of them depends on `opt.**` value but they have chance to miss gitconfig values
- The ComputedValues which have a default value and computed before set_options were going wrong.
  - The ComputedValues which have a default value: `paging-mode`, `inspect-raw-lines`, and `true-color`
  - The ComputedValues which computed before set_options: `true-color`
  - So, only `true-color` was going wrong

I move `set_width` after set_options as well, so this func now doesn't need to set_options inside it.

- I added a test for these computed values are setting correctly into config.
- I removed an unused ComputedValue.

dandavison:
Awesome, thanks very much for these fixes @ulwlu. I did some manual testing to check that `width` worked when set as a command-line option, and in gitconfig, and via `git -c delta.width=xxx ...`.

ulwlu:
@dandavison thanks!

Btw, integrating with delta is almost finished. Now I'm fixing little bugs.

https://github.com/jonas/tig/pull/1140

dandavison:
> Btw, integrating with delta is almost finished. Now I'm fixing little bugs.

Awesome. Those screenshots look really nice. (I just tried compiling but got an error

```
src/line.c:245:3: error: implicit declaration of function 'init_extended_pair' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
                init_extended_pair(++cnt, fg, bg);
```

)

ulwlu:
Umm, init_expanded_pair is only available with ncurse upper than version 6.1, so compiling option may get wrong (such as linking with different path).

I also strrugled at that building. If this get merged jonas will release the new binary anyway, so maybe trying after that is easier...

dandavison:
OK thanks (my Mac has 5.4, but I saw your instructions for building and linking against ncurses).

PR #692, opened 2021-08-18
Remove unnecessary borrows

benarmstead:
These borrows are not necessary and de-referenced by the compiler.


dandavison:
Hey @benarmstead, thanks very much for doing this. A few days ago I disabled this clippy check in github actions as a temporary measure, so I've pushed a commit to your branch reverting that. I'll merge when the build is green!

benarmstead:
No problem, great to hear!

PR #694, opened 2021-08-20
Support `insteadOf` replacements in git remote URLs

### LINKED ISSUE #693: 🚀 Remote URL detection should parse `insteadOf` replacements
Git supports replacing parts of URLs with other strings. For example, I can give a remote URL of `github:user/project` and it will go off to `github.com` and clone it. If I try to push, it will instead use the `ssh` URL (this avoids having to set up different URLs for push and pull manually). I see that `delta` is parsing out the remote based on the `remote.$name.url` setting. However, it should really use `git remote get-url` to have `insteadOf` replaced as intended.

dandavison:
Use libgit2 to query remote URL instead of querying remote.origin.url
directly. This has the consequence that `insteadOf` replacements are
honored.

See https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf

Fixes #693


mathstuf (approved):
Looks like this should do it 👍

PR #718, opened 2021-09-15
Remove superfluous lowercasing of feature names

### LINKED ISSUE #716: 🐛 Themes whose names have uppercase letters are silently ignored
If I create a theme feature in `.gitconfig` (or an included file) named like `[delta "Tak"]`, it will be silently ignored (also in e.g. `--show-themes`)

This appears to be a result of force-lowercasing in some places for comparison and doing original-case lookups in other places.

Obviously there's an easy workaround, but it took quite a while to find out why my theme wasn't getting picked up. :slightly_smiling_face:

Tak:
Fixes #716

I'm a little apprehensive that the lowercasing was there for a reason, but all the tests are passing :shrug:

### DIFF
--- a/src/options/get.rs
+++ b/src/options/get.rs
@@ -79,7 +79,7 @@ pub trait GetOptionValue {
                 return Some(value);
             }
         }
-        for feature in opt.features.to_lowercase().split_whitespace().rev() {
+        for feature in opt.features.split_whitespace().rev() {
             match Self::get_provenanced_value_for_feature(
                 option_name,
                 feature,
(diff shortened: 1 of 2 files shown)


dandavison:
Thanks very much @Tak!

> I'm a little apprehensive that the lowercasing was there for a reason

I've thought about it but I'm not sure why I did that. I think I may have had in mind code I'd written to do something like support "Red" and "red" as color names (which we do support), and then added that canonicalization inappropriately here. I don't see a reason why case-sensitivity of feature names will confuse people, and indeed it seems appropriate.

Issue #41, opened 2019-10-27
Terminal size is off by one on Linux


igrekster:
It appears, the detected terminal size is one character too wide:
![image](https://user-images.githubusercontent.com/1772013/67627856-20383d80-f8b0-11e9-8cff-0ce144327103.png)

I tried in Konsole-19.08.2, xterm-337, tmux-2.9a. I have not noticed any issues in other terminal apps such as `tig`.

I've tried on the latest release - 0.0.14.

igrekster:
I figured it out, it was my `~/.lesskeyrc` -- I have `-J` or `--status-column` enabled which reduces the width by one character. I wonder if it would be possible to detect it, or have a way to override.

dandavison:
Hm, this was supposed to have been fixed in v 0.0.13 by https://github.com/dandavison/delta/pull/22. Let's get to the bottom of this!

I'm not managing to reproduce yet in MacOS iTerm2 with delta 0.0.14. Here's a little experiment. Notice how, in the first case, delta is leaving one character of space unused at the right edge of the terminal. This is what is should always do by default, since #22. In the second case, `less` is invoked with `-J`, and so it inserts an extra space in the left column, but the lines do not overflow on the right because of the extra space which was left there for this reason.

Can you reproduce this? (If not, can you double check the version of `delta` that is being invoked by git (maybe pipe `git` output into a delta executable explicitly) and post the relevant contents of your `~.lesskey`).

**With `LESS` environment variable equal to empty string (`export LESS=`)**:
![image](https://user-images.githubusercontent.com/52205/67630420-6f488900-f844-11e9-87bf-40beb8b10344.png)

**With `export LESS=-J`**:
![image](https://user-images.githubusercontent.com/52205/67630417-6061d680-f844-11e9-8795-2c0f6d466ab9.png)

dandavison:
Also, if it's still mysterious, could you post the values of the environment variables `PAGER`, `BAT_PAGER` and `LESS`?

igrekster:
> Notice how, in the first case, delta is leaving one character of space unused at the right edge of the terminal. This is what is should always do by default, since #22. In the second case, less is invoked with -J, and so it inserts an extra space in the left column, but the lines do not overflow on the right because of the extra space which was left there for this reason.

Well, in my case `less -J` adds **two** spaces:

![image](https://user-images.githubusercontent.com/1772013/67632181-76c76b00-f8f4-11e9-9eae-3c3d0aa088a4.png)

[...]

Issue #10, opened 2019-09-28
Full-width bars are one character too long


nedbat:
Using iterm2 Build 3.3.3beta2 on Mac 10.13.6.  The bars that are meant to span the terminal are one character too long, and so wrap.  It happens with --compare-themes as well as when being used as a pager:
<img width="702" alt="image" src="https://user-images.githubusercontent.com/23789/65821802-85910400-e208-11e9-9e55-2ac85a8cdb30.png">
Is the blank at the start of the line supposed to be there?

dandavison:
Thanks for the report @nedbat. That leading whitespace is indeed not supposed to be there. I'd definitely like to get to the bottom of this. However, it might require a bit of back and forth, because I haven't managed to reproduce it yet.

The short version is that it's looking to me like your system, when it encounters the ANSI escape code at the beginning of the problematic lines, is causing some white space to be rendered (as if it were a space character), and also switching color as the ANSI code dictates. Whereas mine is just switching color.

I've tried iTerm2 3.3.4, 3.3.3, 3.3.2. I haven't found a link to the beta that you're on yet. Could you perhaps try 3.3.4?

Here's how the beginning of `--compare-themes` renders for me. So for me, the filename, the horizontal line and the ensuing diff hunk all appear to start in the first column<sup>1</sup>, whereas for you it looks like those lines start in the second column:

<img width=300px src="https://user-images.githubusercontent.com/52205/65897776-f7ec1a80-e3a7-11e9-91d9-d2013a9ff8c8.png" alt="image" />

The main clue here seems to be that the problematic lines are the ones that start with an ANSI escape code; the initial line ("Theme: 1337") lacks any ANSI codes and starts in the first column for both of us.

```
$ delta --compare-themes 2>/dev/null | head -n10 | cat -A
$
Theme: ^[[1m1337^[[0m$
$
$
^[[34mtests/data/hello.c^[[0m$
[...]

PR #727, opened 2021-10-02
Width can be an offset from the terminal width

th1000s:
No longer subtract 1 from the terminal width to accommodate e.g.
`less --status-columns`, instead set `--width -1` explicitly.

-----

Using line wrapping combined with ANSI codes to extend the terminal width requires exact knowledge of the painted area, so make reductions of it more explicit.

`--width 123-4` is also possible in to make calling with e.g. `$myCOLS-2` possible.


src/options/set.rs
```diff
@@ -515,18 +516,70 @@ fn parse_paging_mode(paging_mode_string: &str) -> PagingMode {
     }
 }
 
+fn parse_width_plusminus(width_arg: &str, terminal_width: usize) -> Result<usize, String> {
```
dandavison, line 519: The name "plusminus" is similar to the name "minusplus" that you're introducing in the main PR. Shall we use something different here, such as `parse_width_specifier()`?

src/options/set.rs
```diff
@@ -515,18 +516,70 @@ fn parse_paging_mode(paging_mode_string: &str) -> PagingMode {
...
+        }
+    };
+
+    Ok(width)
+}
+
 fn set_widths(opt: &mut cli::Opt) {
-    // Allow one character in case e.g. `less --status-column` is in effect. See #41 and #10.
-    opt.computed.available_terminal_width = (Term::stdout().size().1 - 1) as usize;
+    // If one extra one character for e.g. `less --status-column` is required use "-1"
```
dandavison, line 569: There's a redundant copy of the word "one" in this comment

th1000s:
Done and done! I hope the code to produce precise error messages for wrong arguments isn't too much :)

th1000s:
This PR would undo the implicit fixes for `less -J` / `--status-column` (see #10 and #41, 
@igrekster  @fdcds  @nedbat) and require those users to set `width = -1`. Do any distributions ship with these less options out-of-the-box? If so maybe autodetection is required.

src/cli.rs
```diff
@@ -531,9 +531,11 @@ pub struct Opt {
     #[structopt(long = "line-fill-method")]
     pub line_fill_method: Option<String>,
 
-    /// The width of underline/overline decorations. Use --width=variable to extend decorations and
-    /// background colors to the end of the text only. Otherwise background colors extend to the
-    /// full terminal width.
+    /// The width of underline/overline decorations. Can be absolute, absolute [+|-] offset, or
+    /// only an offset - in the last case the detected terminal width is changed
+    /// by the offset. Examples: "72", "80-2", "-2" or "98+2".
```
dandavison, line 536: Can you remind me what a common use case for `+` is?

th1000s: Honestly, none that I can think of - feature completeness? Making sure the ALU still knows how to add?

dandavison: Haha :) okay, I'm just thinking about how to make it as easy as possible to understand. What do you think about something like

```
 The width of underline/overline decorations. Examples: "72", "-2" (auto-detected terminal width minus 2). An expressions such as "74-2" is also valid (equivalent to 72 but may be useful in shell programming if a variable exists holding the value "74").
```

PR #752, opened 2021-10-27
Align expanded tab columns

### LINKED ISSUE #729: 🐛 Tabs are displayed incorrectly.
Tabs characters usually function in columns, meaning that if I write `Hi\tWorld` and `Hello\tWorld`, both occurances of `World` should be aligned. In Delta however this alignment doesn't work and instead tabs are always printed as the exact width that they're set to.

This is what my test file looks like in my editor (and how it should look):

This is what delta produces:

dhduvall:
Vary the number of spaces tabs expand into (with `--tabs` > 0) so that text after the tab starts at the next tabstop and tab columns align.

Fixes #729.

### DIFF
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -148,10 +148,7 @@ impl<'a> Painter<'a> {
         I: Iterator<Item = &'b str>,
     {
         if self.config.tab_width > 0 {
-            // XXX with_capacity()?  We don't know the length of the line, though the caller might.
-            // Obviously we wouldn't know the length of the result, but we could do what bat does
-            // and simply allocate twice the capacity of the original.  For now, just do it the
-            // slow way.
+            // We could use with_capacity(), but any performance increase is negligible.
             let mut buf = String::new();
             let mut cum_width: usize = 0;
 
@@ -854,8 +851,6 @@ mod tests {
 
     #[test]
     fn embedded_tab() {
-        // XXX Maybe someone thinks the tab-is-always-X-spaces behavior is correct, instead of
-        // tab-gets-you-to-the-next-tabstop?  Will that need to be configurable?
         let result = test_expand_tabs(&[], "12\tjunk");
         assert_eq!(result, "12  junk");
     }


th1000s:
* Because one byte or even one char inside a string does not equal a display width of one, that's why a Unicode library has to be asked to group them (i.e. `.graphemes()`, `strip_ansi_codes(..).width()` does something similar internally) so the tab stops are correct. Oh and a tab is always in its own group, so there is no need to call `find()` on the iterator element.

* Indeed. The "correct" place is after the syntax highlighter ran (e.g. `get_syntax_style_sections_for_lines()` in  `Painter::paint_buffered_minus_and_plus_lines()`). Then the the raw lines could either be fixed a) before the diff highlighting runs, or b) the diff output is fixed as well. A pure tabs-to-spaces replacement can take place where it does now (unless there is some language syntax where the highlighter produces different output, maybe Makefiles?). This function can also be reused if strategy a) is chosen.

dhduvall:
Why do we need to iterate over the graphemes? If a tab is guaranteed to be the one character in a grapheme, then we don't need to process the graphemes individually: whatever comes before or after the tab is going to be the same set of bytes/characters/graphemes after the tab replacement as it was before, no? So then we can just ask for the width of that `&str` in its entirety, and let the unicode library do the grapheme processing internally.

dandavison:
> If a tab is guaranteed to be the one character in a grapheme, then we don't need to process the graphemes individually: whatever comes before or after the tab is going to be the same set of bytes/characters/graphemes after the tab replacement as it was before, no?

I believe that's right -- as I understand things, by UTF-8 spec an ASCII byte such as `\t` does not occur in any UTF-8 multibyte characters, but I probably forgot that when I wrote the code, and thought that the only safe way to do it was to iterate over graphemes. @th1000s does that sound right?

th1000s:
Yes, when not taking tabstops into account the graphemes can be ignored.

But with this patch the exact display width up to a tab must be known. That could be done by finding the next tab position, and then letting some other function calculate the width from the previous tab position (or its replacement whitespace) up to this tab position. Internally, that function will do the same grapheme iteration, but we we are stuck with more bookkeeping and str generation - so why not start with graphemes and just add them up?

dandavison:
Just stropping by to say that this branch will be great to have. Let me know when you feel that it's wrapped up.

dhduvall:
Sorry; my life has been upside-down the past few weeks. I'll see if I can give it some attention this long weekend.

PR #758, opened 2021-11-01
Simplify lifetime annotation of Painter

th1000s:
The struct's lifetime is not used outside `new()`, and most
of the methods used more like free functions, so uniformly
use `'a` there. Also fix wrongly used lifetime for `get_syntax()`.


dandavison (approved):
LGTM, thanks for catching this. As you can probably see, the author was new to Rust. In addition to lifetime mistakes, I recall that I wanted to use more associated functions with `self` references, but was losing battles with the borrow checker, often due IIRC to the fact that `self.highlighter` needs to be a mutable reference.

PR #762, opened 2021-11-04
Remove +/- line prefix instead of substituting a space

th1000s:
Simplifies line handling and printing by removing a "magical"
1-offset previously required in various locations.

--------------

Of the failing tests `edits::tests::*` are probably straightforward (or just a chore like the deactivated `wrapping::tests::*`) but `tests::test_example_diffs::tests::test_whitespace_error` and the added `tests::test_example_diffs::tests::test_whitespace_change_highlighted` have off-by-one whitespace diff errors which should be looked at. My guess is that the highlighting should not be there in the first place and can just be ignored.

A few more comments probably need to be updated as well.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -6,7 +6,7 @@ use lazy_static::lazy_static;
 use crate::ansi;
 use crate::color;
 
-#[derive(Clone, Copy, Debug, PartialEq, Default)]
+#[derive(Clone, Copy, PartialEq, Default)]
 pub struct Style {
     pub ansi_term_style: ansi_term::Style,
     pub is_emph: bool,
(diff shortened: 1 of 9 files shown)


dandavison:
This is great! This was an annoying and slightly painful minor issue running through the codebase and it's much clearer with this change. I'll try to look at some of the failing tests.

dandavison:
I've pushed 9f56e26 which I believe fixes the whitespace error detection.

th1000s:
Thanks for the help! The edit operations make sense, and from my understanding it it something downstream of the alignment which expects this empty string in the first position.

dandavison:
> The edit operations make sense

Great, tests are passing (and I believe you have uncommented some previously commented tests). Is there more to do here? (PR is still in draft mode).

th1000s:
I think all the comments are up to date now as well, so this should be all.

src/style.rs
```diff
@@ -16,6 +16,42 @@ pub struct Style {
...
+            "Style {{ {}{}{}{}{}{} }}",
+            ansi,
+            is_set('e', self.is_emph),
+            is_set('o', self.is_omitted),
+            is_set('r', self.is_raw),
+            is_set('s', self.is_syntax_highlighted),
+            deco
+        )
+    }
+}
```
dandavison, line 53: This is fantastic by the way. I saw it in your PR, and am now benefitting.

PR #764, opened 2021-11-05
replace link to XVilka's gist with the updated one

Kr1ss-XD:
While researching true color support of GNU screen, I noticed that this link still points to XVilka's gist, which they have replaced by an actual repository.


dandavison (approved):
Thanks @Kr1ss-XD!

PR #767, opened 2021-11-07
Update private Homebrew formula to 0.9.2

### LINKED ISSUE #766: 🐛 Checksums of private Homebrew packages are broken
https://github.com/dandavison/delta/blob/9a6f7d3effc9c6a3eec6fb2c4b8d6bd6258b67e2/HomeBrewFormula/git-delta.rb#L8-L14

Both two packages have the same hash. I guess you fetched the archives with curl or something and URLs were wrong. curl does not fail on 404 resopnse until `-f` is given. So I guess, the `0019dfc4b32d63c1392aa264aed2253c1e0c2fb09216f8e2cc269bbfb8bb49b5` is a hash of 404 text.

rhysd:
Fix #766

### DIFF
--- a/HomeBrewFormula/git-delta.rb
+++ b/HomeBrewFormula/git-delta.rb
@@ -1,5 +1,5 @@
 class GitDelta < Formula
-  version "0.9.1"
+  version "0.9.2"
   desc "A viewer for git and diff output"
   homepage "https://github.com/dandavison/delta"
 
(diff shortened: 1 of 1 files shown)


dandavison (approved):
Thanks very much @rhysd!

By the way, do you use this private homebrew package? The only reason it's there is because it predates delta ("git-delta") being in public homebrew. But delta's been in public homebrew for over a year now, and so I'd like to discontinue this private formula if possible. Do you have any advice/experience regarding how to go about that?

rhysd:
> By the way, do you use this private homebrew formula?

Actually, no. I was just looking at sources of this repository and noticed the mistake.

PR #770, opened 2021-11-10
Util: check parent process for filename

th1000s:
Usecase: With delta used as a pager the command `git blame main.rs` calls
delta, but the piped data does not mention the filename. Instead try to
get the filename extension from the parent `git` process.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,0 +1,300 @@
+use std::collections::{HashMap, HashSet};
+use sysinfo::{Pid, Process, ProcessExt, SystemExt};
+
+#[derive(Debug, PartialEq)]
+pub enum GitBlameExtension {
+    Some(String),
+    None,
+    NotGitBlame,
+}
+
+pub fn git_blame_filename_extension() -> Option<String> {
+    let mut info = sysinfo::System::new();
+    let my_pid = std::process::id() as Pid;
+
+    // 1) Try the parent process. If delta is set as the pager in git, then git is the parent process.
+    let parent = parent_process(&mut info, my_pid)?;
+
+    match guess_git_blame_filename_extension(parent.cmd()) {
+        GitBlameExtension::Some(ext) => return Some(ext),
+        GitBlameExtension::None => return None,
+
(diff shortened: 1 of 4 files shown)


th1000s:
Nice, this even works on windows, thanks to `sysinfo`:

    windows-latest, x86_64-pc-windows-msvc:
        test utils::tests::test_process_parent_cmd_args ... ok

dandavison:
Excellent! (There seem to be some test failures on one of the linux builds)

src/utils.rs
```diff
@@ -0,0 +1,91 @@
...
+            saw_dash_dash = true;
+            continue;
+        }
+        if saw_dash_dash || !x.starts_with('-') {
+            return x.split('.').last().map(str::to_owned);
+        }
+    }
+
+    None
+}
```
dandavison, line 22: I wonder if there is a crate dedicated to parsing arbitrary command lines known only at run time?

th1000s: Since this is more a command line ignorer than parser it is a rather simple function, see `skip_uninteresting_args`.

th1000s:
Updated to also work in other cases, such as piped input (`git blame src/main.rs | delta`), but this is not guaranteed to work if the git process is too fast or the data too small.

The tests failed on ARM because there they are not executed on hardware but in emulation, so I try to detect that.

Also, `utils.rs`..., any other ideas? :)

dandavison:
> Also, utils.rs..., any other ideas? :)

How about `process_utils.rs`? (We already have `bat_utils/` and `syntect_utils/`; it might be nice to reorganize as `utils/*` at some point.)

 My instinct is to move all the blame-specific stuff from here into `handlers/blame.rs`, once they are brought together in https://github.com/dandavison/delta/pull/761.

I'm using the API provided in this PR in https://github.com/dandavison/delta/pull/774 to interrogate `git grep` arguments; very useful!

dandavison:
Merged -- it looks like the only test complaints were clippy complaining that things were not called, as indeed they will not be until we use this in #761, and the coverage test for which I'll try adding `tarpaulin` as an expected argument along with `test`.

Thanks very much for this -- it is already quite sophisticated.

PR #780, opened 2021-11-17
Use modify label on diffs & add right-arrow option

WayneD:
This branch was inspired by my using delta to filter unified diffs and noticing that it was using a hard-wired "comparing:" label instead of the normal file-modified-label.  This was aggravated by that heading not being in the default --navigate regex, so my n/N commands weren't stopping at it.  I also noticed that having delta do the diff generation resulted in no label at all, not even with --navigate, which could not be fixed by a regex tweak.  So, instead of adding the "comparing:" string to the navigate regex default, I decided to change the diff headings to use the modified label.  While twiddling things I also decided to make the right-arrow configurable because I don't like how the ⟶ has 2 extra spaces on the right side (which might be needed by some shells?).

This branch has 2 commits on it:

1. Change the diff headings to always use file-modified-label.  Whoever wants the old heading can set that option to "comparing:" (which is what one of the modified tests now does, while the other test uses --navigate to have it output 2 Δ headings).
2. Add the handling for --right-arrow=FOO and associated git config parsing.

I've tested the results and things look good to me.  I've also run my changes through clippy and fmt to ensure that they're in good shape.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -550,6 +550,10 @@ pub struct Opt {
     /// Text to display in front of a renamed file path.
     pub file_renamed_label: String,
 
+    #[structopt(long = "right-arrow", default_value = "⟶  ")]
+    /// Text to display with a changed value such as a diff heading, a rename, or a chmod.
+    pub right_arrow: String,
+
     #[structopt(long = "hunk-label", default_value = "")]
     /// Text to display in front of a hunk header.
     pub hunk_label: String,
--- a/README.md
+++ b/README.md
@@ -863,6 +863,9 @@ OPTIONS:
         --file-renamed-label <file-renamed-label>
             Text to display in front of a renamed file path [default: renamed:]
 
+        --right-arrow <right-arrow>
+            Text to display with a changed value such as a diff heading, a rename, or a chmod [default: ⟶  ]
+
         --hunk-label <hunk-label>
             Text to display in front of a hunk header [default: ]
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -73,6 +73,7 @@ pub struct Config {
     pub file_modified_label: String,
     pub file_removed_label: String,
     pub file_renamed_label: String,
+    pub right_arrow: String,
     pub file_style: Style,
     pub git_config_entries: HashMap<String, GitConfigEntry>,
     pub git_config: Option<GitConfig>,
@@ -223,6 +224,7 @@ impl From<cli::Opt> for Config {
         let file_modified_label = opt.file_modified_label;
         let file_removed_label = opt.file_removed_label;
         let file_renamed_label = opt.file_renamed_label;
+        let right_arrow = opt.right_arrow;
         let hunk_label = opt.hunk_label;
 
         let line_fill_method = match opt.line_fill_method.as_deref() {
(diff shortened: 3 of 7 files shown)


src/cli.rs
```diff
@@ -550,6 +550,10 @@ pub struct Opt {
     /// Text to display in front of a renamed file path.
     pub file_renamed_label: String,
 
+    #[structopt(long = "right-arrow", default_value = "⟶  ")]
```
dandavison, line 553: Note to self: there are two trailing spaces in this default value in order to match what was previously hard-coded in various places (whether or not the previous treatment was correct is a separate issue).

dandavison:
Hi @WayneD, thanks very much for this. I agree with both your changes. For example, having navigate work when comparing directory trees with `delta dir1 dir2` or `diff -u dir1 dir2` seems like a nice improvement.

Regarding the spaces after the right arrow, I don't know what's going on. I develop on MacOS using iTerm2 and I think that it's handling of that right-arrow unicode character may be a little buggy.

For example, if I paste `⟶` in to a shell prompt, my line cursor is rendered in the middle of the arrow stem:

Now, if I enter two space characters, it looks like this:

However, I also see similar behavior using the default MacOS Terminal.app emulator:

On paste (where has the head gone?!)

After entering two spaces:

I'm guessing you see more reasonable behavior with your terminal emulator(s)?

WayneMDB:
I see an improper cursor position when editing in vim: my cursor position is off-by-one on the line after a double-wide char. However, default renders fine, with 3 visible spaces after the arrow.  I'm using the ChromeOS terminal into my Linux container.

image

So, moving the arrow into a single spot in the source files has the nice benefit of making the various format lines that use it easier to edit.

WayneD:
I see an improper cursor position when editing in vim: my cursor position is off-by-one on the line after a double-wide char. However, default renders fine, with 3 visible spaces after the arrow.  I'm using the ChromeOS terminal into my Linux container.

image

So, moving the arrow into a single spot in the source files has the nice benefit of making the various format lines that use it easier to edit.

dandavison:
OK great. Not entirely sure whether we should change defaults, I guess it depends on the proportion of users' terminal emulators that are handling this correctly. Neither iTerm2 nor Terminal.app seem to. I've opened https://gitlab.com/gnachman/iterm2/-/issues/10058

(Is that vim as a desktop app, rather than a terminal app?)

Thanks again for the PR!

WayneD:
I use vim as an in-terminal editor (though it does have a gui mode as well). It's for people who like to edit in the old "vi" modal style but with lots of modern features (and I toss in some xemacs line editing into edit mode for good measure).So, it just seems to be unaware that some chars can render double-wide instead of single-wide (and thus it positions the cursor in the wrong spot).

https://en.wikipedia.org/wiki/Vim_(text_editor)

dandavison:
Incidentally, the iTerm2 author gave a very helpful explanation in https://gitlab.com/gnachman/iterm2/-/issues/10058:

image

WayneD:
Since it's such a problematical character, it seems better to switch to one that renders consistently.  How about changing the default to be `🡒` (U+1F852)?  I've switched to that one in my personal ~/.gitconfig.

dandavison:
Hm, neither Chrome nor iTerm2 can render that character on my Mac. (Are you using Linux?)

PR #782, opened 2021-11-18
Navigate regexp

WayneD:
In my navigating of delta diffs, I'd prefer to not stop at each hunk.  My first thought was to override the navigate_regexp value, but it wasn't configurable.  Then I had the idea to change the hunk-label to an empty string, which then made n/N stop at every line.

This branch makes the following changes:

* Turn --navigate-regexp into an option that defaults to an empty string.  When the string is empty, it is auto-populated with a value based on the various labels (the old behavior).
* Change the make_navigate_regexp() func to not add empty regex options to the string.
* Change the code that outputs the hunk-label to output it whenever it is non-empty, not when --navigate is specified. This (1) makes the option work all the time like other labels, and (2) ensures that an empty hunk-label doesn't add a leading space when --navigate is specified.

With these changes I've chosen to not override navigate-regexp and to just force hunk-label to an empty string in my git config. This avoids stopping at each hunk while also making it easy to manually specify an override --hunk-label option when I want to change my mind.  So, if you don't like the first commit in this branch (making navigate_regexp an option) the other 2 commits are good enough for my needs.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -530,9 +530,9 @@ pub struct Opt {
     #[structopt(long = "wrap-right-prefix-symbol", default_value = "…")]
     pub wrap_right_prefix_symbol: String,
 
-    #[structopt(long = "navigate-regex", default_value = "")]
-    /// A regexp to use in the less pager when navigating (auto-generated when empty)
-    pub navigate_regex: String,
+    #[structopt(long = "navigate-regex")]
+    /// A regexp to use in the less pager when navigating (auto-generated when unspecified)
+    pub navigate_regex: Option<String>,
 
     #[structopt(long = "file-modified-label", default_value = "")]
     /// Text to display in front of a modified file path.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -530,9 +530,9 @@ pub struct Opt {
     #[structopt(long = "wrap-right-prefix-symbol", default_value = "…")]
     pub wrap_right_prefix_symbol: String,
 
-    #[structopt(long = "navigate-regexp", default_value = "")]
+    #[structopt(long = "navigate-regex", default_value = "")]
     /// A regexp to use in the less pager when navigating (auto-generated when empty)
-    pub navigate_regexp: String,
+    pub navigate_regex: String,
 
     #[structopt(long = "file-modified-label", default_value = "")]
     /// Text to display in front of a modified file path.
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -530,6 +530,10 @@ pub struct Opt {
     #[structopt(long = "wrap-right-prefix-symbol", default_value = "…")]
     pub wrap_right_prefix_symbol: String,
 
+    #[structopt(long = "navigate-regexp", default_value = "")]
+    /// A regexp to use in the less pager when navigating (auto-generated when empty)
+    pub navigate_regexp: String,
+
     #[structopt(long = "file-modified-label", default_value = "")]
     /// Text to display in front of a modified file path.
     pub file_modified_label: String,
--- a/src/config.rs
+++ b/src/config.rs
@@ -106,7 +106,7 @@ pub struct Config {
     pub minus_file: Option<PathBuf>,
     pub minus_non_emph_style: Style,
     pub minus_style: Style,
-    pub navigate_regex: String,
+    pub navigate_regex: Option<String>,
     pub navigate: bool,
     pub null_style: Style,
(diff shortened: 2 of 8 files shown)


th1000s:
Indeed, the `--navigate-regexp` should be user configurable.

So if I understand correctly, this can make navigation more fine grained by passing something that matches every line? Do you have ideas for other possible regexes which improve navigation?

I've also been playing around with this and want to add a nav marker every term-height/2 changed lines to still be able to jump around when using a lot of context such as `-U200`, but I have not found a good spot to place them (or hide via ANSI magic) yet.

WayneD:
I'm wanting to avoid matching every line.  With the current code, an option of `--hunk-title ''` results in an empty option in the auto-gen regex (`'^(commit|added:|removed:|renamed:|Δ|)'`) which matches every line, making the skipping useless.  The new code avoids adding an empty label in such a case (`'^(commit|added:|removed:|renamed:|Δ)'`) as well as any other empty label (e.g. also specifying `--file-modified-label ''` would result in `'^(commit|added:|removed:|renamed:)'`).

src/cli.rs
```diff
@@ -530,6 +530,10 @@ pub struct Opt {
     #[structopt(long = "wrap-right-prefix-symbol", default_value = "…")]
     pub wrap_right_prefix_symbol: String,
 
+    #[structopt(long = "navigate-regexp", default_value = "")]
+    /// A regexp to use in the less pager when navigating (auto-generated when empty)
+    pub navigate_regexp: String,
```
dandavison, line 535: After anguished debate with myself some months ago I decided to use `regex` as opposed to `regexp` for the regular expression options in the CLI.

src/config.rs
```diff
@@ -106,7 +106,7 @@ pub struct Config {
     pub minus_file: Option<PathBuf>,
     pub minus_non_emph_style: Style,
     pub minus_style: Style,
-    pub navigate_regexp: Option<String>,
+    pub navigate_regexp: String,
```
dandavison, line 109: And if you wouldn't mind making it consistent in the internal variables that would be great too :)

WayneD: Sure, I changed all the navigate regexp to remove the "p".

src/cli.rs
WayneD: Sounds good!

src/cli.rs
```diff
@@ -530,6 +530,10 @@ pub struct Opt {
     #[structopt(long = "wrap-right-prefix-symbol", default_value = "…")]
     pub wrap_right_prefix_symbol: String,
 
+    #[structopt(long = "navigate-regex", default_value = "")]
+    /// A regexp to use in the less pager when navigating (auto-generated when empty)
+    pub navigate_regex: String,
```
dandavison, line 535: Hi @WayneD, this PR seems spot-on to me. Thanks very much. The only change I suggest is that I feel that it's more correct to use `Option<String>` here with no `default_value`. That avoids giving the impression that `""` is a real default value that's going to be used (and is arguably more idiomatic Rust, letting `None` play the role of a null value).

With that change it will appear as `Option<String>` in the `cli::Opt` struct, but will be converted to `String` in the `config::Config` struct.

th1000s:
Apart from the "expert mode" of specifying the navigate-regex directly, might it make sense to specify the navigation stop types in a more user friendly way?

>  invisible stopping point / zero-width space 

Thanks, great idea!

src/cli.rs
WayneD: I finally got around to making this change.  It indeed seems nicer.  The code does treat both None and an empty string as being unspecified since that allows someone to override a default value in their git config with the auto-gen value by specifying an empty string to --navigate-regex.

dandavison:
Great, thanks! I'll merge when the tests pass.

WayneD:
> Apart from the "expert mode" of specifying the navigate-regex directly, might it make sense to specify the navigation stop types in a more user friendly way?

That is an interesting idea, since you could then leave hunk off the list without hard-wiring the other labels to particular values. That makes me think that some kind of normally invalid regex idiom, such as `{added}` and `{hunk}` might be an interesting way to specify such a thing.  So, the default would be similar to:
```
^(commit|{added}|{removed}|{renamed}|{modified}|{label})
```
(though the default does also auto-remove empty labels and their prior vertical bar).  To avoid the stop-on-every-line issue with this sort of label idiom an empty label could be turned into something like `[^\s\S]` that would make that option never match (without having to remove the vertical bar).

PR #783, opened 2021-11-18
Enable faking `extract_args()` parameters from process utils

th1000s:
The fake value is valid for exactly one call, the state for this is stored per-thread.

### DIFF
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -221,6 +221,15 @@ where
     P: ProcessInterface,
     F: Fn(&[String]) -> ProcessArgs<T>,
 {
+    #[cfg(test)]
+    {
+        if let Some(args) = tests::cfg::WithArgs::get() {
+            match extract_args(&args) {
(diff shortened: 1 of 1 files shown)


dandavison:
See https://github.com/dandavison/delta/pull/774#issuecomment-974899069 I'm wondering whether there's an interaction between threads. (But equally possible I'm doing something stupid due to limited understanding of the clever stuff going on here!)

dandavison:
Thanks very much for this! I hope you don't mind me erring on the side of merging fairly aggressively (once the branch seems basically correct) and applying any improvements / beautification subsequently in master. My experience has been that the fewer unmerged feature branches the better in order to maintain velocity with multiple collaborators. What will suffer somewhat is the beauty of our git history.

th1000s:
No problem. Of course ideally this PR and the one depending on it would inform each others design, but that would indeed be too slow. Beautification PR now up :)

Issue #152, opened 2020-04-28
➕Does delta support `diff --word-diff`?


ryanberckmans:
Does delta support `diff --word-diff`?

AvnerCohen:
In a quick check - no. Would be awesome to add though.

dandavison:
Hi @ryanberckmans (and @AvnerCohen), thanks for this! Could you expand on exactly what you'd like to see here?

In a sense Delta does provide something like `--word-diff`, in that it infers and highlights within-line edits (using a Levenshtein-type DP algorithm). However, I'm not saying that there wouldn't be benefit in working with git's native output, and also extending and improving the algorithms implemented by Delta (see #140).

dandavison:
Are you thinking of replacing git's default (and ugly) `{+` and `[-` markers with background colors? (Presumably in this mode, Delta's native within-line edit inference would be disabled).

```
_SUPPORTED_PYTHONS = [-['3.7',-]{+['3.8', '3.7',+} '3.6', '3.5', '3.4',[-'3.3',-] '2.7']
```

AvnerCohen:
How about something like:

_SUPPORTED_PYTHONS = [~['3.7',]~{**['3.8', '3.7',**} '3.6', '3.5', '3.4',[-'3.3',-] '2.7']

With relevant color scheme applied as well ?

What would be nice is if the color is applied not across all lin, but just on the +- parts.

dandavison:
@AvnerCohen could you comment on how you see the role of delta's default within-line edit highlighting versus `--word-diff`, for examples when would you use one versus the other? Do you have examples of cases where Delta's within-line edit inference is not doing a good job? (Delta's with-line edits (like GitHub, GitLab) are displayed on two different lines, unlike `--word-diff` which combines the added and removed on a single line.) Note that you may want to play with `--max-line-distance` to adjust Delta's within-line edit inference algorithm.

Again, I'm not suggesting we shouldn't support `--word-diff`, I just want to understand how people view those two different, but somewhat similar, features.

AvnerCohen:
@dandavison Thanks for the time and effort.
First, I have checked back and seems like I got a bit confused, I actually meant "--color-words" and not "word-diff". 
That being said, I have done couple of tests, and while I remember specific cases where `color-words` was pretty helpful, I couldn't really find a specific case where this would be better than the correct default you suggest, with the specific part of the word highlighted, for me, it feels like it's just a matter of getting used to it.

[...]

PR #785, opened 2021-11-20
New assignment of emph and non-emph styles

### LINKED ISSUE #776: Can't decide if actually a 🐛: Issue with semantics around emph- and non-emph- styles
I have been using a custom diff with git, for a while now, but I have always wanted a fully syntax-highlighted diff, so I am very excited to use Delta and configure it for daily driving use.

I have a sort of conceptual/semantic issue with the fact that there are three color styles. The issue applies to a specific, not rare, situation, where a token is entirely removed or entirely added. 

Let me just show a simple example, with different diff viewing options:

1. git diff
2. diff-highlight
3. diff-so-fancy 
4. My character based diff that uses background color to indicate minus/plus
5. Delta

image

Overall, what I want to do for Delta is to improve its rapid scannability. What I mean by this is minimizing the time taken for my brain to comprehend the change. As lovely as syntax highlighting is, I have come to the unfortunate realization that I think I need to crank this metric back up more before I can daily drive Delta and I really really want to do so. 

Although I do not use (2) or (3), I will rank these 5 options' readability as follows:

### 4, 2, 3, 5, 1

(Let me know if you agree...)

The deficiency in Delta that this highlights is that the line 51 add hunk is shown with `plus-style`, rather than `plus-non-emph-style`. The reason that this is such a big deal is that I have, over the years, found the in-your-face background color contrast mode to be the most effective way to communicate the notion of change in a diff. What this requires is that the highest contrast red and green background colors unambiguously indicate change. 

So, the way that the emphasis works is inconsistent with my desired visual semantics, because as Delta works currently, there are two scenarios in which `plus-style` is used:

- Line is new or entirely changed: **I want to display these lines in bold and with highest visual contrast.**
- Some complete token was removed from (or in the symmetrical case, added to) this line, leaving nothing to emphasize in it: **I do not want to mark the entire line in bold with highest visual contrast; I want to mark it with a low contrast similar to `non-emph-style`.**

Config used in the screenshots is as follows (minus is all the same just on the red channel): 

image

I hope this makes sense. In order to achieve the optimal degree of visual clarity of actual changes made, it seems that I need one more class of style to apply to differentiate between the above two classes of entirely changed line. Basically I think it's best if it is possible to configure this thing so that we can use a style to completely unambiguously specify which literal characters are insertion/deletion characters and which are unchanged context characters. The current implementation of the `{plus,minus}-style` is not capable of satisfying this principle.

----------------

If you'll humor me in continuing down this path of psychovisual efficiency optimization...

- Well, first off I have a request to be able to eliminate the empty line at the top of the Delta hunks. If there is already an option for this, I apologize. But I do not want to lose a line of real estate for no reason.

- Side by side diffs are nice but the problem I have with them is that it takes longer to look through them for me. In particular moving my eyes left and right generates fatigue, not unlike reaching for the mouse, but marginally less acceptable to complain about. Rather than complain, I propose an alternative. I gravitate towards the unified diff for this reason and less duplicated information means information density. Now what this makes me think, though, is, and I've never seen this implemented, the following enhancement to unified mode:
  
  ### **What if we apply the same concept of visual alignment to a unified diff in the transverse axis?**

I will illustrate by generating a fully colored mockup:

image

Which is not far off from maybe:

image

Which really makes me feel like I just want to do it my old way with the plain character diff, but to try to shoehorn syntax highlighting into it somehow... But no, I think the syntax will make character diff actually unreadable with real-life-gnarly code. I think that I should definitely be okay with duplicating the lines: both sets of lines would be
- both highlighted in a dim color and physically immediately adjacent, to indicate no physical character change in those zones (rather than being side by side and physically separated by a scores/hundreds of chars)
- syntax highlighted based on these actual old/new line contents, which is a pretty big deal, allows us to potentially even see syntax errors

It's the physical alignment that I think will really speed up comprehension.

dandavison:
Fixes #776 cc @unphased @th1000s 

This PR implements the suggestion https://github.com/dandavison/delta/issues/776#issuecomment-974493226 addressing @unphased's proposal in #776 to change the way emph and non-emph colors are applied.

To make this easier to think about and discuss, I've added a commit to the repo which gives a simple demo of the various emph and non-emph styles, https://github.com/dandavison/delta/commit/fd9592ee149c783fc50dc056d88784d37940443e:

As per https://github.com/dandavison/delta/issues/776#issuecomment-974493226, what this branch is intended to do is allow lines where "nothing has happened" to be styled differently from a completely new or completely deleted line. In other words, allow the pink "dodo" line above to be styled differently than the emphasized "albatross" line, as GitHub does.

That is done by changing the code so that the pink dodo line gets `minus-non-emph-style`, whereas previously it was receiving `minus-style`.  With delta's defaults that will have no effect, since the two styles are identical, but we can differentiate the two styles, e.g. with

```gitconfig
[delta]
    features = github-theme

[delta "github-theme"]
    light = true
    pink-style = normal "#ffe0e0"
    dark-pink-style = normal "#ffc0c0"
    green-style = syntax "#d0ffd0"
    dark-green-style = syntax "#a0efa0"

    minus-emph-style = github-theme.dark-pink-style
    minus-non-emph-style = github-theme.pink-style
    plus-emph-style = github-theme.dark-green-style
    plus-non-emph-style = github-theme.green-style

    minus-style = minus-emph-style  # this is the difference;
    plus-style = plus-emph-style  # by default these are equal to *-non-emph
```
and that config gives rise to the middle panel here:

<table>
<thead>
<td>delta 0.9.1</td>
<td>delta [this branch]</td>
<td>GitHub</td>
</thead>
<tr>
<td>image</td>
<td>image</td>
<td>image</td>
</tr>
</table>

So that looks promising.

Here's a more complex example (1973650144cd63302730b37066ce1ba314dfef58). The new branch is noisier in this example, because it's emphasizing entirely deleted lines, but then using a non-emph color to fill right.

If anyone is able to offer thoughts on whether this branch is a good direction and what if anything more needs to be done, then that would be great!

<table>
<thead>
<td>delta 0.9.1</td>
<td>delta [this branch]</td>
<td>GitHub</td>
</thead>
<tr>
<td>image</td>
<td>image</td>
<td>image</td>
</tr>
</table>

### DIFF
--- a/etc/examples/styles.py
+++ b/etc/examples/styles.py
@@ -1,4 +1,4 @@
-# albatross albatross albatross albatross
+# avocet avocet avocet avocet
 # bee-eater bee-eater bee-eater bee-eater
-# chaffinch chaffinch chaffinch chaffinch
-# dodo dodo dodo dodo
+# chaffinch chaffinch chaffinch
+# dodo dodo dodo dodo dodo
--- a/etc/examples/styles.py
+++ b/etc/examples/styles.py
@@ -1,0 +1,4 @@
+# albatross albatross albatross albatross
+# bee-eater bee-eater bee-eater bee-eater
+# chaffinch chaffinch chaffinch chaffinch
+# dodo dodo dodo dodo
--- a/src/edits.rs
+++ b/src/edits.rs
@@ -8,10 +8,16 @@ use crate::align;
 /// Infer the edit operations responsible for the differences between a collection of old and new
 /// lines. A "line" is a string. An annotated line is a Vec of (op, &str) pairs, where the &str
 /// slices are slices of the line, and their concatenation equals the line. Return the input minus
-/// and plus lines, in annotated form. Also return a specification of the inferred alignment of
-/// minus and plus lines. `noop_deletions[i]` is the appropriate deletion operation tag to be used
-/// for `minus_lines[i]`; `noop_deletions` is guaranteed to be the same length as `minus_lines`.
-/// The equivalent statements hold for `plus_insertions` and `plus_lines`.
+/// and plus lines, in annotated form.
+///
+/// Also return a specification of the inferred alignment of minus and plus lines: a paired minus
+/// and plus line is represented in this alignment specification as
+/// (Some(minus_line_index),Some(plus_line_index)), whereas an unpaired minus line is
+/// (Some(minus_line_index), None).
+///
+/// `noop_deletions[i]` is the appropriate deletion operation tag to be used for `minus_lines[i]`;
+/// `noop_deletions` is guaranteed to be the same length as `minus_lines`. The equivalent statements
+/// hold for `plus_insertions` and `plus_lines`.
 #[allow(clippy::too_many_arguments)]
 #[allow(clippy::type_complexity)]
 pub fn infer_edits<'a, EditOperation>(
--- a/src/edits.rs
+++ b/src/edits.rs
@@ -4,6 +4,7 @@ use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
 use crate::align;
+use crate::minusplus::MinusPlus;
 
 /// Infer the edit operations responsible for the differences between a collection of old and new
 /// lines. A "line" is a string. An annotated line is a Vec of (op, &str) pairs, where the &str
@@ -95,6 +96,24 @@ where
     }
 
     (annotated_minus_lines, annotated_plus_lines, line_alignment)
+}
+
+// Return boolean arrays indicating whether each line has a homolog (is "paired").
+pub fn make_lines_have_homolog(
+    line_alignment: &[(Option<usize>, Option<usize>)],
+) -> MinusPlus<Vec<bool>> {
+    MinusPlus::new(
+        line_alignment
+            .iter()
+            .filter(|(m, _)| m.is_some())
+            .map(|(_, p)| p.is_some())
+            .collect(),
+        line_alignment
+            .iter()
+            .filter(|(_, p)| p.is_some())
+            .map(|(m, _)| m.is_some())
+            .collect(),
+    )
 }
 
 /// Split line into tokens for alignment. The alignment algorithm aligns sequences of substrings;
--- a/src/features/side_by_side.rs
+++ b/src/features/side_by_side.rs
@@ -85,7 +85,7 @@ pub fn line_is_too_long(line: &str, line_width: usize) -> bool {
 /// structure indicating which of the input lines are too long. This avoids
 /// recalculating the length later.
 pub fn has_long_lines(
-    lines: &LeftRight<&[(String, State)]>,
+    lines: &LeftRight<&Vec<(String, State)>>,
     line_width: &line_numbers::SideBySideLineWidth,
 ) -> (bool, LeftRight<Vec<bool>>) {
     let mut wrap_any = LeftRight::default();
@@ -108,7 +108,7 @@ pub fn has_long_lines(
 
 #[allow(clippy::too_many_arguments)]
 pub fn paint_minus_and_plus_lines_side_by_side(
-    lines: LeftRight<&[(String, State)]>,
+    lines: LeftRight<&Vec<(String, State)>>,
     syntax_sections: LeftRight<Vec<LineSections<SyntectStyle>>>,
     diff_sections: LeftRight<Vec<LineSections<Style>>>,
     line_alignment: Vec<(Option<usize>, Option<usize>)>,
@@ -315,6 +315,7 @@ fn paint_right_panel_plus_line<'a>(
     panel_line
 }
 
+#[allow(clippy::too_many_arguments)]
 fn get_right_fill_style_for_panel<'a>(
     line_is_empty: bool,
     line_index: Option<usize>,
--- a/src/features/side_by_side.rs
+++ b/src/features/side_by_side.rs
@@ -6,6 +6,7 @@ use crate::ansi;
 use crate::cli;
 use crate::config::{self, delta_unreachable, Config};
 use crate::delta::State;
+use crate::edits;
 use crate::features::{line_numbers, OptionValueFunction};
 use crate::minusplus::*;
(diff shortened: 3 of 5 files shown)


unphased:
The new version, especially in the more involved example (thanks for that one, very nice complex example!) is far superior to the shown GitHub output! Kudos! I snagged both versions of your orphaned commits out of GitHub and ran a diff with my char-diff that I described in my thread, here it is for the purposes of edification. I tried a little bit to make a light background but it does not look very good, so I stuck with my usual dark theme... 

image

unphased:
I do wonder, and let's not take my comment here as creeping the scope of your PR here. but:

I think I came up with a point of improvement (open debate!) for multi-line groups of changes. Let me illustrate with delta output:

a

I've got to spend a few seconds working out that I need to hop 4 lines downward to find the corresponding green line and see that no chars got added. 

The reason for having to hop down by 4 lines instead of 1 line is because all minus-lines in a set of changes get grouped together. 

I wonder if this separation is necessary. It sure makes it easier to read the previous code state. But I also worry that there are ambiguities if we try to interleave the lines. We could place the two purple underline lines right next to each other, but I do not know where to place the next one. 
I love how my custom diff never has this ambiguity because it just jumbles it all in there, it shows both the begin state and end state and all common characters and you can mentally work out which is which as long as you can mentally block out one or the other bg-color and keep the non-bg-colored lines in. From my perspective by ensuring no duplication of unchanged characters ever appears, it sidesteps all ambiguities arising from attempting to generate some least confusing way to visually present these duplicated unchanged characters.

I worry that after using my solution for so long, I've spoiled myself, I worry that perhaps it simply looks like an incomprehensible jumble to most everyone, while to me it is the most efficient. What I am grateful for is that in writing this out I think I've begun to articulate a sort of proof for why my character diff colorization output is the tersest possible way to present a diff. 

I haven't made up my mind yet on whether I want it to be a goal of mine to convince you that we should have this as a possible output. I think I still need to think about what the possibilities are when it comes to syntax highlighting done on the unchanged shared characters.

dandavison:
Thanks @unphased! I think there are two or three different issues here. How about we use this ticket for resolving this question:

> Do you think we need a fourth emphasis mode or is there maybe a cleaner way to address the issue I pointed out?

I.e. deciding changes we are going to make to delta's basic unified diff mode (where you have separate removed and added lines), and whether this branch should be merged as-is.

On that topic, I'm a little uncomfortable with the extra visual noise that is introduced when filling right with non-emph color. It's nice that the avocet and albatross lines are emphasized:

<table>
<tr>
<td>image</td>
</tr>
</table>
 

but on the other hand, this branch is producing the left panel when `plus-style=plus-emph-style`, whereas GitHub users are used to the right panel. Is this really what we want, or the change we want to make in this PR slightly more nuanced than what I have right now?

<table>
<thead>
<td>delta [this branch]</td>
<td>GitHub</td>
</thead>
<tr>
<td>image</td>
<td>image</td>
</tr>
</table>

--------------------------------------------------------

Regarding your script, that's definitely interesting. So that's essentially a new type of output: neither unified, nor side-by-side, but "combined" or something, i.e. there's only one copy of each line. So in that respect it's reminiscent of git's `--word-diff` output. If you agree with that distinction, would you mind opening a separate issue to discuss the possibility of delta's supporting a "combined" output format? Or, shall we use

- https://github.com/dandavison/delta/issues/152
?

And then the possibility of re-ordering lines is again a distinct issue I think.

But my top priority right now is -- I'd love your help deciding what exactly is the correct change to unified diff (this branch) to release.

unphased:
Sorry for my delay, I think that I may want to look into this specific question of yours @dandavison further by doing some experiments, but... I'm heavily leaning toward considering github's output a bug and your proposed output superior. 

In particular what master-branch Delta and Github share is the main deficiency i highlight in the beginning of my #776 issue. There is one and the same style being applied by Github to the contextual portions of modified lines and wholly added or deleted lines, which is my pet peeve because it breaks the semantic meaning of the background color on a characterwise basis. 

I think in the rendering what you've got on wholly new lines is plus-emph on all chars and plus- till the beginning of the next line. 

I could see somebody considering this style transition at the end of a line to be curious/weird/bad. 

I think what's behind this is as I suggested earlier an introduction of a fourth style rather than to change all plus lines to be emph because at least that way it would be easy for a user to configure it to (or even to make Delta behave like this) the way Github looks.

unphased:
What I am unsure about, though, is could there be a smarter way to semantically break down the styles, still using only 3, and achieve all of our goals. That i have also not totally thought through.

dandavison:
> My own aesthetic preference would be to have the darker bgcolor extend all the way to end of line on "pure" lines (the rationale is that the emph'd to-end-of-line maps one-to-one with newline insertion)

Me too and I've changed the branch to do that.

@unphased  I think I agree with all or nearly all that you wrote. I've merged this branch. Try out the master branch -- I believe that it makes most or all of the things you're proposing for unified diffs possible.

> Pure added lines should look the same as portions of lines that were added. Unchanged regions I would want colored (lightly) to indicate whether it is part of a minus-line-context or plus-line-context region.

Right. So to do that basically all you have to do is

```gitconfig
    minus-style = minus-emph-style
    plus-style = plus-emph-style
```

Below is an example theme that does it (I would really like to have examples of themes that work well on dark terminal backgrounds; please let me know if you have some!)

```gitconfig
[delta]
    features = hoopoe2 consistent-color-semantics

[delta "consistent-color-semantics"]
    # All added/removed content is styled the same, regardless of
    # whether there is unchanged content on the same line.
    minus-style = minus-emph-style
    plus-style = plus-emph-style

[delta "hoopoe2"]
    light = true

    pink = "#ffe0e0"
    dark-pink = "#ffc0c0"
    green = "#d0ffd0"
    dark-green = "#a0efa0"

    minus-emph-style = normal hoopoe2.dark-pink
    minus-non-emph-style = normal hoopoe2.pink
    plus-emph-style = syntax hoopoe2.dark-green
    plus-non-emph-style = syntax hoopoe2.green
```

dandavison:
>  I will be running master starting from now. Is it true that installation requires nothing other than placing `delta` into `$PATH`? How convenient.

Great. Yes that's right, delta is just an executable. There are no man pages or anything else to install.

Issue #72, opened 2019-12-11
Support --color-moved


dwijnand:
Git recently added `--color-moved` support, to colour moved code: https://stackoverflow.com/a/48166435/463761

Would it be possible to add support for that in delta?

dandavison:
Thanks @dwijnand, I didn't know about this feature in git. Here are some thoughts, let me know if this sounds right to you.

From a quick look at the docs and play around with the feature, it looks to me like git uses color alone to distinguish "moved" code from "changed" code: i.e. git does not, for example, use special `-` and `+` markers for moved blocks.

Now, delta cannot assume that it has access to the original color codes from git. So that suggests that we would have to re-implement git's algorithms for moved code detection. That in itself sounds like a blocker: it would be fragile to attempt to track the details of their algorithm.

However, if git emitted (or could be configured to emit) special `+`/`-` markers for moved code blocks then the situation would be entirely different and we could definitely have delta apply special colors to them. I do wonder whether it would make sense for git to do that. After all, without that, nothing can programmatically obtain the information  from git _output_.

Currently delta only uses the input text and does not read anything from the git repo (i.e. delta does not use a library like libgit2, whereas e.g. bat does). So in fact delta doesn't care or know what repo the input comes from and may not even have access to that repo. I don't know if something like libgit2 might supply information about moved code blocks now or in the future.

In addition to being fragile to reimplement git's algorithms it would also be a potential perfomance hit: delta is currently implemented in a streaming fashion: as it moves through the input, it only stores a small amount of local context in memory. But for delta to do moved code detection, it seems to me that in the worst case it would have to read the diff for one file into memory before starting to emit any output (e.g. if it encountered a `-` block at the beginning of the diff, it would need to read to the end of the file before knowing whether to color that as deleted or moved. I don't know how often that would result in a noticeable performance hit for delta users. Git has the repo available to it and thus probably can do this more efficiently.

What do you think? Am I understanding the situation correctly?

dwijnand:
Thanks for the detailed write-up!  I didn't spend time thinking about it too much, but I can share my thoughts based on your analysis.

> Now, delta cannot assume that it has access to the original color codes from git.

What if:
1. it uses the colour codes you've separately configured in delta; or
2. it uses the `.git/config`/`~/.gitconfig`-configured colour codes, provided it can access them; or
3. it just uses git's defaults (or bombs)

?

dandavison:
[...]

PR #796, opened 2021-11-25
Add simpler colorMoved config snipped, add highlights to all codeblocks

SuperSandro2000:
I had quite some trouble understanding the quite complex colorMoved example and added a simpler one which I got more or less from https://github.com/dandavison/delta/issues/72#issuecomment-971794591 to better understand it.

Feel free to adopt it.

Also I think the map-syntax and --from-ansi feature should get their own head lines instead of being crammed into movedColor but I am not fully sure how the layout should be for those.

PS: thank you so much for this tool. It is really awesome! ❤️


dandavison (approved):
Thanks a lot! I agree. This documentation was a bit rushed.

PR #800, opened 2021-11-25
Remove trailing \r from raw line, in presence of ANSI escape sequences

### LINKED ISSUE #795: 🐛 git's --color-moved with DOS line endings (\r\n) triggers a panic
Using delta 92d6591e and git version 2.30.2. To reproduce:

    $ echo -e "move_this_long_function()\\nfoo()\\nbar()" | sed 's,$,\r,' > a.py
    $ { tail -n 2 a.py; head -n 1 a.py; } > b.py 
    $ git diff --color-moved --no-index --color=always a.py b.py | target/release/delta --no-gitconfig 
    
    thread 'main' panicked at 'String mismatch encountered while superimposing style sections: '
    '', src/paint.rs:921:17
    
    [..]
    4: delta::paint::Painter::paint_buffered_minus_and_plus_lines
    5: delta::delta::delta
    6: delta::main
    $ cat -A a.py
    move_this_long_function()^M$
    foo()^M$
    bar()^M$

dandavison:
Fixes #795 cc @th1000s

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -137,6 +137,17 @@ impl<'a> StateMachine<'a> {
     fn ingest_line(&mut self, raw_line_bytes: &[u8]) {
         // TODO: retain raw_line as Cow
         self.raw_line = String::from_utf8_lossy(raw_line_bytes).to_string();
+        // When a file has \r\n line endings, git sometimes adds ANSI escape sequences between the
+        // \r and \n, in which case byte_lines does not remove the \r. Remove it now.
+        if let Some(cr_index) = self.raw_line.rfind('\r') {
+            if ansi::strip_ansi_codes(&self.raw_line[cr_index + 1..]).is_empty() {
+                self.raw_line = format!(
+                    "{}{}",
+                    &self.raw_line[..cr_index],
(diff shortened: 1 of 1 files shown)


src/delta.rs
```diff
@@ -152,13 +163,6 @@ impl<'a> StateMachine<'a> {
...
         };
         self.line = ansi::strip_ansi_codes(&self.raw_line);
-
-        // Strip the neglected CR.
-        // (CR-LF is unfortunately split by git because it adds ansi escapes between them.
-        //  Thus byte_lines library can't remove the CR properly.)
-        if let Some(b'\r') = self.line.bytes().nth_back(0) {
-            self.line.truncate(self.line.len() - 1);
-        }
     }
```
dandavison, line 166: I owe some tests, but we can see that with this commit we are still fixing the `git add -p` pbug that was the reason for the addition of the lines above (now deleted and replaced by the lines further above):

```
git add a.py && git commit -m .
mv b.py a.py
```

Without any \r-stripping, notice that one of the lines is not rendered due to the \r:

With the code after this commit:

PR #801, opened 2021-11-26
Refactor: further isolate side-by-side implementation

dandavison:
Refactoring to move some side-by-side-specific code paths out of `paint.rs` into `features/side-by-side.rs`.


dandavison:
cc @th1000s I think this is fairly uncontroversial, but do shout out if you have opinions.

PR #805, opened 2021-11-27
Process cleanup

th1000s:
- Test sibling process detection
- Match process binary case insensitively (e.g. `Git.exe`)
- Clean up FakeParentArgs, now has `once`, `for_scope`, or `with`

### DIFF
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -46,31 +46,27 @@ pub enum ProcessArgs<T> {
 }
 
 pub fn git_blame_filename_extension() -> Option<String> {
-    calling_process_cmdline(ProcInfo::new(), blame::guess_git_blame_filename_extension)
+    calling_process_cmdline(ProcInfo::new(), guess_git_blame_filename_extension)
 }
 
-mod blame {
-    use super::*;
+pub fn guess_git_blame_filename_extension(args: &[String]) -> ProcessArgs<String> {
+    let all_args = args.iter().map(|s| s.as_str());
 
-    pub fn guess_git_blame_filename_extension(args: &[String]) -> ProcessArgs<String> {
-        let all_args = args.iter().map(|s| s.as_str());
+    // See git(1) and git-blame(1). Some arguments separate their parameter with space or '=', e.g.
+    // --date 2015 or --date=2015.
+    let git_blame_options_with_parameter =
+        "-C -c -L --since --ignore-rev --ignore-revs-file --contents --reverse --date";
 
-        // See git(1) and git-blame(1). Some arguments separate their parameter with space or '=', e.g.
-        // --date 2015 or --date=2015.
-        let git_blame_options_with_parameter =
-            "-C -c -L --since --ignore-rev --ignore-revs-file --contents --reverse --date";
+    let selected_args =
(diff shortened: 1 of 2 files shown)


src/utils/process.rs
```diff
@@ -421,58 +437,105 @@ where
...
+        static FAKE_ARGS: RefCell<TlsState<Vec<String>>> = RefCell::new(TlsState::None);
+    }
+
+    #[derive(Debug, PartialEq)]
+    enum TlsState<T> {
+        Once(T),
+        Scope(T),
+        With(usize, Rc<Vec<T>>),
+        None,
+        Invalid,
```
dandavison, line 456: Would `Exhausted` or `Used` or `Consumed` be a better name?

(The code here is all quite sophisticated by my Rust standards so I'm really just trying to understand the PR and may not have great suggestions.)

src/utils/process.rs
```diff
@@ -421,58 +437,105 @@ where
...
+                ),
+            );
+            if FAKE_ARGS.with(|a| a.replace(with)) != TlsState::None || args.is_empty() {
+                Self::error("with creation");
             }
+            FakeParentArgs {}
+        }
+        pub fn get() -> Option<Vec<String>> {
+            FAKE_ARGS.with(|a| {
+                let old_value = a.replace_with(|old_value| match old_value {
```
dandavison, line 501: would `let new_value = a.replace_with...` be a better name?

src/utils/process.rs
```diff
@@ -630,22 +705,78 @@ pub mod tests {
...
+        let _args = FakeParentArgs::once(&"git blame do.not.panic");
+        assert_eq!(
+            calling_process_cmdline(ProcInfo::new(), guess_git_blame_filename_extension),
+            Some("panic".into())
+        );
 
-            calling_process_cmdline(ProcInfo::new(), blame::guess_git_blame_filename_extension);
-        }
+        calling_process_cmdline(ProcInfo::new(), guess_git_blame_filename_extension);
+    }
```
dandavison, line 715: I think that this module is getting to the point where it would be easier to understand if we separated the abstract machinery (e.g. `ProcessInterface`, `calling_process_cmdline` , `FakeParentArgs`) from the concrete clients (`determine_calling_process`, `git_blame_filename_extension`). E.g. I feel that this test is starting to mix up testing `git blame` and `FakeParentArgs` logic. Happy for that to be done in subsequent PRs though.

src/utils/process.rs
```diff
@@ -421,58 +437,105 @@ where
...
+                }
+            })
+        }
+        fn error(where_: &str) -> ! {
+            panic!(
+                "test logic error (in {}): wrong FakeParentArgs scope?",
+                where_
+            );
+        }
+    }
```
dandavison, line 523: These interior mutability pointer tricks are new to me, very nice! I'd heard of them while reading Blandy & Orendorff but hadn't tried to understand code using them before. Your comment above was sufficient though to explain what was going on, if taken slowly.

dandavison (approved):
Great, and very educational. I made couple of comments -- let me know if you want to change anything or if I should merge first.

https://github.com/dandavison/delta/pull/807 is branched off this branch, although not in any meaningful way yet.

PR #814, opened 2021-11-30
Fix colorMoved in git config example

maximd:
According to [git documentation for diff.colorMoved](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---color-movedltmodegt), `true` is not a correct value. `default` is a correct value and it also matches the "`--color-moved` support" section of this README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
     diffFilter = delta --color-only
 
 [diff]
-    colorMoved = true
+    colorMoved = default
 
 [delta]
     navigate = true


dandavison:
Actually, `true` is documented as a valid value for `diff.colorMoved` in gitconfig: [https://git-scm.com/docs/git-config#Documentation/git-config.txt-diffcolorMoved](git-config.txt-diffcolorMoved). And it seems to be accepted although not documented as a value for `diff --color-moved`:

```
delta(master *) git -P diff --color-moved=true > /dev/null; echo $?
0
delta(master *) git -P diff --color-moved=truex > /dev/null; echo $?
error: color moved setting must be one of 'no', 'default', 'blocks', 'zebra', 'dimmed-zebra', 'plain'
error: bad --color-moved argument: truex
129
```

However, I think I agree with you that `default` is better, especially given git's apparently inconsistent handling of `true`.

dandavison (approved):
Thanks again!

PR #815, opened 2021-11-30
Fix navigate option in git config example

maximd:
The navigate option does not work when present in the `[core]` section of the git config because it is not a git option. It works correctly when present in the `[delta]` section.


dandavison:
Thank you @maximd! That was careless of me.

PR #831, opened 2021-12-06
Fix more side-by-side line number miscounts

th1000s:
I found more cases where the line number count is off, now these must be increased _and_ decreased afterwards.

And regarding DeltaTest, do bikeshed to your hearts content!

### DIFF
--- a/src/ansi/mod.rs
+++ b/src/ansi/mod.rs
@@ -155,6 +155,25 @@ fn strip_ansi_codes_from_strings_iterator<'a>(
         .join("")
 }
 
+pub fn explain_ansi(line: &str, colorful: bool) -> String {
+    use crate::style::Style;
+
+    parse_style_sections(line)
+        .into_iter()
+        .map(|(ansi_term_style, s)| {
+            let style = Style {
+                ansi_term_style,
+                ..Style::default()
+            };
+            if colorful {
+                format!("({}){}", style.to_painted_string(), style.paint(s))
+            } else {
+                format!("({}){}", style.to_string(), s)
+            }
+        })
+        .collect()
+}
+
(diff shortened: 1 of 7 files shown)


dandavison:
`DeltaTest` looks fantastic! I don't think I have any bikeshedding to do there, at least on a first look.

Have you thought much about how to test that ANSI sequences in the output are correct? In 114ae670223520657208501a3245a3b4261c1093 I added a subcommand that converts ANSI sequences into a human-readable form:

```
git log -p --color=always | delta --parse-ansi | less -RS
```

I was wondering whether it might work to write test assertions using the same ASCII scheme for representing the colors. (Obviously the parenthesized format is something I just made up on the spot and isn't parseable; open to suggestions.)

th1000s:
> ANSI sequences in human-readable form

Nice feature and a a good idea, now `DeltaTest` also has `.explain_ansi()` to turn the ANSI-gibberish into something more readable. This also works with `.inspect()` so the result can just be pasted into `.expect()` when developing and testing. The format looks fine to me as it does not have to be machine readable.

And as planned, the bugfix slips by undetected!

dandavison:
Fantastic. I'm looking forward to porting more tests over the this, and using it for new tests, and trying out some ANSI assertions (an area where delta has hitherto been somewhat sketchy).

dandavison:
It looks like there's another side-by-side line numbers bug in the middle hunk here:

```git -c delta.side-by-side=true show 829-color-words-2```

dandavison:
> It looks like there's another side-by-side line numbers bug in the middle hunk here:

Sorry, total nonsense! It was getting late when I wrote that. The lines are different due to the added line in the first hunk.

PR #833, opened 2021-12-07
Fix --word-diff / --color-words handling

### LINKED ISSUE #829: 🐛 --color-words output doesn't show hunk headers or line numbers
From https://github.com/dandavison/delta/issues/826#issuecomment-986919350

> the handling of --color-words is not quite what I had expected in v0.10.3 either. I suppose it makes sense given that delta is only passing along the raw values, but for a file with multiple hunks, I'm not getting either line numbers from delta or the hunk headers from the raw git. 

It looks like the implementation in https://github.com/dandavison/delta/pull/807 isn't quite right: it's causing the hunk lines to be emitted raw by *not* handling them. But I think that what is needed is more complex: to transition into one of the hunk line states of the state machine, emit raw instead of the normal emission, and report the line as having been handled. I think that this is because we are not transitioning the state correctly

dandavison:
Fixes #829 WIP

https://github.com/dandavison/delta/pull/807 was a pretty bad attempt at handling this correctly: while the colored content from git was emitted raw, in so doing they were not recognized as hunk lines and we lost line numbers and hunk headers.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -17,6 +17,7 @@ use crate::fatal;
 use crate::features::navigate;
 use crate::features::side_by_side::{self, ansifill, LeftRight};
 use crate::git_config::{GitConfig, GitConfigEntry};
+use crate::handlers;
 use crate::minusplus::MinusPlus;
 use crate::paint::BgFillMethod;
 use crate::parse_styles;
@@ -291,7 +292,7 @@ impl From<cli::Opt> for Config {
             } else {
                 line_fill_method
             },
-            line_numbers: opt.line_numbers,
+            line_numbers: opt.line_numbers && !*handlers::hunk::IS_WORD_DIFF,
             line_numbers_format: LeftRight::new(
                 opt.line_numbers_left_format,
                 opt.line_numbers_right_format,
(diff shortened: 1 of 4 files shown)


dandavison:
@mrjoel this branch fixes `--word-diff` / `--color-words` support, but, there's no support for `line-numbers` (or `side-by-side`) -- both will be silently ignored if `--word-diff` / `--color-words` input is detected.

For `line-numbers` the reason is that it would be a bit more work to implement: to calculate line numbers in the "before" and "after" state, delta normally relies on the `-` and `+` prefixes to indicate removed and added lines, but with `--word-diff` we don't have those and we'd have to detect wholly added/removed lines from the ANSI color sequences alone. In contrast, the implementation of `--word-diff` handling in this branch is based on processing all lines in the same way that unchanged (context) lines are usually processed.

For `side-by-side` the reason is that (a) one usually uses this with line numbers and (b) I don't think it makes much sense with `--word-diff` output since lines with both additions and removals will be the same in the left and right panels.

dandavison:
@th1000s de37c9921e3833f1f5240c8d24e0bfd09b3618fc adds a builder method `DeltaTest::with_calling_process(command: &str)` allowing your `FakeParentArgs` functionality to be accessed easily in `DeltaTest`. Hopefully you agree with the spirit of the commit if not the exact implementation :)

PR #836, opened 2021-12-07
Clarify ReadMe hyperlinks field

anyu:
Clarify that hyperlinks param is a config under `[delta]` section (rather than `[blame]`)

```
[delta]
    hyperlinks = true
...
```


dandavison:
Thanks very much @anyu.

Issue #824, opened 2021-12-06
🐛 delta is much slower starting with `0.10.3` (`0.11.0` included)


infokiller:
delta is much slower for me starting from `0.10.3`. I was able to reproduce that without any gitconfig customizations.
I will try to add a docker command to reproduce this, but wanted to open an issue in case others run into this.

dandavison:
Hi @infokiller (and @nymous) thanks. Would you be able to use `make benchmark` in the delta repository to obtain timings? You would need to install the Rust dev tool chain but that should be straightforward. If you can't install the dev tool chain, you can still obtain a timing for your installed version of delta by editing the Makefile target named `benchmark` to change the path to the delta executable.

infokiller:
Thanks for the quick response @dandavison (and wonderful project of course!).
Yes, I didn't know about the benchmark target, I'll run it and post results.

dandavison:
Great! (And you'll see you need to install `hyperfine`).

dandavison:
What's your OS?

infokiller:
```
❯ uname -a
Linux desk18 5.14.18-hardened1-1-hardened #1 SMP PREEMPT Sun, 14 Nov 2021 16:10:03 +0000 x86_64 GNU/Linux
```

infokiller:
Benchmark results are not significantly different:

```
❯ make benchmark
cargo build --release
   Compiling git-delta v0.10.2 (/home/user/src/config/delta)
    Finished release [optimized] target(s) in 11.10s
git log -p 23c292d3f25c67082a2ba315a187268be1a9b0ab > /tmp/delta-benchmark-input.gitdiff
hyperfine 'target/release/delta --no-gitconfig < /tmp/delta-benchmark-input.gitdiff > /dev/null'
Benchmark 1: target/release/delta --no-gitconfig < /tmp/delta-benchmark-input.gitdiff > /dev/null
  Time (mean ± σ):      2.563 s ±  0.074 s    [User: 1.493 s, System: 0.935 s]
  Range (min … max):    2.484 s …  2.737 s    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet PC without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

make benchmark | 37.170 total | 67.89s user | 10.46s system | 210% cpu | 569 MB memory
❯ gco 0.11.0
Previous HEAD position was 2f21f08 Bump version
[...]

Issue #839, opened 2021-12-08
🐛  0.10.x introduced `sysinfo` crate result in very slow speed under Linux


ttys3:
## brief description

0.10.x introduced `sysinfo` crate result in very slow speed under Linux.

**a simple delta diff cost about `300ms` (with 16 logic core cpu)**

Append: 

just test on another Linux computer, 6 core CPU

**a simple delta diff cost about `100ms` (with 6 logic core cpu)**

so for each core,  `get_cpu_frequency()` call cost about `15ms`, if you CPU core is less than 10, this problem is not easy to found, because you are not very obvious to feel the delay gap.

**for any >= 0.10.x  version, the more (logic) CPU cores you have, the more slower you get**

this issue is also mentioned in https://github.com/dandavison/delta/pull/841#issuecomment-989363683

I also tried  delta-0.9.2-x86_64-unknown-linux-gnu  (https://github.com/dandavison/delta/releases/download/0.9.2/delta-0.9.2-x86_64-unknown-linux-gnu.tar.gz) 

which works without slowdown problem on my machine.

the problem seems to be `get_cpu_frequency()` call in `sysinfo` crate

my patched version resolved the slowdown problem:  https://github.com/ttys3/delta/releases/tag/0.11.2

related commit:  https://github.com/ttys3/delta/commit/8f5a8356a815bdf4aca3e73a0ee2c765637cad8a

https://github.com/ttys3/sysinfo/commit/53cf3ddd68cdeea9e544c1183366047394f8700b

since this need to patch the `sysinfo` crate, it is hard to make a `normal` PR -_-

-------------------------------------------------------------------------------------

## my investigation

here's my env:

```
------------
OS: Arch Linux x86_64
Kernel: 5.15.6-arch2-1
Shell: zsh 5.8
Resolution: 3840x2160
Terminal: tmux
CPU: Intel i9-9900 (16) @ 5.000GHz
GPU: NVIDIA GeForce GTX 1060 3GB
Memory: 6747MiB / 32038MiB

```

-----------------------------------------------------------------

my git config related to delta:

```ini
[pager]
    #delta
    log = delta --color-only | less -RFX
    show = delta --color-only | less -RFX
    diff = delta --color-only | less -RFX
```
-----------------------------------------------------------------

## 1. test with git diff

with  delta 0.11.1, do `time git diff Cargo.toml ` cost about `0.5`s each time (repeated 3 times):

```
❯ delta --version
delta 0.11.1

[...]

PR #841, opened 2021-12-08
Query fewer processes when searching for the parent

th1000s:
This query now happens for more invocation types, so speed it up.
Call `refresh_process()` only on pids numerically close to the one
of delta itself.

-----
Follow-up to https://github.com/dandavison/delta/issues/824#issuecomment-988920890:

This now checks the pid range of $delta_pid - 10 / + 20 when looking at the parent and predicted sibling finds nothing.

@infokiller @jebaum  @ttys3 

Could you apply the patch (3) from https://github.com/dandavison/delta/issues/824#issuecomment-988350304 and run `git log -p | time delta` on top of this and see if the slowdown is gone and the calling git process is still found?

Also, is this even required on macOS or is querying the entire process structure less expensive there?

And was the slowdown ever observed when delta was configured as the pager in the git config and `git show` (no pipe to delta) was called directly? It really should not, because the parent process should always be found.

### DIFF
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -275,16 +275,11 @@ trait ProcessInterface {
         self.refresh_process(sibling_pid).then(|| ())?;
         self.process(sibling_pid)
     }
-    #[cfg(test)]
-    fn find_sibling_process<F, T>(&mut self, pid: Pid, extract_args: F) -> Option<T>
+    fn find_sibling_in_refreshed_processes<F, T>(&mut self, pid: Pid, extract_args: &F) -> Option<T>
     where
         F: Fn(&[String]) -> ProcessArgs<T>,
         Self: Sized,
     {
-        self.refresh_processes();
-
-        let this_start_time = self.process(pid)?.start_time();
-
         /*
 
         $ start_blame_of.sh src/main.rs | delta
@@ -302,6 +297,8 @@ trait ProcessInterface {
         Find the common ancestor processes, calculate the distance, and select the one with the shortest.
 
         */
+
+        let this_start_time = self.process(pid)?.start_time();
 
         let mut pid_distances = HashMap::<Pid, usize>::new();
         let mut collect_parent_pids = |pid, distance| {
@@ -407,7 +404,8 @@ where
     started, so no command line can be parsed. Same if the data was piped from an input file.
 
     There might also be intermediary scripts in between or piped input with a gap in pids or (rarely)
-    randomized pids, so check all processes for the closest match in the process tree.
+    randomized pids, so check processes for the closest match in the process tree.
+    The size of this process tree can be reduced by only refreshing selected processes.
 
     100 /usr/bin/some-terminal-emulator
     124  \_ -shell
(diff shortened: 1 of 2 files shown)


dandavison:
Here's my understanding of the situation: evidence has been put forward for two distinct causes of slowness:

1. **Querying many processes**
  A mitigation for this has been released (https://github.com/dandavison/delta/pull/838) which is drastic: it entirely removed the call to `find_sibling_process`. This PR should help with this cause of slowness in that, while it will not make the released version faster, it will make it better at finding processes.

2. **Querying CPU data** (linux only?)
  @ttys3 has found that delta is still unacceptably slow, even with https://github.com/dandavison/delta/pull/838, because of the way that `sysinfo` is querying for CPU data. This PR will not improve that cause of slowness. It may make it worse because of more calls to the `sysinfo` library? I haven't looked into details there.

Is that summary in line with your understanding @th1000s and others?

If that summary is accurate, should we avoid releasing this change until we have solved cause (2)?

ttys3:
@th1000s 
I've made a simple demo to re-produce the Linux slowdown problem

see https://github.com/dandavison/delta/issues/839#issuecomment-989439820

dandavison:
There's some very helpful information from @ttys3 about the CPU-querying performance problems here: https://github.com/dandavison/delta/issues/839#issuecomment-989433383

> so for each core, get_cpu_frequency() call cost about 15ms, if you CPU core is less than 10, this problem is not easy to found, because you are not very obvious to feel the delay gap.

> basically, I think this affects all Linux users.

src/utils/process.rs
```diff
@@ -426,19 +424,31 @@ where
...
-    #[cfg(not(test))]
-    {
-        None
+
+    match info.find_sibling_in_refreshed_processes(my_pid, &extract_args) {
+        None => {
+            #[cfg(test)]
+            {
+                info.refresh_processes();
+                info.find_sibling_in_refreshed_processes(my_pid, &extract_args)
```
dandavison, line 441: Out of interest can you explain why we make this recursive call in the tests? (Not blocking merge as this is a test-only code path).

dandavison:
Thanks @th1000s! I'm merging this now (I took it out of Draft state -- hope that was ok) since both @ttys3 and @unphased have reported good performance with a branch containing this commit, in addition to the no-cpu-querying `sysinfo` commit. To confirm that this branch improves performance when there are many processes I've run it with @infokiller's script:

```
Benchmarking version: 0.10.2
Running command "git diff --no-index a b | delta-0.10.2 >| delta-0.10.2-result" 10 times
101.60 ms per run (1016 ms total)
Benchmarking version: 0.11.0
Running command "git diff --no-index a b | delta-0.11.0 >| delta-0.11.0-result" 10 times
509.40 ms per run (5094 ms total)
Benchmarking version: 0.11.2
Running command "git diff --no-index a b | delta-0.11.2 >| delta-0.11.2-result" 10 times
108.00 ms per run (1080 ms total)
Benchmarking version: procs
Running command "git diff --no-index a b | delta-procs >| delta-procs-result" 10 times
100.70 ms per run (1007 ms total)
```

Meanwhile, I can see that this commit improves calling-process-finding at least in the following experiment:

1. Set `core.pager=delta.sh` where delta.sh contains `bash -c delta`
2. Observe that `git grep` output is handled correctly with this commit, and incorrectly with this commit reverted.

th1000s:
Sure, I just wanted to make sure this got tested first, hence the draft :)

PR #847, opened 2021-12-10
Store history file in data_local_dir for Windows

### LINKED ISSUE #772: 🚀 Store .delta.lesshst in a location other than Home (Windows)
The `get_delta_less_hist_file` function stores the delta history file in the Home directory on Windows. While on *nix, it stores it in an XDG compliant directory.

https://github.com/dandavison/delta/blob/93d37649ae63312548bb319e5d6d7305834e502f/src/features/navigate.rs#L84-L96

Files beginning with a dot (.) are not hidden files on Windows - this is a *nixism. As such, the home directory looks cluttered when applications create these files.

I'd like to request that the this file be created somewhere else - the official place to store them is in ~/AppData/Local (rough equivalent of XDG_DATA_HOME). The function [`dirs_next::data_local_dir()`](https://docs.rs/dirs-next/2.0.0/dirs_next/fn.data_local_dir.html) can be used to get this.

rashil2000:
Closes #772

### DIFF
--- a/src/features/navigate.rs
+++ b/src/features/navigate.rs
@@ -90,9 +90,11 @@ pub fn copy_less_hist_file_and_append_navigate_regex(config: &Config) -> std::io
 
 #[cfg(target_os = "windows")]
 fn get_delta_less_hist_file() -> std::io::Result<PathBuf> {
-    let mut path = dirs_next::home_dir()
-        .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't determine home dir"))?;
-    path.push(".delta.lesshst");
+    let mut path = dirs_next::data_local_dir()
+        .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't determine LocalAppData dir"))?;
+    path.push("delta");
+    std::fs::create_dir_all(&path)?;
+    path.push("delta.lesshst");
     Ok(path)
 }
 
--- a/src/features/navigate.rs
+++ b/src/features/navigate.rs
@@ -91,7 +91,7 @@ pub fn copy_less_hist_file_and_append_navigate_regex(config: &Config) -> std::io
 #[cfg(target_os = "windows")]
 fn get_delta_less_hist_file() -> std::io::Result<PathBuf> {
     let mut path = dirs_next::data_local_dir()
-        .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't determine LocalAppData dir"))?;
+        .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't find AppData\\Local folder"))?;
     path.push("delta");
     std::fs::create_dir_all(&path)?;
     path.push("delta.lesshst");


dandavison:
In case it's helpful at any point, the original PR adding Windows and non-Windows `navigate` history files is https://github.com/dandavison/delta/pull/526

src/features/navigate.rs
```diff
@@ -90,9 +90,11 @@ pub fn copy_less_hist_file_and_append_navigate_regex(config: &Config) -> std::io
 
 #[cfg(target_os = "windows")]
 fn get_delta_less_hist_file() -> std::io::Result<PathBuf> {
-    let mut path = dirs_next::home_dir()
-        .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't determine home dir"))?;
-    path.push(".delta.lesshst");
+    let mut path = dirs_next::data_local_dir()
+        .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't determine LocalAppData dir"))?;
```
dandavison, line 94: I think my only question is this:

How often will `data_local_dir` return `None` on users' Windows systems? Should we fall back to `home_dir`, or something that we can guarantee will exist?

rashil2000: The LocalAppData folder has been a part of Windows since XP if I remember correctly (it might even be older, but I don't think any Rust toolchain would support that far back).

I think it is safe to assume it will always exist, unless a user explicitly deletes it (which is highly unlikely, because this is a hidden folder). Even after deletion, Windows will recreate it.

dandavison: Thanks @rashil2000. 

> "Can't determine LocalAppData dir"

Would this be slightly more helpful to Windows users if it said

> "Can't find AppData\Local folder"

?

dandavison: (I guess I have in mind making it helpful to less expert as well as more expert users)

rashil2000: Hmm, I think that makes sense. 

Before Vista, the location was actually different, but we can leave that out for the sake of brevity.

dandavison: Great, thanks @rashil2000. 

For the record, I see that we're not currently emitting this error message: https://github.com/dandavison/delta/blob/master/src/utils/bat/output.rs#L165-L170

But that's not your fault and a distinct issue from what you're solving here. (I haven't been quite sure how/whether to alert the user to a problem like that: a fatal error is maybe too much? But randomly emitting messages on stderr interleaved with their expected output on stdout is not very elegant.)

dandavison: Confirmed that your message is displayed as expected on macOS with this patch:

```diff
diff --git a/src/features/navigate.rs b/src/features/navigate.rs
index db582dd..a12e87e 100644
--- a/src/features/navigate.rs
+++ b/src/features/navigate.rs
@@ -1,5 +1,4 @@
 use std::io::Write;
-#[cfg(target_os = "windows")]
 use std::io::{Error, ErrorKind};
 use std::path::PathBuf;
 
@@ -88,22 +87,15 @@ pub fn copy_less_hist_file_and_append_navigate_regex(config: &Config) -> std::io
     Ok(delta_less_hist_file)
 }
 
-#[cfg(target_os = "windows")]
 fn get_delta_less_hist_file() -> std::io::Result<PathBuf> {
-    let mut path = dirs_next::data_local_dir()
-        .ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't find AppData\\Local folder"))?;
+    let mut path: PathBuf =
+        None.ok_or_else(|| Error::new(ErrorKind::NotFound, "Can't find AppData\\Local folder"))?;
     path.push("delta");
     std::fs::create_dir_all(&path)?;
     path.push("delta.lesshst");
     Ok(path)
 }
 
-#[cfg(not(target_os = "windows"))]
-fn get_delta_less_hist_file() -> std::io::Result<PathBuf> {
-    let dir = xdg::BaseDirectories::with_prefix("delta")?;
-    dir.place_data_file("lesshst")
-}
-
 // LESSHISTFILE
 //        Name of the history file used to remember search commands
 //        and shell commands between invocations of less.  If set to
diff --git a/src/utils/bat/output.rs b/src/utils/bat/output.rs
index 62761c2..3e390a4 100644
--- a/src/utils/bat/output.rs
+++ b/src/utils/bat/output.rs
@@ -162,10 +162,15 @@ fn _make_process_from_less_path(
         p.env("LESSCHARSET", "UTF-8");
         p.env("LESSANSIENDCHARS", "mK");
         if config.navigate {
-            if let Ok(hist_file) = navigate::copy_less_hist_file_and_append_navigate_regex(config) {
-                p.env("LESSHISTFILE", hist_file);
-                if config.show_themes {
-                    p.arg("+n");
+            match navigate::copy_less_hist_file_and_append_navigate_regex(config) {
+                Ok(hist_file) => {
+                    p.env("LESSHISTFILE", hist_file);
+                    if config.show_themes {
+                        p.arg("+n");
+                    }
+                }
+                Err(err) => {
+                    fatal(format!("Error: {}", err));
                 }
             }
         }
```

dandavison:
Thanks very much for the work here @rashil2000. If you're interested in making any more improvements to delta that would of course be fantastic. We have a guide to the codebase here: [ARCHITECTURE.md](https://github.com/dandavison/delta/blob/master/ARCHITECTURE.md), and I am always happy to answer questions about development directions, implementation, etc.

rashil2000:
Thank you @dandavison!

I use delta all day, and will be very happy to contribute to it! But I'm very new to Rust itself, so it might be a while before I make any productive contribution 😅

src/features/navigate.rs
rashil2000: > But that's not your fault and a distinct issue from what you're solving here. (I haven't been quite sure how/whether to alert the user to a problem like that: a fatal error is maybe too much? But randomly emitting messages on stderr interleaved with their expected output on stdout is not very elegant.)

Is there an issue open for this? I'll take a look and try to gather what I can.

dandavison: There's no open issue yet. (There are arguably more important bugs though!)

dandavison:
> I use delta all day, and will be very happy to contribute to it! 

Excellent!

> But I'm very new to Rust itself, so it might be a while before I make any productive contribution 😅

I wouldn't be so sure of that :) -- there's quite a lot of code in delta providing examples of how things are done, and lots of tests providing examples of how to test things, so it might not be a bad way to learn Rust.

Maybe the most important thing is finding something to work on which you personally care about, and which is also the right level of difficulty for an initial project.

PR #849, opened 2021-12-12
Add features from DELTA_FEATURES env var instead of replacing with them

### LINKED ISSUE #848: 🐛 DELTA_FEATURES env var should add to other features, not replace them
I think it would be more useful if `DELTA_FEATURES` added to features specified in git config, rather than replacing them as it does today. For example, this would make it more convenient as a way to toggle `side-by-side` mode. I'm thinking that if a user does need to deactivate the features in git config they have `--no-gitconfig`.

dandavison:
Fixes #848

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -320,10 +320,10 @@ pub struct Opt {
     pub color_only: bool,
 
     ////////////////////////////////////////////////////////////////////////////////////////////
-    #[structopt(long = "features", default_value = "", env = "DELTA_FEATURES")]
+    #[structopt(long = "features")]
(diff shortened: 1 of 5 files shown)


michaelblyons:
Can you tweak [the Release text for v0.11.3](https://github.com/dandavison/delta/releases/tag/0.11.3)? It seems to be missing an initial `` ` `` on two lines.

dandavison:
> Can you tweak the Release text for v0.11.3? It seems to be missing an initial ` on two lines.

Thanks! Fixed.

PR #859, opened 2021-12-17
Honor default-language option

dandavison:
Fixes https://github.com/dandavison/delta/discussions/858 

The `default-language` option had been introduced initially in order to determine language for `git blame` (where there are no file names). However, subsequently `git blame` came to determine the language by inspecting the command line of the calling process (e.g. if delta sees the calling process is `git blame src/file.ts` then the language is typescript). Meanwhile `default-language` acquired its obvious function of providing a default language for diffs; this PR fixes that.

Note that the best place to set `default-language` is probably in your per-repository git settings. I.e. `$my_repo/.git/config`.

### DIFF
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -70,7 +70,8 @@ pub enum StyleSectionSpecifier<'l> {
 
 impl<'p> Painter<'p> {
     pub fn new(writer: &'p mut dyn Write, config: &'p config::Config) -> Self {
-        let default_syntax = Self::get_syntax(&config.syntax_set, None);
+        let default_syntax =
+            Self::get_syntax(&config.syntax_set, None, config.default_language.as_deref());
 
         let panel_width_fix = ansifill::UseFullPanelWidth::new(config);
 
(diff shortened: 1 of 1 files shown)


imbrish:
I had the same intuition about the `default-language` option as described in #858, namely wanted to use it to assume `bash` syntax for the files without an extension.

But it did not work, and thus I found the discussion and this PR. I wonder what stopped this from being merged?

imbrish:
@dandavison I have just rebased this branch onto main - it went smoothly - the code does what it is supposed to and all tests pass. I could take it from here, but not sure if there was something else you had in mind with this - any chance you remember? 

At the least I could add tests for the new (and obvious) function of `default-language` described above.

Can you allow me to push to this branch? Or should I supersede this pull request with a new one?

dandavison:
Hi @imbrish, great! Thanks for looking at this. Please go ahead and supersede this with a new branch. I think this functionality makes sense to have (I'd probably use it for the same as you -- executable shell files). I can't remember why I didn't merge it, it's possible it was accidentally forgotten about.

michaelblyons:
Is this PR obsolete now?

dandavison:
Yep! Thanks.

Superseded by https://github.com/dandavison/delta/pull/1655

PR #861, opened 2021-12-20
Remove redundent spaces in README example

TennyZhuang:
The config examples may be copied to user config file, and the redundant spaces are ugly.


dandavison:
Agreed, thank you!

PR #869, opened 2021-12-22
Speed up the calling process query

th1000s:
Always determining the parent process is faster now that this information is needed most of the time anyhow.

Also fixes the Linux process query slowdown for good.

### DIFF
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -364,7 +364,11 @@ trait ProcessInterface {
                     };
                     iter_parents(self, pid, &mut sum_distance);
 
-                    Some((length_of_process_chain, args))
+                    if length_of_process_chain == usize::MAX {
+                        None
+                    } else {
+                        Some((length_of_process_chain, args))
+                    }
                 }
                 _ => None,
             })
@@ -846,6 +850,20 @@ pub mod tests {
         );
         // ignored: dropping causes a panic while panicing, so can't test
         calling_process_cmdline(ProcInfo::new(), guess_git_blame_filename_extension);
+    }
+
+    #[test]
+    fn test_process_blame_no_parent_found() {
+        let two_trees = MockProcInfo::with(&[
+            (2, 100, "-shell", None),
+            (3, 100, "git blame src/main.rs", Some(2)),
+            (4, 100, "call_delta.sh", None),
+            (5, 100, "delta", Some(4)),
+        ]);
+        assert_eq!(
+            calling_process_cmdline(two_trees, guess_git_blame_filename_extension),
+            None
+        );
     }
 
     #[test]
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -226,6 +226,13 @@ struct ProcInfo {
 }
 impl ProcInfo {
     fn new() -> Self {
+        // On Linux sysinfo optimizes for repeated process queries and keeps per-process
+        // /proc file descriptors open. This caching is not needed here, so
+        // set this to zero (this does nothing on other platforms).
+        // Also, there is currently a kernel bug which slows down syscalls when threads are
+        // involved (here: the ctrlc handler) and a lot of files are kept open.
+        sysinfo::set_open_files_limit(0);
+
         ProcInfo {
             info: sysinfo::System::new(),
         }
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -1,9 +1,9 @@
-use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::path::Path;
-use sysinfo::{Pid, Process, ProcessExt, ProcessRefreshKind, SystemExt};
+use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 
 use lazy_static::lazy_static;
+use sysinfo::{Pid, Process, ProcessExt, ProcessRefreshKind, SystemExt};
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum CallingProcess {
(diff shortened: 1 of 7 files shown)


src/utils/process.rs
```diff
@@ -23,23 +25,48 @@ pub struct CommandLine {
...
+    // it does not need to know its parent process.
+    std::thread::Builder::new()
+        .name("find_calling_process".into())
+        .spawn(move || {
+            let caller = determine_calling_process();
+
+            let (caller_mutex, determine_done) = &**CALLER;
+
+            let mut calling_process = caller_mutex.lock().unwrap();
+            *calling_process = caller;
```
dandavison, line 44: What do you think about rearranging the variable names like this? I think it's a bit more consistent.

```suggestion
            let calling_process = determine_calling_process();

            let (caller_mutex, determine_done) = &**CALLER;

            let mut caller = caller_mutex.lock().unwrap();
            *caller = calling_process;
```

src/utils/process.rs
```diff
@@ -23,23 +25,48 @@ pub struct CommandLine {
...
+        })
+        .unwrap();
+}
+
+#[cfg(not(test))]
+pub fn calling_process() -> MutexGuard<'static, CallingProcess> {
+    let (caller_mutex, determine_done) = &**CALLER;
+    determine_done
+        .wait_while(caller_mutex.lock().unwrap(), |pending| {
+            *pending == CallingProcess::Pending
```
dandavison, line 55:
```suggestion
        .wait_while(caller_mutex.lock().unwrap(), |caller| {
            *caller == CallingProcess::Pending
```

src/utils/process.rs
```diff
@@ -23,23 +25,48 @@ pub struct CommandLine {
...
+lazy_static! {
+    static ref CALLER: Arc<(Mutex<CallingProcess>, Condvar)> =
+        Arc::new((Mutex::new(CallingProcess::Pending), Condvar::new()));
 }
 
-lazy_static! {
-    static ref CACHED_CALLING_PROCESS: Option<CallingProcess> = determine_calling_process();
+pub fn start_determining_calling_process_in_thread() {
+    // Do not keep the join handle so the main thread can exit early if
+    // it does not need to know its parent process.
```
dandavison, line 35: This comment is referring to the fact that there is no return value of this function (and that we use the `spawn` method on the builder as opposed to the `thread::spawn` function), right?

src/utils/process.rs
```diff
@@ -226,6 +253,13 @@ struct ProcInfo {
 }
 impl ProcInfo {
     fn new() -> Self {
+        // On Linux sysinfo optimizes for repeated process queries and keeps per-process
+        // /proc file descriptors (fds) open. This caching is not needed here, so
+        // set this to zero (this does nothing on other platforms).
+        // Also, there is currently a kernel bug which slows down syscalls when threads are
+        // involved (here: the ctrlc handler) and a lot of files are kept open.
```
dandavison, line 260: Out of interest, have you located this bug in the linux kernel issue tracker?

dandavison: So this is what allows us to fearlessly query many processes on linux again? Very interesting to see that you've got to the bottom of this problem.

src/utils/process.rs
dandavison, line 55: Beautiful. And very useful to me to have an example of how to do this.

src/utils/process.rs
th1000s: The builder is used to set the thread name - in case someone without access to the sources wonders why delta looks at all the processes. It could also return a `JoinHandle`, but this is ignored. I updated the comments to reflect this.

src/utils/process.rs
th1000s: As the Linux bug tracker suggests, for now I have reported it to the Debian kernel team, see [#1001818](https://bugs.debian.org/1001818) for a minimal C and C++ example.

And yes, on Linux this makes process queries as fast as e.g. `ps auxf`.

th1000s:
`panic at 'attempt to subtract with overflow', [..] sysinfo-0.22.3/src/lib.rs:132:28` -- (`let diff = max - **x;`)

Looks like on 32bit sysinfo does not like (concurrent?) `set_open_files_limit(0)`, @GuillaumeGomez?

Also, `#[cfg(test)] pub fn calling_process() -> Box<CallingProcess>` is currently called 4412 times during `cargo test`.

GuillaumeGomez:
> Also, `#[cfg(test)] pub fn calling_process() -> Box<CallingProcess>` is currently called 4412 times during `cargo test`.

You mean in `sysinfo`? As far as I can see, there is no `calling_process` function.

GuillaumeGomez:
Forgot to mention but I opened https://github.com/GuillaumeGomez/sysinfo/pull/650. Please tell me if it fixes your issue.

dandavison:
@GuillaumeGomez thanks for helping with this. I've created a version of this PR with your sysinfo fix: https://github.com/dandavison/delta/pull/872

It appears that it is still failing, due to the subtraction on the line before the line that you fixed:

```
thread 'tests::test_example_diffs::tests::test_single_character_line_is_not_whitespace_error'
panicked at 'attempt to subtract with overflow',
/cargo/git/checkouts/sysinfo-bc2293244598544b/62374ac/src/lib.rs:132:28
```

https://github.com/GuillaumeGomez/sysinfo/blob/62374ac97393dec978d44a888f6b0907b211938d/src/lib.rs#L132-L133

GuillaumeGomez:
I updated the PR, both subtractions are checked now.

dandavison:
Great, looks like it works now. I'll leave it up to @th1000s to finalize things here.

GuillaumeGomez:
Ok, I'll release a new sysinfo version then.

th1000s:
Thanks, version 0.22.4 fixes the crash.

And the `calling_process()` function is from delta, this patchset also reduces the number of times it is called during testing.

dandavison:
Thanks @th1000s and @GuillaumeGomez!

dandavison:
The docker script provided by @infokiller in https://github.com/dandavison/delta/issues/973 seems to show that having if there are many processes then delta is still significantly slowed down.

PR #872, opened 2021-12-28
Speed up the calling process query -- with sysinfo fix

dandavison:
Ref #869
Ref https://github.com/GuillaumeGomez/sysinfo/pull/650


dandavison:
This was just a test branch for https://github.com/dandavison/delta/pull/869

PR #874, opened 2021-12-30
Put file-modified label in front of mode changes.

WayneD:
When using `--navigate` with files that get their mode modified, the file-modified label was not being prefixed, so these changes were not being marked as skip destinations.  This is particularly bad if a file has line changes along with the mode change **and** the user has an empty hunk label (since that would make the entire file's changes get skipped with an "n").

I added a test of a mode-change with a line-change, and a test for a mode change where the old & new mode values are not one of the two expected value-pairs.  I also used format_file() on the mode filenames since the other format() headings were using it.

### DIFF
--- a/src/tests/test_example_diffs.rs
+++ b/src/tests/test_example_diffs.rs
@@ -8,6 +8,7 @@ mod tests {
     use crate::tests::ansi_test_utils::ansi_test_utils;
     use crate::tests::integration_test_utils;
     use crate::tests::test_utils;
+    use regex::Regex;
 
     #[test]
     fn test_added_file() {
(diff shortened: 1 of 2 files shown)


dandavison:
LGTM, thanks @WayneD.

src/tests/test_example_diffs.rs
```diff
@@ -1606,6 +1607,40 @@ src/align.rs:71: impl<'a> Alignment<'a> { │
...
 
+    #[test]
+    fn test_file_mode_change_unexpected_bits() {
+        let config =
+            integration_test_utils::make_config_from_args(&["--navigate", "--right-arrow=->"]);
+        let output =
+            integration_test_utils::run_delta(GIT_DIFF_FILE_MODE_CHANGE_UNEXPECTED_BITS, &config);
+        let output = strip_ansi_codes(&output);
+        assert!(output.contains(r"Δ src/delta.rs: 100700 -> 100644"));
+    }
```
dandavison, line 1618: Incidentally, there's a utility to make writing these tests easier now. It doesn't yet have an `expect_contains()` method but I'll add that, at which point this test will be able to be written as

```rust
    fn test_file_mode_change_unexpected_bits() {
        DeltaTest::with(&["--navigate", "--right-arrow=->"])
            .with_input(GIT_DIFF_FILE_MODE_CHANGE_UNEXPECTED_BITS)
            .expect_contains(r"Δ src/delta.rs: 100700 -> 100644");
    }
```

WayneD: Thanks for pointing that out.  I've created a pull request that implements this (along with some other changes).

PR #875, opened 2021-12-30
Handle a mode change on a renamed file.

WayneD:
I changed the diff parsing to cache the mode info from the old/new mode lines until the parsing bumps into the start of the actual related diff, finds the diff line for an unrelated diff, or runs out of input.  This allows the mode info to be output in conjunction with a file event instead of as a separate heading (that used to have an empty name when the mode change was for a rename).

The mode info is passed down into the draw routines as a separate "addendum" string that the draw code can use as it likes. Currently that means that it appends the addendum string to the non-raw string in parens. I imagine that future draw routines could decide to put it in a separate box or some other per-routine method. There is currently a single function in src/handlers/draw.rs that joins the strings in the same way for everyone.

A couple examples of how the new code looks:

```
Δ foo.rs (mode +x)
───────────────────────────────────────────────────
```

```
renamed: old-longer-name 🡒 shorter-name (mode +x)
───────────────────────────────────────────────────
```

Would it look better on its own line?

```
Δ foo.rs
• mode +x
───────────────────────────────────────────────────
```

```
renamed: old-longer-name 🡒 shorter-name
• mode +x
───────────────────────────────────────────────────
```

Would it look better appended after a "•" character?

```
Δ foo.rs • mode +x
───────────────────────────────────────────────────
```

```
renamed: old-longer-name 🡒 shorter-name • mode +x
───────────────────────────────────────────────────
```

Should it be a user option? If so, we can do that later.


dandavison:
Hi @WayneD, thanks for working on this!

I haven't thought about how this should work -- can I check, is this the intended behavior?

```
$ git mv src/delta.rs src/delta.1.rs
$ chmod +x src/delta.1.rs
$ git add src/
$ git diff --cached | delta --no-gitconfig

src/delta.rs: mode +x
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

renamed: src/delta.rs ⟶   src/delta.1.rs
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

src/handlers/diff_header.rs
```diff
@@ -262,16 +262,32 @@ fn parse_diff_header_line(
...
+                return Some(first_path);
+            }
+        }
+        if llen >= 7 && line[0] == "a" && line[1] == "/" {
+            for j in 3..llen - 3 {
+                if line[j] == " " && line[j + 1] == "b" && line[j + 2] == "/" {
+                    let first_path = _parse_file_path(&line[..j].join(""), true);
+                    return Some(first_path);
+                }
+            }
```
dandavison, line 290: I'm thinking that, instead of adding this logic here, perhaps we should modify `_parse_file_path` to return a richer data type and pattern match here as required? For example, `_parse_file_path` supports multiple prefixes via the  `DIFF_PREFIXES` variable, rather than hard-coding support for `a/` and `b/`.

WayneD:
> I haven't thought about how this should work -- can I check, is this the intended behavior?

Yeah, that's what I ended up with given the current way the parsing works.  My thoughts on an alternative method that delays the mode info pending more parsing would allow the 2 bits of info to be output in a more unified manner. Combined with your thoughts on wanting improved prefix handling, I think it would be better to go with the delayed-mode method (which should mean that the diff line is only ever used when no renaming is going on, and thus the current splitting function wouldn't need to change).

dandavison:
> that's what I ended up with given the current way the parsing works. My thoughts on an alternative method that delays the mode info pending more parsing would allow the 2 bits of info to be output in a more unified manner. Combined with your thoughts on wanting improved prefix handling, I think it would be better to go with the delayed-mode method

That sounds good. I agree that, seeing as you're working on this, it makes sense to try to achieve an output format that you're really happy with. What do you think, something like this maybe? (This is just the first thing that occurred to me; I haven't thought through the problem/solution space properly.)

> renamed: src/delta.rs ⟶   src/delta.1.rs (mode +x)

WayneD:
I fixed the conflicts with the DeltaTest strings now that the delta-test branch has merged.  You'll note that my latest version uses parentheses to display the mode info, which I thought was a good choice in the various output idioms.

src/handlers/diff_header.rs
```diff
@@ -76,20 +67,48 @@ impl<'a> StateMachine<'a> {
...
+    pub fn handle_diff_header_mode_line(&mut self) -> std::io::Result<bool> {
+        if !self.line.starts_with("old mode ") && !self.line.starts_with("new mode ") {
+            return Ok(false);
+        }
+        self.state = State::DiffHeader(DiffType::Unified);
+        if !self.should_handle() || self.config.color_only {
+            return Ok(false);
+        }
+        if self.line.starts_with("old") {
+            self.mode_info = self.line[9..].to_string();
```
dandavison, line 87: Can we make it clear where 9 comes from, either via the code or in a comment? Could this perhaps be rewritten as

```rust
if let Some(mode_info) = self.line.strip_prefix("old mode ")
```
?

(and below for `new_mode`)

src/handlers/diff_header.rs
```diff
@@ -76,20 +67,48 @@ impl<'a> StateMachine<'a> {
...
                 &mut self.painter,
+                &mut self.mode_info,
                 self.config,
             )?;
             handled_line = true;
         }
         Ok(handled_line)
     }
 
+    pub fn handle_diff_header_mode_line(&mut self) -> std::io::Result<bool> {
```
dandavison, line 78: (I think we should add brief doc comments to all public functions/methods although I don't think I've been great about doing that everywhere. I've been using `///` and `//` inconsistently, I don't think that matters too much until we make a library out of delta and publish the documentation)

src/handlers/diff_header.rs
```diff
@@ -154,7 +167,43 @@ impl<'a> StateMachine<'a> {
...
+            let name = get_repeated_file_path_from_diff_line(&self.diff_line)
+                .unwrap_or_else(|| "".to_string());
+            let line = format!("{}{}", label, format_file(&name));
+            let _got = write_generic_diff_header_header_line(
+                &line,
+                &line,
+                &mut self.painter,
+                &mut self.mode_info,
+                self.config,
+            );
```
dandavison, line 205: Rather than assigning to an unused variable to avoid handling the case where this returns an error, I believe the more correct implementation would be along the lines of the diff below. I.e. change this function to return a `Result` enum, and propagate the error handling upwards. This also fits the pattern in `delta::consume` where functions named `handle_foo` all return a `Result` (although this one is slightly different as it doesn't return a bool saying whether it has handled a line).  

```diff
commit 212931987f591ae70d4da594024c46eae3b105f5
Author: Dan Davison <dandavison7@gmail.com>
Date:   Mon Jan 3 16:13:55 2022 -0500

    WIP

diff --git a/src/delta.rs b/src/delta.rs
index 1e6fd84..7ba2997 100644
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -173,7 +173,7 @@ impl<'a> StateMachine<'a> {
                 || self.emit_line_unchanged()?;
         }
 
-        self.handle_pending_mode_line_with_diff_name();
+        self.handle_pending_mode_line_with_diff_name()?;
         self.painter.paint_buffered_minus_and_plus_lines();
         self.painter.emit()?;
         Ok(())
diff --git a/src/handlers/diff_header.rs b/src/handlers/diff_header.rs
index 61b2aa9..ea1c4dd 100644
--- a/src/handlers/diff_header.rs
+++ b/src/handlers/diff_header.rs
@@ -176,7 +176,7 @@ impl<'a> StateMachine<'a> {
         )
     }
 
-    pub fn handle_pending_mode_line_with_diff_name(&mut self) {
+    pub fn handle_pending_mode_line_with_diff_name(&mut self) -> std::io::Result<()> {
         if !self.mode_info.is_empty() {
             let format_label = |label: &str| {
                 if !label.is_empty() {
@@ -196,13 +196,15 @@ impl<'a> StateMachine<'a> {
             let name = get_repeated_file_path_from_diff_line(&self.diff_line)
                 .unwrap_or_else(|| "".to_string());
             let line = format!("{}{}", label, format_file(&name));
-            let _got = write_generic_diff_header_header_line(
+            write_generic_diff_header_header_line(
                 &line,
                 &line,
                 &mut self.painter,
                 &mut self.mode_info,
                 self.config,
-            );
+            )
+        } else {
+            Ok(())
         }
     }
 }
diff --git a/src/handlers/diff_header_diff.rs b/src/handlers/diff_header_diff.rs
index c2d7a86..fdd2d39 100644
--- a/src/handlers/diff_header_diff.rs
+++ b/src/handlers/diff_header_diff.rs
@@ -22,7 +22,7 @@ impl<'a> StateMachine<'a> {
             } else {
                 State::DiffHeader(DiffType::Unified)
             };
-        self.handle_pending_mode_line_with_diff_name();
+        self.handle_pending_mode_line_with_diff_name()?;
         self.handled_diff_header_header_line_file_pair = None;
         self.diff_line = self.line.clone();
         if !self.should_skip_line() {
```

src/handlers/diff_header.rs
WayneD: Looks like when I snagged that 9.. idiom from your code I failed to include your trailing comment about "old mode".len(). In any case, the strip_prefix() idiom is nicer.

src/handlers/diff_header.rs
WayneD: I tossed in a few /// comments for this & the plus & minus routines.

src/handlers/diff_header.rs
WayneD: Sounds good!

dandavison:
Excellent, thanks a lot for this work @WayneD!

PR #876, opened 2022-01-02
DeltaTest improvements

WayneD:
- Added .expect_contains(), .expect_raw_contains(), and .expect_contains_once() member functions to DeltaTestOutput.  These functions also output some helpful debug info when the assert fails.
- Separated .with_config_and_input() into .with_config() and (the tweaked) .with_input().
- Made .with_config() create a DeltaTest object (like .with() does).
- Renamed .with() as .with_args().
- Moved .explain_ansi() from DeltaTestOutput to DeltaTest, which must now be called prior to .with_input() since the latter stashes off both the raw_output & the processed output in the object.
- Changed .expect() and .expect_skip() to return Self so that they can continue a chain of multiple expect calls (e.g. a series of partial matches with different skip values).
- The processed output text can be accessed via `test_obj.output` (see also `test_obj.raw_output`).
- Renamed .expect_skip() to .expect_after_skip().
- Changed .expect() to start at the first line.
- Added .expect_after_header() that works like the old .expect().
- Renamed lines_match() to assert_lines_match() and made it match all lines (no skip number).
- Added assert_lines_match_after_skip() to work like the old lines_match() function, but with the .expect_after_skip() arg order. 
- Converted some old-style tests into DeltaTest style tests.

### DIFF
--- a/src/tests/integration_test_utils.rs
+++ b/src/tests/integration_test_utils.rs
@@ -12,6 +12,7 @@ use crate::cli;
 use crate::config;
 use crate::delta::delta;
 use crate::git_config::GitConfig;
+use crate::tests::test_utils;
 use crate::utils::process::tests::FakeParentArgs;
 
 pub fn make_options_from_args_and_git_config(
@@ -94,19 +95,25 @@ pub fn get_line_of_code_from_delta(
 //     line1"#;`  // line 2 etc.
 // ignore the first newline and compare the following `lines()` to those produced
 // by `have`, `skip`-ping the first few. The leading spaces of the first line
+// to indicate the last line in the list). The leading spaces of the first line
 // are stripped from every following line (and verified), unless the first line
 // marks the indentation level with `#indent_mark`.
-pub fn lines_match(expected: &str, have: &str, skip: Option<usize>) {
+pub fn assert_lines_match_after_skip(skip: usize, expected: &str, have: &str) {
     let mut exp = expected.lines().peekable();
-    assert!(exp.next() == Some(""), "first line must be empty");
-    let line1 = exp.peek().unwrap();
+    let mut line1 = exp.next().unwrap();
+    let allow_partial = line1 == "#partial";
+    assert!(
+        allow_partial || line1.is_empty(),
+        "first line must be empty or \"#partial\""
+    );
+    line1 = exp.peek().unwrap();
     let indentation = line1.find(|c| c != ' ').unwrap_or(0);
     let ignore_indent = &line1[indentation..] == "#indent_mark";
     if ignore_indent {
         let _indent_mark = exp.next();
     }
 
-    let mut it = have.lines().skip(skip.unwrap_or(0));
+    let mut it = have.lines().skip(skip);
 
     for (i, expected) in exp.enumerate() {
         if !ignore_indent {
@@ -119,23 +126,47 @@ pub fn lines_match(expected: &str, have: &str, skip: Option<usize>) {
         assert_eq!(
             &expected[indentation..],
             it.next().unwrap(),
-            "on line {} of input",
-            i + 1
+            "on line {} of input:\n{}",
+            i + 1,
+            delineated_string(have),
         );
     }
-    assert_eq!(it.next(), None, "more input than expected");
+    if !allow_partial {
+        assert_eq!(it.next(), None, "more input than expected");
+    }
+}
+
+pub fn assert_lines_match(expected: &str, have: &str) {
+    assert_lines_match_after_skip(0, expected, have)
+}
+
+pub fn delineated_string(txt: &str) -> String {
+    let top = "▼".repeat(100);
+    let btm = "▲".repeat(100);
+    let nl = "\n";
+    top + &nl + txt + &nl + &btm
 }
 
 pub struct DeltaTest {
     config: config::Config,
     calling_process: Option<String>,
+    explain_ansi_: bool,
 }
 
 impl DeltaTest {
-    pub fn with(args: &[&str]) -> Self {
+    pub fn with_args(args: &[&str]) -> Self {
         Self {
             config: make_config_from_args(args),
             calling_process: None,
+            explain_ansi_: false,
+        }
+    }
+
+    pub fn with_config(config: config::Config) -> Self {
+        Self {
+            config: config,
+            calling_process: None,
+            explain_ansi_: false,
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/tests/integration_test_utils.rs
```diff
@@ -185,32 +212,51 @@ impl DeltaTestOutput {
...
+        self.expect_skip(crate::config::HEADER_LEN, expected)
     }
 
-    pub fn expect(self, expected: &str) -> String {
-        self.expect_skip(crate::config::HEADER_LEN, expected)
+    pub fn expect_contains(self, expected: &str) -> Self {
+        let output_contains_text = self.output.contains(expected);
+        if !output_contains_text {
+            println!("Output does not contain \"{}\":\n{}", expected, self.output);
+            assert!(output_contains_text);
```
dandavison, line 233: I'm thinking that the printed message should go to stderr; the `assert!` macro actually supports that -- the [optional second argument](https://doc.rust-lang.org/std/macro.assert.html#custom-messages) is a string message which will be written to stderr.

src/tests/integration_test_utils.rs
```diff
@@ -126,16 +127,29 @@ pub fn lines_match(expected: &str, have: &str, skip: Option<usize>) {
     assert_eq!(it.next(), None, "more input than expected");
 }
 
-pub struct DeltaTest {
+pub struct DeltaTest<'a> {
     config: config::Config,
+    config_ref: Option<&'a config::Config>,
```
dandavison, line 132: I'm not so keen on this technique. I'm thinking that the simplest solution would be to change the signature of `with_config` so that it takes ownership of a `Config` struct:

```rust
    pub fn with_config(config: config::Config) -> Self {
```

and then, in any calling code that needs to re-use the same config twice, perhaps use an anonymous function to create the same config multiple times, e.g.

```rust
        let config =
            || make_config_from_args(&default_wrap_cfg_plus(&["--side-by-side", "--width", "55"]));
        {
            DeltaTest::with_config(config())
            ...
```

There are probably other solutions.

dandavison:
cc @th1000s who's the creator of `DeltaTest`.

src/tests/integration_test_utils.rs
WayneD: Nice! That simplifies things.

src/tests/integration_test_utils.rs
WayneD: I like that idea quite a bit.

src/tests/integration_test_utils.rs
```diff
@@ -152,65 +183,89 @@ impl DeltaTest {
...
-    pub fn expect_skip(self, skip: usize, expected: &str) -> String {
-        let processed = self.format_output();
-        lines_match(expected, &processed, Some(skip));
-        processed
+    pub fn expect(self, expected: &str) -> Self {
+        self.expect_skip(0, expected)
     }
 
-    pub fn expect(self, expected: &str) -> String {
+    pub fn expect_after_hdr(self, expected: &str) -> Self {
```
dandavison, line 237: Do you mind if we call this `expect_after_header()`? (I'm not keen on abbreviations in public APIs)

src/tests/integration_test_utils.rs
```diff
@@ -119,23 +126,47 @@ pub fn lines_match(expected: &str, have: &str, skip: Option<usize>) {
...
 }
 
 pub struct DeltaTest {
     config: config::Config,
     calling_process: Option<String>,
+    explain_ansi_: bool,
 }
 
 impl DeltaTest {
     pub fn with(args: &[&str]) -> Self {
```
dandavison, line 157: I feel like with the renaming and refactoring it would now make sense to call this `with_args()`.

src/tests/integration_test_utils.rs
```diff
@@ -119,23 +126,47 @@ pub fn lines_match(expected: &str, have: &str, skip: Option<usize>) {
...
         );
     }
-    assert_eq!(it.next(), None, "more input than expected");
+    if !allow_partial {
+        assert_eq!(it.next(), None, "more input than expected");
+    }
+}
+
+pub fn lines_match(expected: &str, have: &str) {
+    lines_match_skip(0, expected, have)
```
dandavison, line 140: Can we call these `assert_lines_match` and `assert_lines_match_after_skip`? Sorry, you appear to have opened a Pandora's box... :)

src/tests/integration_test_utils.rs
WayneD: Sounds fine to me.  I had even wondered if you might prefer that unabbreviated.

src/tests/integration_test_utils.rs
WayneD: OK -- that's nice and clear.

src/tests/integration_test_utils.rs
WayneD: These are nice and clear.  I had contemplated changing expect_skip() to expect_after(), but I was already worried about going overboard on tweaking someone else's code. 😄 Your suggestion of skip -> after_skip is even clearer, as is the assert prefix for the lines_match functions.  I'm also changing expect_skip() to expect_after_skip().

th1000s:
If calling `with_config(config: config::Config)` using the same config multiple times happens more often in the future then storing a `Cow<'a, &Config>` might help as well. Which is what you have been reimplementing manually anyhow :)

> tweaking someone else's code.

Oh please do, nothing helps more than someone who didn't write the code noticing that the names or logic are not as clear as they could be. Maybe also rename `set_cfg` to `set_config`?

dandavison:
This PR is great IMO Wayne, and I'm glad to hear that Thomas approves.

> Maybe also rename set_cfg to set_config

That SGTM. Do you want to make that final change and then I'll merge this?

dandavison:
Thanks again for this Wayne.

> If calling `with_config(config: config::Config)` using the same config multiple times happens more often in the future then storing a `Cow<'a, &Config>` might help as well.

My first thought was to try to use `Cow`, but there were errors: our `Config` is not `clone`-able, and yet my understanding is that `Cow` needs to be able to create an owned version of a data structure from a reference. I tentatively concluded that what we needed was not exactly `Cow` but an even simpler enum over `owned|reference` which supports borrowing, but I didn't think it through / investigate properly.

th1000s:
Indeed, `Cow<&Config>` is nonsense. Making `Config` cloneable required a few derives and a hack for `GitConfig`, see #884

PR #882, opened 2022-01-03
Create user manual

dandavison:


th1000s:
While you are at it, the beautiful satellite image of the amazon delta is much bigger than it has to be, and a png. At 1MB it is always the last to load if it is not cached.

dandavison:
> While you are at it, the beautiful satellite image of the amazon delta is much bigger than it has to be, and a png. At 1MB it is always the last to load if it is not cached.

Thanks, done cd47b21176b19eb700562dcb0999ef4b99c6a893.

PR #884, opened 2022-01-04
Make Config cloneable for DeltaTest, store as Cow<Config> there

th1000s:
Only cloneable when testing, the types git2::Config and
git2::Repository in GitConfig contain C pointers and can't
really be cloned.


WayneD:
Nice work!  My rust-novice eyes can follow most of that, and I think the GitConfig work-around seems quite reasonable for the tests.

dandavison:
Great!

PR #902, opened 2022-01-14
Add Criterion benchmarks (and librarify)

th1000s:
Benches require a library crate, so I had to split git-delta into a binary and library. But I have no idea what effects that has on e.g. cargo publishing or doc generation.

Also, the benches are quite noisy, `show` by 5% and `blame` by 10% so the `--noise-threshold` has to be quite high.

Commands (if `cargo-criterion` is installed, else `cargo bench`):

`cargo criterion --bench blame -- --sample-size 80 --measurement-time 30` 

`cargo criterion --bench blame -- --warm-up-time 3 --sample-size 80`

But if a smaller set than a full `blame` run should be benchmarked this can now be added easily.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -104,7 +104,14 @@ dependencies = [
  "lazy_static",
  "memchr",
  "regex-automata",
+ "serde",
 ]
+
+[[package]]
+name = "bumpalo"
+version = "3.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a4a45a46ab1f2412e53d3a0ade76ffad2025804294569aae387231a0cd6e0899"
 
 [[package]]
 name = "bytelines"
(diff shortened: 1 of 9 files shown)


dandavison:
I'm certainly in favour of any improvements to our benchmarking tools.

> I have no idea what effects that has on e.g. cargo publishing or doc generation.

Yes. I'm not familiar with how this works. When I tried refactoring as a library before (https://github.com/dandavison/delta/pull/398) I didn't merge it because I didn't want to publish a library crate that people might actually start using, when the API surface area wasn't at all something I'd thought about  or was prepared to maintain.

dandavison:
I've rebased this -- I'd like to go ahead and make delta have a basic library structure, even if it's just a very basic library initially.

As a next step, before merging this PR, would it make sense to continue restructuring  to minimize what's exposed in `lib`?

th1000s:
Continued in #1013

PR #914, opened 2022-01-16
Drop deprecated options

dandavison:
Fixes 906-912 but let's close them when this is released.


Iron-E:
Ever since this change I have been having an issue with `git`:

```log
❯ g diff main
error: Found argument '--plus-color' which wasn't expected, or isn't valid in this context

	Did you mean '--plus-style'?

	If you tried to supply `--plus-color` as a value rather than a flag, use `-- --plus-color`

USAGE:
    delta --plus-style <STYLE>

For more information try --help
```

Here is my `.gitconfig`:

```gitconfig
[core]
	pager = delta

[interactive]

[delta]
	navigate = true  # use n and N to move between diff sections

[merge]
	conflictstyle = diff3

[diff]
	colorMoved = default
```

(This is the same as the example from the README)

dandavison:
Hi @Iron-E, somehow you must be passing `--plus-color` to delta. Perhaps an environment variable, or an alias.

To investigate aliases, can you post the results of `which delta` and `type -a delta` and `which git` and `type -a git`?

To investigate relevant environment variables, in particular `GIT_PAGER`, could you run the command

```
env | grep -i pager
```

I can reproduce the error you see by setting the value of `GIT_PAGER` like this:

```
src/delta(master) GIT_PAGER='delta --plus-color=green' git show
error: Found argument '--plus-color' which wasn't expected, or isn't valid in this context
```

Another possibility is that git is using a config file different from, or in addition to, the one you showed.

Iron-E:
`$GIT_PAGER` was it. Thank you!

PR #915, opened 2022-01-16
Add clap cosmetic argument names

dandavison:


th1000s:
I for my part find these `SCREAMING_SNAKE_CASE` default specifiers visually jarring and yes, I know Pythons argparse and now clap use them by defaults. Its ok when it is just a short `N` or `NUM`, but these longer ones...  but I see you try to shorten them to reduce the `--for-bar <FOO_BAR>` cases.

dandavison:
> I for my part find these SCREAMING_SNAKE_CASE default specifiers visually jarring

Agreed. What do you think the best solution is?

1. **Disable them entirely**
  I'm not sure if it's possible. E.g. `value_name` can't be set to `None`
2. **Lower case**
  I worry that they're a bit hard to visually separate from the (important) option name
3. **Use a different color**
  It may be possible but I haven't found how yet.

I think it it at least a bit better now in this PR -- before they were screaming back the option name verbatim.

dandavison:
Incidentally, check out the concise help via `delta -h` in this branch -- I think it's going to be useful.

dandavison:
Merging this one which uses a smaller set of shorter names (e.g. `STYLE_STRING` many times over) but I'd be happy if we can identify further improvements.

th1000s:
> informal distinction

That's quite formal already and a good reason. I can also see that these are easier to visually separate.

I'd probably replace

    STYLE_STRING => STYLE
    FORMAT_STRING and other *FORMAT* => FMT
    NUMBER_OF_SPACES / WIDTH => N

> check out the concise help via `delta -h`

I couldn't help but notice this breaking the terminal width :)

        --merge-conflict-theirs-diff-header-decoration-style <MERGE_CONFLICT_THEIRS_DIFF_HEADER_DECORATION_STYLE>    Style string for the decoration of the header above the 'theirs' merge conflict diff [default: box]

dandavison:
> I'd probably replace

OK, I think I like those. People just have to understand once, but they are forever after condemned to look at them...

> I couldn't help but notice this breaking the terminal width :)

Right, I found I couldn't change that one for some reason. I reported a bug against clap: https://github.com/clap-rs/clap/issues/3300 but it's of course entirely possible I'm missing something.

PR #921, opened 2022-01-18
Update sysinfo version

GuillaumeGomez:
In the new `sysinfo` version, I turned `Pid` into a newtype.

### DIFF
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -5,6 +5,8 @@ use std::sync::{Arc, Condvar, Mutex, MutexGuard};
 use lazy_static::lazy_static;
 use sysinfo::{Pid, PidExt, Process, ProcessExt, ProcessRefreshKind, SystemExt};
 
+pub type DeltaPid = u32;
+
 #[derive(Clone, Debug, PartialEq)]
 pub enum CallingProcess {
     GitDiff(CommandLine),
@@ -283,7 +285,7 @@ impl ProcInfo {
 
 trait ProcActions {
     fn cmd(&self) -> &[String];
-    fn parent(&self) -> Option<Pid>;
+    fn parent(&self) -> Option<DeltaPid>;
     fn start_time(&self) -> u64;
 }
 
@@ -294,8 +296,8 @@ where
     fn cmd(&self) -> &[String] {
         ProcessExt::cmd(self)
     }
-    fn parent(&self) -> Option<Pid> {
-        ProcessExt::parent(self)
+    fn parent(&self) -> Option<DeltaPid> {
+        ProcessExt::parent(self).map(|p| p.as_u32())
     }
     fn start_time(&self) -> u64 {
         ProcessExt::start_time(self)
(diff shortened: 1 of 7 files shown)


dandavison:
Hi @GuillaumeGomez, thanks very much for doing this. It looks like there are some relevant test failures, but let us know if you'd like us to finalise the PR now.

GuillaumeGomez:
Ah indeed! I'll fix the PR as soon as possible then.

GuillaumeGomez:
The clippy lints seem buggy so just in case I put them into their own commit.

GuillaumeGomez:
CI is happy! \o/

dandavison:
Thanks @GuillaumeGomez. Would it also be an option for delta to work with native integer types, and only convert to the `sysinfo::Pid` type at the sites where delta is calling the library?

Perhaps another way of asking my question is -- I'm curious, can you explain why the `sysinfo::Pid` type needs to be used by clients of the sysinfo library, rather than being private to the library?

GuillaumeGomez:
> Thanks @GuillaumeGomez. Would it also be an option for delta to work with native integer types, and only convert to the `sysinfo::Pid` type at the sites where delta is calling the library?

`Pid` implements `From` where the conversion is from the native integer.

> Perhaps another way of asking my question is -- I'm curious, can you explain why the `sysinfo::Pid` type needs to be used by clients of the sysinfo library, rather than being private to the library?

The problem is that this native type isn't the same on all platforms, making it complicated to write one code to handle all platforms at once. With this wrapping, it's now possible. I added a common fallback with `from_u32` and `as_u32`, like that if you really need to work on the wrapped type, you can.

src/utils/process.rs
```diff
@@ -872,10 +891,15 @@ pub mod tests {
     #[test]
     fn test_process_blame_no_parent_found() {
         let two_trees = MockProcInfo::with(&[
-            (2, 100, "-shell", None),
-            (3, 100, "git blame src/main.rs", Some(2)),
-            (4, 100, "call_delta.sh", None),
-            (5, 100, "delta", Some(4)),
+            (Pid::from_u32(2), 100, "-shell", None),
+            (
+                Pid::from_u32(3),
```
th1000s, line 896: What about moving all the `Pid::from_u32(..)` calls into `MockProcInfo::with(..)`?

GuillaumeGomez: What do you think @dandavison ?

dandavison: Yes, I'd also prefer to continue to use native integers where possible, and only wrap as `sysinfo::Pid` where necessary -- i.e. close to the library call boundaries.

GuillaumeGomez: Fine by me! Will sent an update as soon as possible.

GuillaumeGomez:
I removed the `sysinfo::Pid` usage as much as possible, making the changes much smaller.

dandavison:
Thanks a lot @GuillaumeGomez!

PR #925, opened 2022-01-19
Make blame-timestamp-format actually useful

th1000s:


dandavison:
Should this work? (It doesn't seem to)

```
git -c delta.blame-timestamp-format='%Y-%m-%d' blame --date=format:'%Y-%m-%d' src/handlers/hunk_header.rs
```

dandavison:
Thanks for doing this, my first pass at blame handling was definitely a work-in-progress!

Do you think it could be appropriate to inspect the git config `[blame.date](https://git-scm.com/docs/git-config#Documentation/git-config.txt-blamedate)` key, and also look for `--date` among the calling process arguments, to automatically set the format, or is that going too far? (We have the machinery sitting there for both of course)

th1000s:
> '%Y-%m-%d'

That is because `DateTime` also requires a the time component of the input.

Parsing the exact blame key should work (is this in the git config data which delta has read already?). For even more customized date formats I've been thinking about parsing arbitrary strings like  GNUs `date --date="last week"` does. However I don't think these corner cases are worth it. Instead what about trying to parse Author + Date in the regex, and if that fails using both together `( (author) (date) | (.*) )`, dropping the `{timestamp}` field and using the fallback string in the `{author}` field (and ensuring each `blame-format` placeholder is unique, just as `{n}` must be in `blame-separator-format`.

If a user customized their blame timestamp this heavily, they might want to see it in delta as well.

th1000s:
Now accepts most timestamp formats, but `--date=human` or `-s` is still not accepted.

Issue #128, opened 2020-04-26
No output for empty commits


dandavison:
Given

```
commit 0f2025dfa9c75a25bde4e0431534e8efd7152ff0
Author: Dan Davison <dandavison7@gmail.com>
Date:   Sun Apr 26 17:03:26 2020 -0400

    Initial commit

diff --git a/file b/file
new file mode 100644
index 0000000..e69de29
```

delta outputs

```
commit 0f2025dfa9c75a25bde4e0431534e8efd7152ff0
Author: Dan Davison <dandavison7@gmail.com>
Date:   Sun Apr 26 17:03:26 2020 -0400

    Initial commit

```

whereas delta should additionally output

```
added: file
```

PR #945, opened 2022-02-06
Add support for irreversible-delete

wescande:
Add support for `git diff -D` (aka [irreversible delete](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---irreversible-delete))

The patch is adding a title for removed file, even when there is no file content in the diff output
Without the patch, there was no output related to the file at all

A image is better than a thousand word: wrong output on the left:x:, Better output on the right :heavy_check_mark: 
image

### DIFF
--- a/src/handlers/diff_header.rs
+++ b/src/handlers/diff_header.rs
@@ -14,9 +14,11 @@ const DIFF_PREFIXES: [&str; 6] = ["a/", "b/", "c/", "i/", "o/", "w/"];
 
 #[derive(Debug, PartialEq)]
 pub enum FileEvent {
+    Added,
     Change,
     Copy,
     Rename,
+    Removed,
     NoEvent,
 }
 
@@ -49,6 +51,25 @@ impl<'a> StateMachine<'a> {
         Ok(handled_line)
     }
 
+    fn should_write_generic_diff_header_header_line(&mut self) -> std::io::Result<bool> {
+        // In color_only mode, raw_line's structure shouldn't be changed.
+        // So it needs to avoid fn _handle_diff_header_header_line
+        // (it connects the plus_file and minus_file),
+        // and to call fn handle_generic_diff_header_header_line directly.
+        if self.config.color_only {
+            write_generic_diff_header_header_line(
+                &self.line,
+                &self.raw_line,
+                &mut self.painter,
+                &mut self.mode_info,
+                self.config,
+            )?;
+            Ok(true)
+        } else {
+            Ok(false)
+        }
+    }
+
     #[inline]
     fn test_diff_header_minus_line(&self) -> bool {
         (matches!(self.state, State::DiffHeader(_)) || self.source == Source::DiffUnified)
@@ -62,7 +83,6 @@ impl<'a> StateMachine<'a> {
         if !self.test_diff_header_minus_line() {
             return Ok(false);
         }
-        let mut handled_line = false;
 
         let (path_or_mode, file_event) =
             parse_diff_header_line(&self.line, self.source == Source::GitDiff);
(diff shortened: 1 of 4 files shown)


robinmoussu:
Nice, I just hit this limitation today and was going to report a bug!

dandavison:
Hi @wescande, thanks for this and I'm sorry I've been slow to review this one. It looks really clean and well-tested -- I'll look properly soon.

wescande:
I notice this allow to resolve #128 (even if it wasn't done for it initially)
I could either update this change by [removing the `#[ignore]`](https://github.com/dandavison/delta/blob/master/src/tests/test_example_diffs.rs#L20) or we can do it in a follow-up as you prefer.

src/handlers/diff_header.rs
```diff
@@ -153,7 +155,72 @@ impl<'a> StateMachine<'a> {
...
+    }
+
+    #[inline]
+    fn test_diff_header_file_operation_line(&self) -> bool {
+        (matches!(self.state, State::DiffHeader(_)) || self.source == Source::DiffUnified)
+            && (self.line.starts_with("deleted file mode ")
+                || self.line.starts_with("new file mode "))
+    }
+
+    /// Check for and handle the "deleted file ..."  line.
```
dandavison, line 170: Is `FileEvent::Added` used?

src/handlers/diff_header.rs
```diff
@@ -153,7 +155,72 @@ impl<'a> StateMachine<'a> {
...
+            FileEvent::Removed => {
+                // self.mode_info = mode_info;
+                self.minus_file = name;
+                self.plus_file = "/dev/null".into();
+                self.minus_file_event = FileEvent::Change;
+                self.plus_file_event = FileEvent::Change;
+                self.current_file_pair = Some((self.minus_file.clone(), self.plus_file.clone()));
+            }
+            FileEvent::Added => {
+                // self.mode_info = mode_info;
```
dandavison, line 197: (let's remove the commented lines)

src/handlers/diff_header.rs
```diff
@@ -153,7 +155,72 @@ impl<'a> StateMachine<'a> {
...
+        // So it needs to avoid fn _handle_diff_header_header_line
+        // (it connects the plus_file and minus_file),
+        // and to call fn handle_generic_diff_header_header_line directly.
+        if self.config.color_only {
+            write_generic_diff_header_header_line(
+                &self.line,
+                &self.raw_line,
+                &mut self.painter,
+                &mut self.mode_info,
+                self.config,
```
dandavison, line 217: (It would be nice to avoid the 3x repetition of this. Doesn't block this PR though.)

dandavison (changes requested):
Great! Couple of questions -- sorry again for being so slow this month.

dandavison:
> I notice this allow to resolve https://github.com/dandavison/delta/issues/128 (even if it wasn't done for it initially)
> I could either update this change by [removing the #[ignore]](https://github.com/dandavison/delta/blob/master/src/tests/test_example_diffs.rs#L20)

Awesome, yes please feel free to remove it in this PR if convenient.

src/handlers/diff_header.rs
wescande: `Added` and `Removed` are both with very small amount of use. There is the declaration then 2 locations for each:
in `parse_diff_header_line`:
```rust
        line if line.starts_with("new file mode ") => {
            (line[14..].to_string(), FileEvent::Added) // "new file mode ".len()
        }
        line if line.starts_with("deleted file mode ") => {
            (line[18..].to_string(), FileEvent::Removed) // "deleted file mode ".len()
        }
```

Then in `handle_diff_header_file_operation_line`:
```rust
        match file_event {
            FileEvent::Removed => [...],
            FileEvent::Added => [...],
            _ => (),
        }
```
The `Added` is what allow us to solve #128 

---

I hesitated to modify `get_file_change_description_from_file_paths` to use. This code:
```rust
            (minus_file, "/dev/null", _, _) => [...],
            ("/dev/null", plus_file, _, _) =>  [...],
```
would have become something like:
```rust
            (_, _, FileEvent::Removed, _) => [...],
            (_, _, _, FileEvent::Added) =>  [...],
```
And the complete match in which they are should have been refactored. to no longer take into account the filename but only the event

As it was my first contribution, I judged it was a bit too much of refactoring.
But now that I feel a bit more comfortable with your code if you want I can work on it.

dandavison:
Excellent, thanks very much for this work @wescande, and thanks for the extra refactor you did there. If you'd like to do any follow-up refactoring that would be very welcome of course.

PR #950, opened 2022-02-07
Fix unused imports and functions warnings on windows

th1000s:
Otherwise these warnings clutter the github diff display.

_super::super!_


dandavison:
Thank you!

(Work in progress PR https://github.com/dandavison/delta/pull/939 should finally fix the bugs with absolute paths e.g. when delta is invoked not from repo root dir. Possibly these bugs only affect `hyperlink` users. And then I should make all these tests pass on Windows.)

PR #967, opened 2022-02-15
Clarify language: "in front of" → "before"

waldyrious:
The expression "in front of" can have different interpretations when it refers to text. Some people think of the "front" of some text to be what's ahead of it in the direction it's facing — which would be at its right, for left-to-right text. That would have the opposite meaning of what is intended here, which is a prefix.

Using "before" makes the documentation less ambiguous.


dandavison:
Thanks @waldyrious (This ambiguity hadn't occurred to me previously.)

PR #972, opened 2022-02-17
Enable commit links for GitLab

jwarlander:
This change makes sure that `delta` recognizes GitLab remotes, and renders the proper commit links without the user having to do custom configuration.

### DIFF
--- a/src/git_config/git_config_entry.rs
+++ b/src/git_config/git_config_entry.rs
@@ -13,30 +13,21 @@ pub enum GitConfigEntry {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct GitRemoteRepo {
-    repo_slug: String,
-    commit_link_format: String,
+pub enum GitRemoteRepo {
+    GitHubRepo(String),
+    GitLabRepo(String),
 }
 
 impl GitRemoteRepo {
-    pub fn github_repo(repo_slug: String) -> Self {
-        Self {
-            repo_slug,
-            commit_link_format: "https://github.com/{repo_slug}/commit/{commit}".to_string(),
-        }
-    }
-
-    pub fn gitlab_repo(repo_slug: String) -> Self {
-        Self {
-            repo_slug,
-            commit_link_format: "https://gitlab.com/{repo_slug}/-/commit/{commit}".to_string(),
-        }
-    }
-
     pub fn format_commit_url(&self, commit: &str) -> String {
-        self.commit_link_format
-            .replace("{repo_slug}", &self.repo_slug)
-            .replace("{commit}", commit)
+        match self {
+            Self::GitHubRepo(repo_slug) => {
+                format!("https://github.com/{}/commit/{}", repo_slug, commit)
+            }
+            Self::GitLabRepo(repo_slug) => {
+                format!("https://gitlab.com/{}/-/commit/{}", repo_slug, commit)
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


jwarlander:
@dandavison, do let me know if there's anything I should fix / adjust on this one, or if you want an issue created related to the changes, etc -- happy to do the legwork, as long as I know what it is :relaxed:

src/git_config/git_config_entry.rs
```diff
@@ -13,8 +13,31 @@ pub enum GitConfigEntry {
...
+            repo_slug,
+            commit_link_format: "https://gitlab.com/{repo_slug}/-/commit/{commit}".to_string(),
+        }
+    }
+
+    pub fn format_commit_url(&self, commit: &str) -> String {
+        self.commit_link_format
+            .replace("{repo_slug}", &self.repo_slug)
+            .replace("{commit}", commit)
+    }
```
dandavison, line 40: Hi @jwarlander, thanks very much for this PR!

Sorry for being slow. I'd been caught between wanting to leave a comment to keep things moving and wanting to find time not to leave a lazy comment, if you know what I mean.

My one question about this is whether perhaps the implementation isn't quite the idiomatic way to do this in Rust. 
 Since a repo cannot simultaneously be a GitHub repo and a GitLab repo, I was thinking that the way we'd implement this would be to add a `GitLabRepo` variant to the `GitRemoteRepo` enum. Maybe you've already thought about this while you were implementing this PR. What do you think, could it maybe be something like the code fragment below?

```rust
#[derive(Clone, Debug, PartialEq)]
pub enum GitRemoteRepo {
    GitHubRepo(String),
    GitLabRepo(String),
}

impl GitRemoteRepo {
    pub fn format_commit_url(&self, commit: &str) -> String {
        match self {
            Self::GitHubRepo(repo_slug) => {
                format!("https://github.com/{}/commit/{}", repo_slug, commit)
            }
            Self::GitLabRepo(repo_slug) => {
                format!("https://gitlab.com/{}/-/commit/{}", repo_slug, commit)
            }
        }
    }
}

fn format_commit_line_captures_with_osc8_commit_hyperlink(
    captures: &Captures,
    repo: &GitRemoteRepo,
) -> String {
    let commit = captures.get(2).unwrap().as_str();
    format!(
        "{prefix}{osc}8;;{url}{st}{commit}{osc}8;;{st}{suffix}",
        url = repo.format_commit_url(commit),
        commit = commit,
        prefix = captures.get(1).map(|m| m.as_str()).unwrap_or(""),
        suffix = captures.get(3).unwrap().as_str(),
        osc = "\x1b]",
        st = "\x1b\\"
    )
}

```

There might be better ways to do this; for example it would be nice if `GitHubRepo(String)` were more self-documenting regarding what the string data is.

jwarlander: No worries, I definitely know how it goes, and I imagine you've got plenty to do besides working on this project :sweat_smile: 

Anyhow, yeah, that works of course!

It just didn't occur to me at the time of writing that you could actually have an impl for an enum type -- haven't had the opportunity to work with those a lot yet, and they're significantly more capable in Rust than other languages I've used :sunglasses:

jwarlander: It would definitely be more self-documenting with the variants as structs, even if *slightly* more verbose; have a look at my most recent update to see how it would look.

dandavison:
Excellent, that looks perfect and I agree with you -- the structs are better. I've tested quickly on the iTerm2 repo and hyperlinks in git log and git blame link to the gitlab remote correctly. I'll merge this when the tests pass. Thanks a lot!

src/git_config/git_config_entry.rs
```diff
@@ -32,19 +46,45 @@ lazy_static! {
...
+            })
+        } else if let Some(caps) = GITLAB_REMOTE_URL.captures(s) {
+            Ok(Self::GitLabRepo {
+                repo_slug: format!(
+                    "{user}{groups}/{repo}",
+                    user = caps.get(1).unwrap().as_str(),
+                    groups = caps.get(2).map(|x| x.as_str()).unwrap_or_default(),
+                    repo = caps.get(3).unwrap().as_str()
+                ),
+            })
```
dandavison, line 85: If you felt like it, it might be nice to separate these into separate functions and have `from_str` implemented as "return the result from the first parser function that succeeds".

jwarlander: My first thought, before the weekend, was to see if it could be done using serde.. But that seemed like total overkill 😅

It would be nice with something else in the long run.. Like, each repo type just knowing how its URL is represented, then registering itself so that the parser can try them all.

That could enable specifying full-fledged new repo types in a config file, say, for those who have hosted GitHub or GitLab instances, etc.

dandavison:
Merged. Thanks a lot for this work @jwarlander!

Issue #974, opened 2022-02-19
🐛  `git grep` output parsed, printed incorrectly for .properties files with hyphen in path


jdpopkin:
When I `git grep` for something that's found in a `.properties` file whose directory name contains a hyphen, I get bad results. The highlighting is wrong on the matching line, and the filename itself is incorrect - the hyphen within the path gets replaced with a colon.

To repro:

```
mkdir /tmp/delta-test
cd /tmp/delta-test
mkdir src
mkdir src-ext
echo 'foo' > src{,-ext}/bar.properties
git init
git add .
git commit -m 'Add properties files'
git grep 'oo'
```

Here's what I get:

<img width="335" alt="Screen Shot 2022-02-19 at 11 02 15 AM" src="https://user-images.githubusercontent.com/3980541/154808961-086722df-8cd7-4dde-a5ef-479a87e053f2.png">

Note that the first line has incorrect highlighting (the filename-highlighting cuts off where the `-` should be and the match isn't highlighted at all) and the hyphen in `src-ext` is actually replaced by a `:`.

`git --no-pager grep 'oo'` results:

```
src-ext/bar.properties:1:foo
src/bar.properties:1:foo
```

I looked into this and I have a fix (sort of). I'll post a PR soon.

--
- [X] Please include the raw text output from git, so that we can reproduce the problem.
      (You can use `git --no-pager` to produce the raw text output.)
- [X] A screenshot of Delta's output is often helpful also.

Thanks for filing a Delta bug report!

PR #975, opened 2022-02-19
Fix parsing for .properties files with `-` in path

jdpopkin:
This adds a test that repros the problem in #974 and fixes it, at least for this specific filetype. The fix just involves changing the maximum file extension length for one of the regexes. The same problem would still happen to files with longer extensions.

There's probably a small tradeoff between performance and file extension length, and there will probably always be someone out there who wants to use even longer file extensions. But I think `.properties` is worth supporting - it's a common enough filetype to have its own [Wikipedia article](https://en.wikipedia.org/wiki/.properties).


dandavison:
Thanks @jdpopkin! I agree this should be fixed and your solution LGTM. I'll merge when the tests pass.

Issue #794, opened 2021-11-24
🐛 Colored output from standard grep is not parsed correctly


dandavison:
Delta officially supports two ways of highlighting grep output:

1. `git grep` with delta configured as git's pager either via `core.pager = delta` or `pager.grep = delta`
2. `rg --json ... | delta`

However, delta does try to highlight other grep output that is piped to it. This issue shows it failing to handle output from standard grep (`grep (BSD grep) 2.5.1-FreeBSD`)

<table><tr><td><img width=600px src="https://user-images.githubusercontent.com/52205/143273008-648b3fe7-73e6-4ff3-b9e8-5aadaf69314f.png" alt="image" /></td></tr></table>

<table><tr><td><img width=600px src="https://user-images.githubusercontent.com/52205/143272159-29254e0a-cf89-4e71-9628-79a48aa6c969.png" alt="image" /></td></tr></table>

cf

Similarly

<table><tr><td><img width=600px src="https://user-images.githubusercontent.com/52205/143273924-be49b509-cecf-4a31-ae17-5a7e59a3a152.png" alt="image" /></td></tr></table>

dandavison:
It looks like `parse_style_sections` is failing to parse the ANSI escape sequences emitted by BSD grep and GNU grep. For example,

```
delta(794-ansi-parse-bug *) grep --color=always -Hn s src/**/*.rs | head | bat -A   
src/align.rs:1:u␛[01;31m␛[Ks␛[m␛[Ke·␛[01;31m␛[Ks␛[m␛[Ktd::cmp::max;␊
src/align.rs:2:u␛[01;31m␛[Ks␛[m␛[Ke·␛[01;31m␛[Ks␛[m␛[Ktd::collection␛[01;31m␛[Ks␛[m␛[K::VecDeque;␊
src/align.rs:4:con␛[01;31m␛[Ks␛[m␛[Kt·SUBSTITUTION_COST:·u␛[01;31m␛[Ks␛[m␛[Kize·=·1;␊
src/align.rs:5:con␛[01;31m␛[Ks␛[m␛[Kt·DELETION_COST:·u␛[01;31m␛[Ks␛[m␛[Kize·=·1;␊
src/align.rs:6:con␛[01;31m␛[Ks␛[m␛[Kt·INSERTION_COST:·u␛[01;31m␛[Ks␛[m␛[Kize·=·1;␊
src/align.rs:11:····Sub␛[01;31m␛[Ks␛[m␛[Ktitution,␊
[...]

PR #976, opened 2022-02-20
Support CSI sequences other than SGR

yoichi:
Handle CSI sequences other than SGR (Refs https://en.wikipedia.org/wiki/ANSI_escape_code).
It solves https://github.com/dandavison/delta/issues/794


dandavison:
This is fantastic @yoichi, thanks very much! I hadn't understood what the problem was.

PR #977, opened 2022-02-20
Fix performance when many processes

### LINKED ISSUE #973: 🐛 delta `0.12.0` is slower than `0.11.3`
Hi!
I see something similar to #824 (which is fixed in `0.11.2`). I will try to reproduce it in docker and post the output.

dandavison:
cc @th1000s @infokiller fixes #973 

Let's discuss in #973

### DIFF
--- a/etc/examples/973-processes-performance-bug
+++ b/etc/examples/973-processes-performance-bug
@@ -7,7 +7,11 @@ build_oci_img() {
   RUN curl -fsSL 'https://sh.rustup.rs' | sh -s -- -y --profile minimal
   RUN /usr/local/bin/cargo install --version 0.11.3 git-delta && cp /usr/local/bin/delta /usr/local/bin/delta-0.11.3
   RUN /usr/local/bin/cargo install --version 0.12.0 git-delta && cp /usr/local/bin/delta /usr/local/bin/delta-0.12.0
-  WORKDIR /repo
+  RUN echo "I've forgotten how to invalidate layer cache entries properly: 1" > /dev/null
+  RUN git clone https://github.com/dandavison/delta.git
+  WORKDIR delta
+  RUN git checkout 973-processes-performance
+  RUN cargo build --release && cp ./target/release/delta /usr/local/bin/delta-dev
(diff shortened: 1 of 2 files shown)


etc/examples/973-processes-performance-bug
```diff
@@ -0,0 +1,59 @@
+build_oci_img() {
+  # RUN apk add --no-cache bash git
+  docker build "$@" - <<'EOF'
+  FROM ubuntu
+  RUN apt update && apt install -y bc curl gcc git
```
dandavison, line 5: @infokiller totally tangential but I changed this to ubuntu to make it work for me -- do you know why your archlinux version didn't work for me (on MacOS)? Have I got a stale archlinux image cached?

```
 => CACHED [2/7] RUN pacman -Sy > /dev/null                                                                                                                                                                              0.0s
 => CACHED [3/7] RUN pacman -S --noconfirm --needed coreutils curl gcc bc git > /dev/null                                                                                                                                0.0s
 => ERROR [4/7] RUN curl -fsSL 'https://sh.rustup.rs' | sh -s -- -y --profile minimal                                                                                                                                    1.3s
------
 > [4/7] RUN curl -fsSL 'https://sh.rustup.rs' | sh -s -- -y --profile minimal:
#6 0.354 info: downloading installer
#6 1.244 /tmp/tmp.VVpvUf3oyz/rustup-init: /usr/lib/libc.so.6: version `GLIBC_2.34' not found (required by /usr/lib/libgcc_s.so.1)
```

infokiller: I have no idea but a stale image seems plausible, does anything change when running `docker pull archlinux` before the script?
Anyway, obuntu is fine and probably a better choice because of its popularity. I mostly use alpine and debian/ubuntu in docker, I think I was just being lazy using archlinux...

dandavison: OK thanks, I haven't managed to fix it yet by deleting cached image layers or by `docker pull archlinux`, but will post here if I do.

Issue #929, opened 2022-01-24
Should `delta` follow symlinks?


rrbutani:
(filing this as a feature request instead of a bug report because I'm not sure if this is the intended behavior or not)

---

Currently, when using `delta` as a diff tool with paths that are symlinks, the literal contents of the symlink (as in, the path of the file that's being pointed to) is used in the diff instead of the contents of the file that's being pointed to.

For example:
```bash
echo "hello" > foo
ln -s foo bar
delta -s foo bar
```
yields:
```
❯ delta foo bar -s

removed: foo
─────────────────────────────────────────────────────────────────────────────────

───┐
0: │
───┘
│ 1  │hello                              │    │

added: bar
─────────────────────────────────────────────────────────────────────────────────

───┐
1: │
───┘
│    │                                   │ 1  │foo
\ No newline at end of file
```
instead of no output (as `diff`, for example, does).

This is because of `git diff --no-index`'s behavior with respect to symlinks. As far as I know there isn't any user-facing way to change this behavior ([1](https://public-inbox.org/git/xmqqk27m1ixf.fsf@gitster.mtv.corp.google.com/T/), [2](https://git.vger.kernel.narkive.com/q1CpMpoI/grep-doesn-t-follow-symbolic-link); there was a [patch](https://public-inbox.org/git/20161111201958.2175-1-dennis@kaarsemaker.net/) that changed the default behavior to be to follow symlinks but afaict it was never merged).

---

Is this intended behavior for `delta`?

Adding calls to [`fs::canonicalize`](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) to [these two paths](https://github.com/dandavison/delta/blob/8db8e908432539543341109f94a19dc2c6f4af7f/src/subcommands/diff.rs#L27-L28) would change `delta`'s behavior w.r.t. symlinks to match `diff` and other tools but would also mean there isn't a way to actually diff the literal contents of symlinks (`diff` has `--no-dereference` for example). I'm not sure if this matters.

(Also right now `delta` just lets `git diff --no-index` error when a file that's specified doesn't exist; if we call `canonicalize` in `delta` we'd need to handle the possibility of broken symlinks. I assume just printing an error message matching that of `git diff` (i.e. `error: Could not access: ...`) for symmetry is acceptable)

If this is deemed a reasonable change, I'm very happy to make a PR with the above change and some tests.

dandavison:
[...]

PR #978, opened 2022-02-21
Fall back to plain diff when process substitution is used

th1000s:
When detecting input generated by `delta <(echo foo) <(echo bar)`
fall back to plain `diff` instead of `git diff --no-index`.

This does not respect various git settings anymore (the original reason
to switch from `diff` to `git diff`), but is better than just showing
the names of the temporary files.

### DIFF
--- a/src/subcommands/diff.rs
+++ b/src/subcommands/diff.rs
@@ -1,5 +1,5 @@
 use std::io::{ErrorKind, Write};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process;
 
 use bytelines::ByteLinesReader;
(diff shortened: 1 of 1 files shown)


dandavison:
Hi Thomas, do you think there's a viable alternative involving working with the file descriptor(s) and/or creating temporary files, along the lines of the discussion here: https://github.com/dandavison/delta/issues/929#issuecomment-1020597339 ?

If possible I'd really like to keep the behavior consistent and use `git diff`

th1000s:
That approach looks great and all the caveats I could think of are already being discussed, I like the piping-to-stdin so only one tmp-file has to be created.

But in the meantime, isn't perfect the enemy of good (enough)? :)

dandavison:
> But in the meantime, isn't perfect the enemy of good (enough)? :)

Yes, you're right! I just wanted to check that you're on board with the other solution. (I'm not aware that it's being actively worked on.)

PR #984, opened 2022-02-25
Calling process: Query more parents, disable full scan

th1000s:
Also look at grandparent process when determining the caller

Disable full process scans: The env var `DELTA_CALLING_PROCESS_QUERY_ALL` re-enables this
last resort method. However usually this is just an expensive scan which doesn't find the caller anyhow.

### DIFF
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -528,8 +528,20 @@ where
 
     match info.find_sibling_in_refreshed_processes(my_pid, &extract_args) {
         None => {
-            info.refresh_processes();
-            info.find_sibling_in_refreshed_processes(my_pid, &extract_args)
+            #[cfg(not(target_os = "linux"))]
+            let full_scan = true;
+
+            // The full scan is expensive on Linux and rarely successful, so disable it by default.
+            #[cfg(target_os = "linux")]
+            let full_scan = std::env::var("DELTA_CALLING_PROCESS_QUERY_ALL")
+                .map_or(false, |v| !["0", "false", "no"].iter().any(|&n| n == v));
+
+            if full_scan {
+                info.refresh_processes();
+                info.find_sibling_in_refreshed_processes(my_pid, &extract_args)
+            } else {
+                None
+            }
         }
         some => some,
     }
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -298,6 +298,7 @@ impl ProcInfo {
 trait ProcActions {
     fn cmd(&self) -> &[String];
     fn parent(&self) -> Option<DeltaPid>;
+    fn pid(&self) -> DeltaPid;
     fn start_time(&self) -> u64;
 }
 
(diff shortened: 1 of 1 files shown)


th1000s:
The full process scan could still be done on macOS and Windows, I don't think listing all processes requires thousands of syscalls there.

dandavison:
I've updated @infokiller's docker script to benchmark this branch against 0.11.3 and 0.12.0 (see #977) and the results look good for this branch:

```
Benchmarking version: 0.11.3
delta 0.11.3
Running command "git diff --no-index a b | delta-0.11.3 >| delta-0.11.3-result" 10 times
61.40 ms per run (614 ms total)
Benchmarking version: 0.12.0
delta 0.12.0
Running command "git diff --no-index a b | delta-0.12.0 >| delta-0.12.0-result" 10 times
59.70 ms per run (597 ms total)
Benchmarking version: dev
delta 0.12.0
Running command "git diff --no-index a b | delta-dev >| delta-dev-result" 10 times
58.30 ms per run (583 ms total)
Spinning up 5000 processes
Process counts:
      1 bash
      1 ps
   5000 sleep
      1 sort
      1 uniq
Benchmarking version: 0.11.3
delta 0.11.3
Running command "git diff --no-index a b | delta-0.11.3 >| delta-0.11.3-result" 10 times
155.20 ms per run (1552 ms total)
Benchmarking version: 0.12.0
delta 0.12.0
Running command "git diff --no-index a b | delta-0.12.0 >| delta-0.12.0-result" 10 times
345.90 ms per run (3459 ms total)
Benchmarking version: dev
delta 0.12.0
Running command "git diff --no-index a b | delta-dev >| delta-dev-result" 10 times
68.80 ms per run (688 ms total)
```

src/utils/process.rs
```diff
@@ -498,18 +514,29 @@ where
...
+        }
     }
 
     match info.find_sibling_in_refreshed_processes(my_pid, &extract_args) {
         None => {
-            info.refresh_processes();
-            info.find_sibling_in_refreshed_processes(my_pid, &extract_args)
+            // The full scan is expensive and rarely successful, so disable it by default
+            if std::env::vars()
+                .any(|(k, v)| k == "DELTA_CALLING_PROCESS_QUERY_ALL" && v != "0" && v != "false")
```
dandavison, line 533: Better to use [`std::env::var`](https://doc.rust-lang.org/std/env/fn.var.html) here?

dandavison:
Looks great @th1000s! One nitpick (and one clippy complaint).

dandavison:
Sorry the clippy error was nothing to do with this PR; I've rebased on master to get rid of it.

I am curious whether you prefer `std::env::vars` over the "get key" API `std::env::var(key)` (two places in `process.rs`)

dandavison:
Also, this PR is in Draft status: could you let me know whether it's ready to go out?

(I'm quite keen to make a release soon with the new bat assets, and would like to include this)

src/utils/process.rs
th1000s: I just never bothered to look up the equivalent for the `var()` call, which is `.map_or(false, ...)`, until now :)

And I think the underlying `getenv()` has to walk the environ list just the same.

th1000s:
Because I can't test the behavior elsewhere I have only disabled the full scan on Linux now. Other OSes probably have faster syscalls for this query anyhow.

> clippy

The 1.58 clippy wants to undo what the (I presume) 1.59 one suggested!

dandavison:
Thanks very much @th1000s!

PR #991, opened 2022-02-26
Include information about GIT_PAGER in manual

katylava:
under configuration instructions

`delta` is so awesome that I spent hours trying to get it to work, and eventually found this little thorn in my env. Not sure this is where you'd want to mention it, but I think it's worth mentioning somehow, somewhere.

### DIFF
--- a/manual/src/configuration.md
+++ b/manual/src/configuration.md
@@ -31,6 +31,8 @@ The most convenient way to configure delta is with a `[delta]` section in `~/.gi
 
 </sub>
 
+If you have the `GIT_PAGER` environment variable set, you'll need to unset it.
+
 Use `delta --help` to see all the available options.
 
 Note that delta style argument values in ~/.gitconfig should be in double quotes, like `--minus-style="syntax #340001"`. For theme names and other values, do not use quotes as they will be passed on to delta, like `theme = Monokai Extended`.


manual/src/configuration.md
```diff
@@ -31,6 +31,8 @@ The most convenient way to configure delta is with a `[delta]` section in `~/.gi
 
 </sub>
 
+If you have the `GIT_PAGER` environment variable set, you'll need to unset it.
```
dandavison, line 34: Hi @katylava, good call! This is definitely worth mentioning. Shall we maybe tweak that to say that they can set `GIT_PAGER=delta`, or unset it?

katylava: I think saying more would require more explanation as well, since the documentation on this page is just to get people started, it should probably be kept short. Like I think if you say they can set it, you'll also have to explain that it overrides what is in .gitconfig under core.pager. And then explain why they might want it in one place over the other.

dandavison: Yes, you're right. OK, thanks a lot -- I'll merge this now and maybe explain more later.

(It might be nice at some point to make a `delta doctor` command that, among other things, would check for relevant environment variables and warn if they look like not what the user wanted.)

Issue #754, opened 2021-10-29
🐛 CRLF line endings are stripped from diff


plgruener:
I'm trying to show CRLF line endings in my diffs (on Linux, to quickly spot if someone committed a file with CRLF instead of LF). Less has the option `-u/--underline-special` to do just that, so I could just use that as pager for delta. However, delta _always_ seems to strip the carriage return byte from diffs, see:

![delta-CRLF](https://user-images.githubusercontent.com/5796100/139351646-83bb2209-6f72-46f6-94a9-b5629a865d8b.png)

Here is a hexdump of both diffs. You can see the CR-bytes (0x0d) are missing.
![delta-CRLF-hexdump2](https://user-images.githubusercontent.com/5796100/139352663-957e4c70-1cb2-45ec-9817-b4093860a2ab.png)

Even if delta strips the ANSI color escape codes (which is turned off here with `--raw`), it should never just delete any other character from the output. Let the pager decide to show those control chars or not.

dandavison:
Thanks @plgruener. Do you have time to look at https://github.com/dandavison/delta/pull/664 and help me decide what the correct course of action is? In that PR, code was added that deliberately strips trailing CR. cc @norisio in case you have time to advise also.

yoichi:
https://github.com/dandavison/delta/pull/664 is not the only cause. byte_lines may also strip CR.
The correct course will be

* prevent byte_lines removing CR
* strip CR only if delta writes to console

dandavison:
> This (=CR erases the line) does not happen with the default pager. Why does this happen with delta?

I think this is because:

Suppose we have a line ending with CR LF.

Git changes that to CR ANSI LF.

With delta, bytelines strips the LF but leaves the CR (due to ANSI escape added by git). So now the line ends with CR ANSI *but no LF*. I think this causes the line to be overwritten by the terminal emulator.

Whereas with the usual pager the line ends CR ANSI LF.



, with delta, byte lines h

yoichi:
> This (=CR erases the line) does not happen with the default pager. Why does this happen with delta?

I found right_fill_background_color's output overwrites the text before CR.
Without interactive.diffFilter, there is only reset (ESC[m) after CR, the text before CR remains intact.

> I've tried this by applying sample code in https://rosettacode.org/wiki/Check_output_device_is_a_terminal#Rust
> But it didn't work with git add -p.

`git add -p` (git-add--interactive.perl) sends diff to pipe (not directly to the terminal) then print it.
[...]

Issue #188, opened 2020-05-13
Add support for arbitrary encoding 🚀


Alphare:
Hi,
I've quickly looked at the project and it looks awesome. Also thanks for supporting Mercurial. :)

The project uses `String` for handling hunks, which forces the data to be UTF8, or since https://github.com/dandavison/delta/commit/9d1aafee2465b3b89de09ae72e1b8b8fbda4a981, display garbage.

Some projects historically did not start with UTF8, either because they predated the standard or for other reasons which makes it impossible for them to use VCS or VCS-related software that does not support handling arbitrary bytes without completely losing their (often decades-long) history.

Have you considered it? Is it a wontfix? 

Thanks

dandavison:
Hi @Alphare thanks for getting in touch. Absolutely, let's fix this. I hasdn't thought about this previously but based on your message I'm thinking that we need to support multiple different encodings, in different files/commits within a repo. How does the following sound? cc @fgsch

- We add a new command line option `--encoding`. For example, a user could pass `--encoding=utf-8,another-encoding`. (I.e. singular name "encoding", but supporting a comma-separated list; I am not sure if the command line app libraries in rust support passing an argument multiple times and assembling a list of values).
- Delta will try each encoding in turn, until one succeeds.
- If none succeed, then delta will do a lossy decoding, using the first encoding in the supplied list, producing output that will look like `Hello �World` (or just a lot of `�`).
- The default value of `--encoding` will be `utf-8`.

Recently opened issues #187 and  #150 led me to put in the quick fix 9d1aafee2465b3b89de09ae72e1b8b8fbda4a981 to avoid crashing in the face of invalid utf-8, but I was about to modify that in any case, so I think this work fits in. I also think we can do this without impacting performance on the common case (actually, it [might](https://github.com/whitfin/bytelines) improve it).

fgsch:
I'd like to add that this is not necessarily historical. Utf-8 (or any encoding really) makes sense for text, but there are cases where you want to have arbitrary bytes. 

Adding something like `--encoding` makes sense to me. 
Would the retry be at line level or entire output?

dandavison:
The retry would be at line level (delta only reads a limited number of lines into memory at a time -- basically one  diff hunk).

I haven't looked into this properly yet: for entirely non-textual files, I was thinking that git would often output text like `binary files a/mybinary b/mybinary differ", in which case there will be no issue (and there's a unit test that delta handles that output from git appropriately).

[...]

PR #1000, opened 2022-03-03
New option caret-encode to avoid the display issue with git add -p

yoichi:
# Summary

Add new option `--caret-encode` to replace control chacters in output with caret notation.
It introduces another solution for https://github.com/dandavison/delta/pull/664, and fixes https://github.com/dandavison/delta/issues/754

To avoid https://github.com/dandavison/delta/pull/664, users are required to use following configuration:

```
[interactive]
    diffFilter = delta --color-only --caret-encode
```

I chose the option name by referring to ascii::caret_encode.

# Limitations

`git diff --color=never` cause recurrence of https://github.com/dandavison/delta/issues/754
because bytelines removes CR just before LF.
I think we should deal with it by extending bytelines or not using bytelines.


yoichi:
I found that `git diff --color=always` does not guarantee the escape sequence between CR and LF on each line.
So this It wasn't enough for #754.

> I think we should deal with it by extending bytelines or not using bytelines.

is needed...

dandavison:
Ho @yoichi, I think patching bytelines, or not using bytelines, are good ideas. The bytelines author was very helpful when I contacted them about an issue.

Not using it does make sense though, since it's not a mainstream library. I think the reason I started using it was to support non-utf8 encoded input. I'll find the relevant issues.

dandavison:
See https://github.com/dandavison/delta/issues/188 for encoding discussion. I think I never finished the work discussed.

dandavison:
bytelines allows us to output lossy decoding. Before bytelines I was doing this, which meant an error message would appear in the middle of user's output: https://github.com/dandavison/delta/commit/9d1aafee2465b3b89de09ae72e1b8b8fbda4a981

We should look how other Rust projects deal with non utf8 input; do they not use `stdin.lock().lines()`?

norisio:
There seems need to a new issue (or #754 probably) to discuss how to preserve CR.

dandavison:
I have to say my instinct is that this doesn't feel like the right solution. But I don't think I have the problem clear in my mind so I'm sorry to say something vague like that. I don't really want to make delta's instructions for `git add -p` more complicated (that's my fault because the name `--color-only` was never a very good name!)

What is this there about this problem that is specific to `git add -p`?

yoichi:
The problem specfic to `git add -p` is that if we preserve CRs (and other control characters) and show them,
we must translate them to something printable (e.g. caret notation), since bare control characters will be handled
by the terminal.

After I've read the following comment from @dandavison on #754

> I do think it's worth bearing in mind that delta's output is for humans to look at, not for machines to parse

then I came up with an idea: we can choose delta's bahavior that it always translate control characters in its output
to printables, without introducing new option, and not limited to `git add -p`.

yoichi:
In 4ac5035, I've removed the option caret-encode added in this PR.

Note 
* The replacement process can be placed earlier, but it makes the input caret notation and control characters indistinguishable.
* This change allows us to use reverse video to distinguish between them.

Issue #949, opened 2022-02-07
🚀 GitHub-like style


daniel-liuzzi:
<details>
  <summary>My config</summary>

  ```ini
  file-style = "#ffffff"
  file-decoration-style = "#333333" ul box
  hunk-header-decoration-style = none
  hyperlinks = true
  line-numbers = true
  line-numbers-minus-style = "#993333"
  line-numbers-zero-style = "#666666"
  line-numbers-plus-style = "#336633"
  line-numbers-left-format = "{nm:>3} ⋮ "
  line-numbers-left-style = "#333333"
  line-numbers-right-format = "{np:>3} │ "
  line-numbers-right-style = "#333333"
  max-line-length = 0
  minus-style = syntax "#330000"
  navigate = true
  plus-style = syntax "#003300"
  minus-emph-style = syntax "#660000"
  plus-emph-style = syntax "#006600"
  relative-paths = true
  syntax-theme = Dracula
  ```
</details>

I'm wondering if it's possible to have Delta render a diff similar to how GitHub does it.

For instance, this is what 37dd3391844efefbfe59acbbf3ad661b497ed9ce looks like on GitHub:

![image](https://user-images.githubusercontent.com/143884/152861797-6e906e87-5503-4863-b51e-1dd2a66cc7b9.png)

And this is how it looks on Delta:

![image](https://user-images.githubusercontent.com/143884/152861994-fddeb5de-b4c6-4cb6-9305-25341c3bcac5.png)

It's _almost_ there, except for the background color not covering line numbers at left.

I tried this, but it didn't quite work:

```ini
line-numbers-minus-style = syntax "#330000"
line-numbers-plus-style = syntax "#336633"
```

![image](https://user-images.githubusercontent.com/143884/152862811-c43d4231-e0af-49d1-a82b-3c22efa4dced.png)

Notice how it paints _columns_ rather than lines. Also, the background covers line numbers but not the padding around them, which looks strange. 

Is this possible? I went through the `-style` options in `delta --help` but didn't find anything I'm not already using. On a final note, being able to control line number foreground/background color independently from the diff context would be the icing on the cake!

Thanks in advance for any help and for this awesome project!

dandavison:
Hi @daniel-liuzzi, thanks very much for this and sorry to be slow.

So let's take the line number background color issue first. I definitely sympathize with what you're aiming for here. It's slightly complicated: I think there might be a bug involved, and also some limitations of what's possible with current config. 

[...]

PR #1009, opened 2022-03-14
Row-oriented line number styles

dandavison:
See #949


mkonig:
Hi

Upfront: I have no clue about rust and my comment is to be interpreted with this in mind. Thanks.

I tried it with the current master branch (01f92de6df52457aceb68a5c89b2bb440a6b64d7) and it works fine. 
The test error got resolved by adding
        (None, None) => styles[Plus],// FIXME!
to line 256.
You hab already FIXME there, so I guess you are aware.
I like this feature. It makes non side-by-side much more readable.

dandavison:
See also https://github.com/dandavison/delta/pull/1558

lukejans:
Is it possible to add a few more niceties that GitHub adds. One thing I was thinking of was when the line wrapping occurs with line numbers set to true that it adds a blank line number column. Another thing that would be cool is for a GitHub style hunk header. I already kinda got this working by using `delta --color-only` when setting the pager in the git config.

**current behaviour**:
```
• @@ -16,61 +16,68 @@ setup() {
  11┊  11│ a long line that wraps to a new line but breaks
the line number format.
  12┊  12│
  13┊  13│     foo
  14┊    │-    deleted this
    ┊  15│+    added this
  16┊  16│
```

**desired behaviour**:
```
   •┊   •| @@ -11,10 +16,10 @@ hunk-header {
  11┊  11│ a long line that wraps to a new line but breaks
    ┊    | the line number format.
  12┊  12│
  13┊  13│ foo
  14┊    │-deleted this
    ┊  15│+added this
  16┊  16│
```

PR #1013, opened 2022-03-16
Librarification: git-delta-lib

th1000s:
Split delta into a binary (`git-delta`) and library (`git-delta-lib`) crate, and enable using the `main` function in the `git-delta` crate[1]. Both continue to live in the same repository but on separate branches.

The crate name is `git-delta-lib` but internally it is known as `libdelta`, maybe change it to `delta_lib` or so?

Development would switch to the library (which also generates a `target/release/delta`) and for a new `git-delta` release the corresponding commits in `git-delta-lib` would be published. This version number would then be used as a dependency in `git-delta`, and its own version number would be bumped, as in [2].

Should the current `master` branch become the `git-delta-lib` branch? The `git-delta` one (which just needs the `main.rs`, `Cargo.*` and the `README.md`, maybe a few more files) could be created/started from scratch to avoid a commit which just deletes a lot of files. The `README.md` should probably be mirrored though.

1:  see a5ed3504d939aaf to preview `git-delta`, which also has a very small `Cargo.toml` now
```rust
use libdelta::mainfn::mainfn;

fn main() -> std::io::Result<()> {
    mainfn()
}
```

2:
```toml
[package]
name = "git-delta"
# ...
version = "0.13.0"

[[bin]]
name = "delta"
path = "src/main.rs"

[dependencies]
git-delta-lib = "0.0.0"
# plus a patch section  for now
```


dandavison:
This is exciting!

I think I'm going to need to ask a few questions before I fully understand what's being proposed.

> Both continue to live in the same repository but on separate branches.

Would this be temporary, or is that the proposed way that things will be in the future?

Two Rust application projects with multi-crate structure that come to mind are [bat](https://github.com/sharkdp/bat/) (I may be wrong but as I understand it this publishes a single `bat` crate containing both a library and an application which uses the library?) and [ripgrep](https://github.com/BurntSushi/ripgrep/) (again my understanding is superficial and may be wrong but I think it publishes multiple crates using Cargo's "workspaces", one of which is the application). How does our situation and proposed strategy compare with these?

It would be nice if everything could proceed with minimal (risk of) disruption to CD pipelines and to the various package  managers / distros that package delta. I do not know how those work (and did not set any of them up) but I assume that they rely on the delta application being built in branch `master` via `cargo build --release`.

p.s. I believe that rebasing this branch will clear up the confusing diff involving `wrapping.rs`.

th1000s:
With this patch `git-delta-lib` would become a multi crate which provides both the binary and the library, but both are tied to the same version number `0.0.x`. While working on the librarification nothing should change for the delta command line interface, so it gets/keeps its separate crate to signify that (otherwise, to make library changes the number would have to be bumped to 0.13, 0.14 etc. in quick succession and thus new binary releases with no apparent changes would happen).

Once this is stable the number can be bumped and `git-delta` can provide binary and library interfaces:

```
--(0.12.0)----(0.12.1)----.   +(0.13.0)------------>           ,---(1.0.0)---->      # git-delta
                           \                                  /
                            `--(0.0.0)---(0.0.1)----(0.0.2)--'                       # git-delta-lib
```

The distributions seem to use the release tar files, and some may directly use the tags, none of that relies on a specific branch.

And indeed, I noticed that I filed the PR against the wrong branch, not your `git-delta-lib`. I cherry-picked the extra commits to have identical delta binaries, thus the diff might be a bit weird. Could you rebase the `git-delta-lib` branch on top of the current master branch?

dandavison:
> Could you rebase the git-delta-lib branch on top of the current master branch?

Done

>  I noticed that I filed the PR against the wrong branch, not your git-delta-lib.

So should we use the PR branch selector to change the target branch now?

th1000s:
Internally it is `delta_lib` now so it is next to `delta` in e.g. the `cargo doc` output. I reduced the public interface to just `mainfn()` and added minimal docs. I think this could be published as `0.0.1` (you still have to increment the version number).

dandavison:
Fantastic. And thanks for the explanations.

PR #1017, opened 2022-03-18
Delta library

dandavison:
From @th1000s in https://github.com/dandavison/delta/pull/1013#issuecomment-1069772562

With this patch `git-delta-lib` would become a multi crate which provides both the binary and the library, but both are tied to the same version number `0.0.x`. While working on the librarification nothing should change for the delta command line interface, so it gets/keeps its separate crate to signify that (otherwise, to make library changes the number would have to be bumped to 0.13, 0.14 etc. in quick succession and thus new binary releases with no apparent changes would happen).

Once this is stable the number can be bumped and `git-delta` can provide binary and library interfaces:

```
--(0.12.0)----(0.12.1)----.   +(0.13.0)------------>           ,---(1.0.0)---->      # git-delta
                           \                                  /
                            `--(0.0.0)---(0.0.1)----(0.0.2)--'                       # git-delta-lib
```


dandavison:
Uploaded [0.0.1](https://crates.io/crates/git-delta-lib) to crates.io

th1000s:
Indeed, leaving this on the existing master branch is probably best, this way everyone with a cloned delta repo can continue working without resetting the upstream. Also, e.g. [lib.rs](https://lib.rs/crates/git-delta-lib) this is now marked as a `bin+lib` crate!

extrawurst:
What is needed to get this librification to get over the finish line? I see the lib on crates just looks like the binary readme, no docs (that I can find). I am wondering when this is ready for primetime in something like my [gitui](https://github.com/extrawurst/gitui)?

dandavison:
Hi @extrawurst 

> What is needed to get this librification to get over the finish line?

Labour :)

It's definitely an interesting and valuable Rust project for someone to work on. Speaking for myself, my job/life isn't currently giving me time to work on anything but essential Delta activities.

extrawurst:
Is that labour communicated somewhere so it can be picked up by someone?

dandavison:
Good call -- I think that would be your ticket https://github.com/dandavison/delta/issues/317, but that wasn't clearly linked to this work. I've added a comment.

PR #1023, opened 2022-03-24
Add --panel-center-offset

th1000s:
Changes the relative size of the left/right panels in side-by-side mode.

Also allow `--width` to take negative values from the command line.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -837,6 +837,17 @@ pub struct Opt {
     /// Options are: auto, always, and never.
     pub paging_mode: String,
 
+    #[clap(
+        long = "panel-center-offset",
+        value_name = "OFFSET",
+        allow_hyphen_values = true
+    )]
+    /// Offset the center between the left and right panels.
+    ///
+    /// In side-by-side mode move the center between the two panels to the left (negative values)
+    /// or to the right (positive values).
+    pub panel_center_offset: Option<String>,
+
     #[clap(long = "parse-ansi")]
     /// Display ANSI color escape sequences in human-readable form.
     ///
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -837,6 +837,17 @@ pub struct Opt {
...
+    #[clap(
+        long = "panel-center-offset",
+        value_name = "OFFSET",
+        allow_hyphen_values = true
+    )]
+    /// Offset the center between the left and right panels.
+    ///
+    /// In side-by-side mode move the center between the two panels to the left (negative values)
+    /// or to the right (positive values).
+    pub panel_center_offset: Option<String>,
```
dandavison, line 849: Is it possible to have the default value of "0" declared here? (I think having a default in the `--help` text helps readers a bit with comprehending how to use the option.)

dandavison: (I just have this one question on this PR)

th1000s: Oh, sorry, I forgot to mention that this is done already. While working on it I added support for an offset of ±9999 (which hides one side completely). For that to work the actual terminal width must be known, and to avoid copying this[1] statement again I separated the terminal width from the full-width-decorations setting, please have a look at that first.

1:
```rust
match width {
    Fixed(w) => w / 2,
    Variable => actual_width_from_elsewhere / 2,
};
```

dandavison: Sorry I didn't mean to forget about this PR!

I think what I was trying to say is that, since its default value is effectively "0" (as returned by the `None` match arm, is it possible for the clap entry to represent it, not as `Option<String>` but rather as plain `String`, with the `default_value` set to `"0"`, so that the default is printed out in `--help` text?

th1000s: Oh, now I see, so the option is mostly self-explanatory. Yes, good idea.

PR #1024, opened 2022-03-28
Separate output-width and full-width-decorations.

th1000s:
An argument like --width=123+variable was not possible, the
Width::Variable variant always used the full terminal
width - which it did not hold and had to be provided.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -980,15 +980,26 @@ pub struct Opt {
     /// Defaults to color.diff.whitespace if that is set in git config, or else 'magenta reverse'.
     pub whitespace_error_style: String,
 
-    #[clap(short = 'w', long = "width", value_name = "N")]
-    /// The width of underline/overline decorations.
+    #[clap(
+        short = 'w',
+        long = "width",
+        value_name = "N",
+        allow_hyphen_values = true
+    )]
+    /// The width of underline/overline(?) decorations.
     ///
     /// Examples: "72" (exactly 72 characters), "-2" (auto-detected terminal width minus 2). An
     /// expression such as "74-2" is also valid (equivalent to 72 but may be useful if the caller
-    /// has a variable holding the value "74"). Use --width=variable to extend decorations and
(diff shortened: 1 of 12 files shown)


src/tests/test_example_diffs.rs
```diff
@@ -198,10 +199,17 @@ mod tests {
...
+            "49",
+            "--background-color-extends-to-terminal-width-NOT",
+        ]);
         let output = integration_test_utils::run_delta(SUBMODULE_DIFF_LOG, &config);
         let output = strip_ansi_codes(&output);
-        assert_eq!(output, SUBMODULE_DIFF_LOG_EXPECTED_OUTPUT);
+        println!("HAVE\n{}\n####\n", output);
+
+        println!("WANT:\n{}", SUBMODULE_DIFF_LOG_EXPECTED_OUTPUT);
+        // assert_eq!(output, SUBMODULE_DIFF_LOG_EXPECTED_OUTPUT);
```
th1000s, line 212: This also influences (and influenced via `--width=variable`) whether the output of the submodule listing is this


```
Submodule submoduleB 0ffa700..0c8b00d:
─────────────────────────────────────────────────
```

or that

```
Submodule submoduleB 0ffa700..0c8b00d:
──────────────────────────────────────
```

src/cli.rs
```diff
@@ -980,15 +980,26 @@ pub struct Opt {
...
+    pub width: Option<String>,
+
+    #[clap(long = "background-color-extends-to-terminal-width-NOT")]
+    /// What am I?
+    ///
+    /// Use --width=variable to extend decorations and
     /// background colors to the end of the text only. Otherwise background colors extend to the
     /// full terminal width.
-    pub width: Option<String>,
+    pub background_color_extends_to_terminal_width_not: bool,
```
th1000s, line 1003: Because bools default to `false` this either needs a name like `decorations-limited-to-textwidth` (?) or an `Option<String>` which then defaults to full-width-decorations.

PR #1030, opened 2022-03-31
Add terminal width fallback via stty if on windows/MSYS2

th1000s:


th1000s:
I could finally verify that this workaround fixes the problem on Windows.

Also added a `DELTA_NO_WORKAROUNDS` env var to make future debugging easier, e.g. when eminence/terminal-size#25 is fixed.

src/utils/workarounds.rs
```diff
@@ -0,0 +1,60 @@
...
+    match (height, width) {
+        (24..=25, 79..=80) => guess_real_width(width, term_stdout).unwrap_or(width),
+        _ => width,
+    }
+    .into()
+}
+
+#[cfg(not(target_os = "windows"))]
+pub fn windows_msys2_width_fix(height_width: (u16, u16), _: &console::Term) -> usize {
+    let _ = NO_WORKAROUNDS;
```
dandavison, line 58: What's this for?

th1000s: Just so this const &str does not trigger an unused variable warning in the non-windows case.

dandavison: Ah of course, thanks.

dandavison:
Thank you @th1000s.

PR #1037, opened 2022-04-06
Fix warning highlight for trailing whitespace

### LINKED ISSUE #137: Highlight empty lines?
The standard pager for git will highlight additions of lines whose only content are spaces or tabs in red.  Is it possible to have similar behavior in delta?

Thanks in advance.

wescande:
Fix #137
and added some corresponding test

`git diff` without pager on the left, current delta on the middle, proposed version on the right
image

---

Because the whitespace error style is applied on a complete section I needed to split the section in smaller section when there was space at the end. Thus I needed to edit some of `test_infer_edits`. 

I suppose this change does not contains all the test case for handling the trailing whitespace perfectly.
Maybe this should also be done in :
* align::Operation::Substitution
* align::Operation::Insertion

### DIFF
--- a/src/edits.rs
+++ b/src/edits.rs
@@ -90,12 +90,31 @@ where
     }
     // Emit any remaining plus lines
     for plus_line in &plus_lines[plus_index..] {
-        annotated_plus_lines.push(vec![(noop_insertions[plus_index], plus_line)]);
+        if let Some(content) = get_contents_before_trailing_whitespace(plus_line) {
+            annotated_plus_lines.push(vec![
+                (noop_insertions[plus_index], content),
+                (noop_insertions[plus_index], &plus_line[content.len()..]),
+            ]);
+        } else {
+            annotated_plus_lines.push(vec![(noop_insertions[plus_index], plus_line)]);
+        }
         line_alignment.push((None, Some(plus_index)));
         plus_index += 1;
     }
 
     (annotated_minus_lines, annotated_plus_lines, line_alignment)
+}
+
+// Return `None` if there is no trailing whitespace.
+// Return `Some(content)` where content is trimmed if there was some trailing whitespace
+fn get_contents_before_trailing_whitespace(line: &str) -> Option<&str> {
+    let content = line.trim_end();
+    // if line has a trailing newline, do not consider it as a 'trailing whitespace'
+    if !content.is_empty() && content != line.trim_end_matches('\n') {
+        Some(content)
+    } else {
+        None
+    }
 }
 
 // Return boolean arrays indicating whether each line has a homolog (is "paired").
@@ -228,14 +247,19 @@ where
                     },
                     minus_section,
                 ));
-                annotated_plus_line.push((
-                    if coalesce_space_with_previous {
-                        plus_op_prev
-                    } else {
-                        noop_insertion
-                    },
-                    plus_section(n, &mut y_offset),
-                ));
+                let op = if coalesce_space_with_previous {
+                    plus_op_prev
+                } else {
+                    noop_insertion
+                };
+                let plus_section = plus_section(n, &mut y_offset);
+                if let Some(non_whitespace) = get_contents_before_trailing_whitespace(plus_section)
+                {
+                    annotated_plus_line.push((op, non_whitespace));
+                    annotated_plus_line.push((plus_op_prev, &plus_section[non_whitespace.len()..]));
+                } else {
+                    annotated_plus_line.push((op, plus_section));
+                }
                 minus_op_prev = noop_deletion;
                 plus_op_prev = noop_insertion;
             }
@@ -553,7 +577,8 @@ mod tests {
                     ],
                 ],
                 vec![vec![
-                    (PlusNoop, "á á b á á "),
+                    (PlusNoop, "á á b á á"),
+                    (PlusNoop, " "),
                     (Insertion, "c"),
                     (PlusNoop, " á á b"),
                 ]],
@@ -574,9 +599,19 @@ mod tests {
                     vec![(MinusNoop, "cccc "), (Deletion, "c"), (MinusNoop, " ccc")],
                 ],
                 vec![
-                    vec![(PlusNoop, "bbbb "), (Insertion, "!"), (PlusNoop, " bbb")],
+                    vec![
+                        (PlusNoop, "bbbb"),
+                        (PlusNoop, " "),
+                        (Insertion, "!"),
+                        (PlusNoop, " bbb"),
+                    ],
                     vec![(PlusNoop, "dddd d ddd")],
-                    vec![(PlusNoop, "cccc "), (Insertion, "!"), (PlusNoop, " ccc")],
+                    vec![
+                        (PlusNoop, "cccc"),
+                        (PlusNoop, " "),
+                        (Insertion, "!"),
+                        (PlusNoop, " ccc"),
+                    ],
                 ],
             ),
             0.66,
@@ -615,7 +650,8 @@ mod tests {
                     (MinusNoop, "EditOperation>("),
                 ]],
                 vec![vec![
-                    (PlusNoop, "fn coalesce_edits<'a, "),
+                    (PlusNoop, "fn coalesce_edits<'a,"),
+                    (PlusNoop, " "),
                     (Insertion, "'b, "),
                     (PlusNoop, "EditOperation>("),
                 ]],
@@ -636,7 +672,8 @@ mod tests {
                     (MinusNoop, ":"),
                 ]],
                 vec![vec![
-                    (PlusNoop, "for _ in range(0, "),
+                    (PlusNoop, "for _ in range(0,"),
+                    (PlusNoop, " "),
                     (Insertion, "int("),
                     (PlusNoop, "options[\"count\"])"),
                     (Insertion, ")"),
@@ -654,7 +691,12 @@ mod tests {
             vec!["a b a"],
             (
                 vec![vec![(MinusNoop, "a "), (MinusNoop, "a")]],
-                vec![vec![(PlusNoop, "a "), (Insertion, "b "), (PlusNoop, "a")]],
+                vec![vec![
+                    (PlusNoop, "a"),
+                    (PlusNoop, " "),
+                    (Insertion, "b "),
+                    (PlusNoop, "a"),
+                ]],
             ),
             1.0,
         );
@@ -663,7 +705,12 @@ mod tests {
             vec!["a b b a"],
             (
                 vec![vec![(MinusNoop, "a "), (MinusNoop, "a")]],
-                vec![vec![(PlusNoop, "a "), (Insertion, "b b "), (PlusNoop, "a")]],
+                vec![vec![
+                    (PlusNoop, "a"),
+                    (PlusNoop, " "),
+                    (Insertion, "b b "),
+                    (PlusNoop, "a"),
+                ]],
             ),
             1.0,
         );
@@ -684,7 +731,8 @@ mod tests {
                     (MinusNoop, " from any one of them."),
                 ]],
                 vec![vec![
-                    (PlusNoop, "so it is safe to read "),
+                    (PlusNoop, "so it is safe to read"),
+                    (PlusNoop, " "),
                     (Insertion, "build"),
                     (Insertion, " "),
                     (Insertion, "info"),
@@ -761,7 +809,7 @@ mod tests {
 
     #[test]
     fn test_infer_edits_14() {
-        assert_paired_edits(
+        assert_edits(
             vec!["a b c d ", "p "],
             vec!["x y c z ", "q r"],
             (
@@ -783,7 +831,8 @@ mod tests {
                         (Insertion, "x"),
                         (Insertion, " "),
                         (Insertion, "y"),
-                        (PlusNoop, " c "),
+                        (PlusNoop, " c"),
+                        (Insertion, " "),
                         (Insertion, "z"),
                         (PlusNoop, " "),
                     ],
@@ -795,6 +844,7 @@ mod tests {
                     ],
                 ],
             ),
+            1.0,
         );
     }
 
@@ -834,7 +884,8 @@ mod tests {
                 vec![vec![
                     (PlusNoop, ""),
                     (Insertion, "c "),
-                    (PlusNoop, "a a a a a a "),
+                    (PlusNoop, "a a a a a a"),
+                    (Insertion, " "),
                     (Insertion, "c"),
                     (Insertion, " "),
                     (Insertion, "c"),
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -522,9 +522,15 @@ impl<'p> Painter<'p> {
             let line_has_emph_and_non_emph_sections =
                 style_sections_contain_more_than_one_style(style_sections);
             let should_update_non_emph_styles = non_emph_style.is_some() && *line_has_homolog;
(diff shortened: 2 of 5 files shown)


src/edits.rs
```diff
@@ -90,7 +90,16 @@ where
     }
     // Emit any remaining plus lines
     for plus_line in &plus_lines[plus_index..] {
-        annotated_plus_lines.push(vec![(noop_insertions[plus_index], plus_line)]);
+        // Split the line if it contains extra whitespace at the end preceded with non whitespace
+        let line = plus_line.trim_end_matches(&[' ', '\t', '\n']);
```
martinetd, line 94: \n won't match anything if this is already line-split, did you mean \r ?

martinetd:
Thanks for this! I noticed the feature missing when I got a warning from `git add -p` about whitespace error and been using this for a couple of hours without problem.

A couple of notes regarding differences with `git --no-pager diff` since you noted trailing new lines as a todo:

- you only consider actively adding whitespaces at the end of the line, but git prints spaces in red on modified lines even if it's not being actively added, e.g.:
```
-some line with trailing spaces    
+some new line with trailing spaces    
```
would print red at the end.

- while we are here, git also warns when adding whitespaces at start of a line followed by tabs (`^ +\t` would have spaces in red).

That being said, it works for me™ and code doesn't look obviously bad to me.

src/edits.rs
wescande: The line still contain the `\n` at the end and it need to be taken care of

wescande:
> you only consider actively adding white spaces at the end of the line, but git prints spaces in red on modified lines even if it's not being actively added, e.g.:

I updated the CL to take this into account ! thanks for pointing it out. There is now a new test with this specific setup.

> while we are here, git also warns when adding white spaces at start of a line followed by tabs (^ +\t would have spaces in red).

Not sure how to do that with current implementation, as all the `\t` are converted to space before applying styles. It would require a deeper refactoring of line handling. We can split the line to have a section specific for the starting space before a tab, but how can we be sure when applying the style that it's a white space error ?

amtoine:
if this PR is stale, i'd love to take it and complete it, i truely need this feature to make proper reviews :relieved:

amtoine:
the PR works but there are just conflicts to solve :yum: 

also i'd say that maybe having the addition whitespaces in green and the removal whitespaces in red would look a bit better rather than all red? :blush:

dandavison:
@wescande I'm very sorry I didn't give this work proper attention when you did it. I think there are two factors: (1) it arrived at a time when I had a new job and (2) it's a pretty technical area and for some reason I hadn't noticed problems with whitespace highlighting.

Hi @amtoine, very nice to see you here! Unless @wescande wants to work on this then it would be fantastic if you could finalize this work. I'd like to suggest a path forwards.

1. Could you create a PR containing a set of (failing) tests that demonstrate the bug(s) that we are fixing. This will be extremely helpful for me to understand exactly what we are fixing. This step might be quite quick. See also the discussion in #137. We can merge this PR with `#[ignore]` annotations on the failing tests.

2. Then, let's discuss briefly to check that this PR is indeed the way we want to fix those bugs. Once we are in agreement about exactly what the behavior should be, a second PR with the fix (starting with a commit to remove the `#[ignore]` annotations).

What do you think?

wescande:
Hi 
I forgot about this me too :)

Hi @amtoine , I rebased the PR in order to solve the merge conflict in `src/edit.rs` (it was a trivial fix) but it looks like there are some logical change that conflict with this change when I look at the output on some diff.
I did the same example as in the first comment and now it appear like this: :(
image
**left == git raw**  | **middle == delta master** | **right == this PR**.
The diff data is the one from  `DIFF_WITH_WHITESPACE_COMPLEX_ERROR` in `test_example_diffs.rs`)

Because of that, I am no longer able to run the `test_whitespace_complex_error` (looks like the test was a good test ^^)
There are 2 other tests that are failing and I don't know how to easily fix them
`test_infer_edits_14` and `test_infer_edits_9`. 
I suspect all these issue are caused by #1244  that changed the way the token were grouped together.

> the PR works but there are just conflicts to solve yum

For me it no longer works. Did you do some more change ?

> also i'd say that maybe having the addition whitespaces in green and the removal whitespaces in red would look a bit better rather than all red? blush

If Dan think it is a good idea why not, but I think atomic change are better and therefore it should not be done in this PR.
Also I like having it display in pale red, it make it obvious it is an error. (and isn't it possible to change the theme color to put it like you want  ?)

> i'd love to take it and complete it,

Please feel free to do so :pray: 


Hi @dandavison, I am not sure I understand your concern. There are test added with the PR that cannot pass without it (even with it they do not pass anymore :sob: ). What would be the value of adding the test separately ?

The behavior is well explained in the first picture, and can be summarized in "trailing whitespace does not appear as an error".
So this:
```
"- bar  "
"+ bAr  "
```
Does not display the 2 spaces at the end of bar as an error.

wescande:
A few hours laters...
I added a new small test (a subset of the whitespace_complex) + fix the` test_infer_edits_9`

Current patch still fail `test_infer_edits_14` and the `test_whitespace_complex_error`.
But it is already looking better since `foo1` now has a proper whitespace warning.



I'll have a look at the empty line now

wescande:
And done :heavy_check_mark:  !
 We are back to the visual from the first comment of this PR. and all test are green :tada: 

I fixed the `test_infer_edits_14`. Test was failing because it expected both minus and plus line to have the same number of token, but this PR increase the number of token for plus line in order to change the style only from whitespace error.

I fixed the last test_whitespace_complex_error and added yet another test related to empty line whitespace. This actually made me realize I could simplify the logic and remove the temporary vec storage

----
Unrelated:
While writing this message I noticed that windows and macos unit test in the CI  are twice slower than linux unit-test. Not sure if this is expected.

amtoine:
@wescande 
amazing, very quick :star_struck: 
i'm fine if this PR lands, no need to almost start from scratch if @dandavison is ok with the changes :relieved: 

> If Dan think it is a good idea why not, but I think atomic change are better and therefore it should not be done in this PR.
>
ohh yes of course :+1: 
i could not find anything looking like color control in your changes, do you know where this is?

> Also I like having it display in pale red, it make it obvious it is an error.
>
but it's not always an error, right?
for instance, in a *GitHub* README, to force a newline, you add two trailing whitespaces at the end of the line :yum:

> and isn't it possible to change the theme color to put it like you want ?
>
i could not find anything in the `help` of `delta` when searching for "whitespace" :confused: 


again, thanks for taking this after all this time, i think it's a truely important feature :relieved: 
i really do not understand why *GitHub* does not do the same :scream:

src/edits.rs
```diff
@@ -228,14 +237,19 @@ where
...
+                    noop_insertion
+                };
+                let plus_section = plus_section(n, &mut y_offset);
+                let line = plus_section.trim_end_matches(&[' ', '\t', '\n']);
+                if !line.is_empty() && line != plus_section.trim_end_matches('\n') {
+                    annotated_plus_line.push((op, line));
+                    annotated_plus_line.push((plus_op_prev, &plus_section[line.len()..]));
+                } else {
+                    annotated_plus_line.push((op, plus_section));
+                }
```
dandavison, line 252: @wescande Thanks very much for wading into what must be the most complex/obscure part of the codebase! I haven't fully paged in how this all works. Can you help me: it looks like there's some quite similar logic going on here and on line 95 above; why is that and is it possible to do it just once rather than twice?

Also, WDYT about writing it something like the following, to make it self-documenting and easier to understand for future readers:

```rust
if let Some(non_whitespace) = get_contents_before_trailing_whitespace(plus_section)
{
    annotated_plus_line.push((op, non_whitespace));
    annotated_plus_line.push((plus_op_prev, &plus_section[non_whitespace.len()..]));
} else {
    annotated_plus_line.push((op, plus_section));
}

fn get_contents_before_trailing_whitespace(line: &str) -> Option<&str> {
    let line = line.trim_end_matches(&[' ', '\t', '\n']);
    if !line.is_empty() && line != line.trim_end_matches('\n') {
        Some(line)
    } else {
        None
    }
}
```

src/paint.rs
```diff
@@ -522,9 +522,15 @@ impl<'p> Painter<'p> {
...
-            for (style, _) in style_sections.iter_mut() {
+
+            // TODO: Git recognizes blank lines at end of file (blank-at-eof)
+            // as a whitespace error but delta does not yet.
+            // https://git-scm.com/docs/git-config#Documentation/git-config.txt-corewhitespace
+            let mut is_whitespace_error = whitespace_error_style.is_some();
+            for (style, s) in style_sections.iter_mut().rev() {
+                if is_whitespace_error && !s.trim().is_empty() {
+                    is_whitespace_error = false;
+                }
```
dandavison, line 533: Why does `!s.trim().is_empty()` mean that `s` is not a whitespace error? Isn't `s = "xxx  "` a counterexample? Are you redefining `is_whitespace_error` to mean "whole line is whitespace"?

dandavison (changes requested):
Thanks very much for working on this @wescande. It's a really obscure part of the code and I've asked a couple of questions about the changed non-test code.

src/paint.rs
wescande: Because of the modification inside `edit.rs` I should not have to deal with `"xxx "` as a token but `["xxx", " "]` instead.
So the whole  `" "`is a token with whitespace error.

I am indeed re-using the `is_whitespace_error` to be flexible between "whole line is whitespace" and "part of the line is whitespace".

So to answer your question, since I am iterative with a `rev()`, I can say something like while I am encountering only whitespace, I apply the whitespace_error style. Once I encounter a non-whitespace token, I must stop applying the whitespace error style for the whole line

----

I also have a local change to consider this as an error :  `"   \t"` as asked by martinetd,  but it only works with the option `--tabs=0` so I do not want to bloat this PR with it. making it works with and without the tab expansion will make it a heavy refactor of the codebase (and I do not use tabs so ... :shrug: )

martinetd: > I also have a local change to consider this as an error : "   \t" as asked by martinetd, but it only works with the option --tabs=0 so I do not want to bloat this PR with it. making it works with and without the tab expansion will make it a heavy refactor of the codebase (and I do not use tabs so ... shrug )

Yeah; I think that can come later (if at all).

Unfortunately even if I do have a properly configured editor, patches I review often don't so displaying such problems can be useful.
In that case, just displaying "<space><tab>" as an error isn't enough as having someone add a line with tabs in the middle of a space-indented file is also a problem, so displaying tabs differently would also be necessary -- and we're getting quite out of scope of this PR.
(expanding the offtopic point: gerrit displays them with a lightly shaded `»   ` of appropriate size, which I kind of like and have reproduced in gertty at some point, so I might do that here eventually)

Your work so far is already more than useful, and it's great to see interest to merge it now. Thanks a bunch everyone! :)

src/edits.rs
wescande: Agreed this is not an easy part

Yes we can definitively do that only once, I did it when trying to make a more complex split with the `" \t"`.
I uploaded the change accordingly. It is not really smaller but it will easier for re-reader :+1:

wescande:
>  i could not find anything looking like color control in your changes, do you know where this is?

I do not know :/

> for instance, in a GitHub README, to force a newline, you add two trailing whitespaces at the end of the line yum

@amtoine  I do recommend using `<br>` instead of double trailing space in markdown file. But I do not know if github handle that properly. (and occasionally I still use the double space ... :innocent: ) 

> i could not find anything in the help of delta when searching for "whitespace" confused

Some test of this PR are using `--whitespace-error-style`, it should be what you are looking for. And by looking at themes.gitconfig some theme are already defining some color.

amtoine:
@dandavison 
the tests i've conducted give good results, solving the issue of invisible whitespaces :ok_hand: :yum:

src/paint.rs
```diff
@@ -522,9 +522,15 @@ impl<'p> Painter<'p> {
...
-            let is_whitespace_error =
-                whitespace_error_style.is_some() && is_whitespace_error(style_sections);
-            for (style, _) in style_sections.iter_mut() {
+
+            // TODO: Git recognizes blank lines at end of file (blank-at-eof)
+            // as a whitespace error but delta does not yet.
+            // https://git-scm.com/docs/git-config#Documentation/git-config.txt-corewhitespace
+            let mut is_whitespace_error = whitespace_error_style.is_some();
+            for (style, s) in style_sections.iter_mut().rev() {
+                if is_whitespace_error && !s.trim().is_empty() {
```
dandavison, line 531: Re the comment above: here we're using `trim()` and there `trim_end_matches()`. It feels like we should use the same method everywhere?

src/edits.rs
```diff
@@ -228,14 +246,18 @@ where
...
-                        noop_insertion
-                    },
-                    plus_section(n, &mut y_offset),
-                ));
+                let op = if coalesce_space_with_previous {
+                    plus_op_prev
+                } else {
+                    noop_insertion
+                };
+                let line = plus_section(n, &mut y_offset);
```
dandavison, line 254: It looks like this variable should be named `plus_section` for consistency with `minus_section` and `plus_section` elsewhere in this function.

dandavison:
@wescande there's a couple of questions here that I'd love to get your answers to while this is still fresh in your brain. But other than that I'm ready to ship this and get it in the next release.

src/paint.rs
```diff
@@ -522,9 +522,15 @@ impl<'p> Painter<'p> {
             let line_has_emph_and_non_emph_sections =
                 style_sections_contain_more_than_one_style(style_sections);
             let should_update_non_emph_styles = non_emph_style.is_some() && *line_has_homolog;
-            let is_whitespace_error =
-                whitespace_error_style.is_some() && is_whitespace_error(style_sections);
-            for (style, _) in style_sections.iter_mut() {
+
+            // TODO: Git recognizes blank lines at end of file (blank-at-eof)
+            // as a whitespace error but delta does not yet.
```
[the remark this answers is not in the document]
wescande: Yes this is still a valid TODO
Cf this screenshot

src/edits.rs
```diff
@@ -90,14 +90,32 @@ where
...
         plus_index += 1;
     }
 
     (annotated_minus_lines, annotated_plus_lines, line_alignment)
 }
 
+// Return `None` if there is no trailing whitespace.
+// Return `Some(content)` where content is trimmed if there was some trailing whitespace
+fn get_contents_before_trailing_whitespace(line: &str) -> Option<&str> {
+    let content = line.trim_end_matches(&[' ', '\t', '\n']);
```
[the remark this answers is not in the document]
wescande: Replaced with `trim_end()`.
I was just using the same method as the one just below. No specific reason for it there, but there is a reason for the one just below. Added a comment saying so

src/paint.rs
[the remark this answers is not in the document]
wescande: Me from the past already open #1038  but unless advertise on it I no longer have this local change.

wescande:
Sorry, github notification are not appearing in my main mail box and I missed your comment.
I resolved all of them. 
Feel free to take another look and add more comment if you have some.

dandavison (approved):
Fantastic, thanks very much for this work @wescande!

dandavison:
I was hoping to get ripgrep-format output into the next release: https://github.com/dandavison/delta/pull/1410

amtoine:
gotcha, just to have an idea :relieved: 

for the time being, i've compiled from source and it works like a charm so far :star_struck:

PR #1044, opened 2022-04-13
Remove `-style` suffix for delta colours

srasku:
If you append the `-color` suffix you get the following error:
    - Invalid color or style attribute:

### DIFF
--- a/manual/src/choosing-colors-styles.md
+++ b/manual/src/choosing-colors-styles.md
@@ -8,7 +8,7 @@ Here's an example:
 
 ```gitconfig
 [delta]
-    minus-style = red bold ul "#ffeeee"
+    minus = red bold ul "#ffeeee"
 ```
 
 That means: For removed lines, set the foreground (text) color to 'red', make it bold and underlined, and set the background color to `#ffeeee`.


dandavison:
Hi @srasku, thanks! So, hang on, I'm slightly confused. It's true that `minus-color` is incorrect. The correct form is `minus-style`. But your diff is changing it to just `minus`, which isn't correct.

srasku:
@dandavison, I made an error in the commit message.  The headline is correct but the body is wrong.  The whole thing should read:
> Remove `-style` suffix for delta colours
>
>If you append the `-style` suffix you get the following error:
>    - Invalid color or style attribute:

I do get the error message specified unless I remove the suffix.
```
(venv) stephenrasku@CLIR-00161 Code/clir 691 $ cat ~/.gitconfig
...
[core]
	pager = delta
	excludesfile = ~/.gitignore_global
[interactive]
[delta]
	navigate = true
	minus-style = red bold ul '#ffeeee'
[merge]
	conflictstyle = diff3
[diff]
	colorMoved = default
(venv) $ git diff
Invalid color or style attribute: 
(venv) $ cat ~/.gitconfig
...
	pager = delta
	excludesfile = ~/.gitignore_global
[interactive]
[delta]
	navigate = true
	minus = red bold ul '#ffeeee'
[merge]
	conflictstyle = diff3
[diff]
	colorMoved = default
(venv) $ git diff
(venv) $
```

dandavison:
Hi @srasku, the problem is that you're using single quote characters around your color. If you change that to `red bold ul "#ffeeee"` it works. The reason you didn't see an error with `minus` is that `minus` isn't a value that delta recognizes! But `minus-style` is the correct form: https://github.com/dandavison/delta/blob/3ccf2bc8d26b9296a08110d9ac7bc60d86bbeaa6/src/cli.rs#L797

srasku:
Yes, I can confirm this is the issue.  Thanks for clearing it up.

PR #1057, opened 2022-04-25
Fix git-grep match-highlighting at line-start

### LINKED ISSUE #1056: 🐛 `git grep` handler doesn't highlight matches that start in column 0
Hi! It looks like the `git grep` handler can't highlight the grep-match if the grep-match starts at the beginning of the line of code. I'm not sure if this is always the case or if it's specific to my type of terminal in some way (I'm using Terminal.app on macOS).

To repro, 

```
mkdir /tmp/delta-test
cd /tmp/delta-test
echo 'foo' > broken.md
echo '...foo' > working.md
git init
git add .
git commit -m 'Add files to repro bug'
git grep 'foo'
```

On my machine, that leaves us with one line where foo is highlighted (`working.md`) and another line where it is not (`broken.md`):

Screen Shot 2022-04-25 at 7 49 27 PM

Not a big deal for a trivial case like this (where the entire line matches the query exactly), but with a more complicated regex it can be annoying.

I have a fix for this. It's a little bit weird, but I _think_ it should be okay.

- [x] Please include the raw text output from git, so that we can reproduce the problem.
The problem is in handling of ANSI color codes, which I suspect cannot be pasted here? But here's a `cat -v` representation of what `GIT_PAGER=less git color.grep=always grep 'foo'` does here:

```
broken.md^[[36m:^[[m1^[[36m:^[[m^[[1;31mfoo^[[m
working.md^[[36m:^[[m1^[[36m:^[[m...^[[1;31mfoo^[[m
```

(Pretty hard to visually parse this, but: it highlights "foo" correctly on both lines.)

- [x] A screenshot of Delta's output is often helpful also.

Thanks for filing a Delta bug report!

jdpopkin:
This commit fixes highlighting the part of the line of code that matches
the git grep query in cases where the match starts at the beginning of
the lines.

Fixes #1056.

### DIFF
--- a/src/handlers/grep.rs
+++ b/src/handlers/grep.rs
@@ -222,14 +222,14 @@ fn get_code_style_sections<'b>(
     non_match_style: Style,
     grep: &GrepLine,
 ) -> Option<StyleSectionSpecifier<'b>> {
-    if let Some(raw_code_start) = ansi::ansi_preserving_index(
+    if let Some(prefix_end) = ansi::ansi_preserving_index(
         raw_line,
         match grep.line_number {
-            Some(n) => format!("{}:{}:", grep.path, n).len(),
-            None => grep.path.len() + 1,
+            Some(n) => format!("{}:{}:", grep.path, n).len() - 1,
+            None => grep.path.len(),
         },
     ) {
-        let match_style_sections = ansi::parse_style_sections(&raw_line[raw_code_start..])
+        let match_style_sections = ansi::parse_style_sections(&raw_line[(prefix_end + 1)..])
             .iter()
             .map(|(ansi_term_style, s)| {
                 if ansi_term_style.foreground.is_some() {
@@ -899,5 +899,48 @@ mod tests {
         // No fake parent grep command
         let apparently_grep_output = "src/co-7-fig.rs:xxx";
         assert_eq!(parse_grep_line(apparently_grep_output), None);
+    }
+
+    #[test]
+    fn test_get_code_style_sections() {
+        use crate::ansi::strip_ansi_codes;
+        use crate::handlers::grep::get_code_style_sections;
+        use crate::paint::StyleSectionSpecifier;
+        use crate::style::Style;
+
+        let fake_parent_grep_command = "git --doesnt-matter grep --nor-this nor_this -- nor_this";
+        let _args = FakeParentArgs::for_scope(fake_parent_grep_command);
+
+        let miss = Style::new();
+        let hit = Style {
+            is_emph: true,
+            ..miss
+        };
+
+        let escape = "\x1B";
+        let working_example = format!("foo/bar/baz.yaml{escape}[36m:{escape}[m1090{escape}[36m:{escape}[m  - {escape}[1;31mkind: Service{escape}[mAccount");
+        let stripped = strip_ansi_codes(&working_example);
+        let grep = parse_grep_line(&stripped).unwrap();
+
+        assert_eq!(
+            get_code_style_sections(&working_example, hit, miss, &grep),
+            Some(StyleSectionSpecifier::StyleSections(vec![
+                (miss, "  - "),
+                (hit, "kind: Service"),
+                (miss, "Account")
+            ]))
+        );
+
+        let broken_example = format!("foo/bar/baz.yaml{escape}[36m:{escape}[m2{escape}[36m:{escape}[m{escape}[1;31mkind: Service{escape}[m");
+        let broken_stripped = strip_ansi_codes(&broken_example);
+        let broken_grep = parse_grep_line(&broken_stripped).unwrap();
+
+        assert_eq!(
+            get_code_style_sections(&broken_example, hit, miss, &broken_grep),
+            Some(StyleSectionSpecifier::StyleSections(vec![(
+                hit,
+                "kind: Service"
+            )]))
+        );
     }
 }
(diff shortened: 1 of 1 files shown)


src/handlers/grep.rs
```diff
@@ -900,4 +900,46 @@ mod tests {
...
+        assert_eq!(
+            get_code_style_sections(&working_example, hit, miss, &grep),
+            Some(StyleSectionSpecifier::StyleSections(vec![
+                                                      (miss, "  - "),
+                                                      (hit, "kind: Service"),
+                                                      (miss, "Account")
+            ]))
+        );
+
+        let broken_example = format!("foo/bar/baz.yaml{escape}[36m:{escape}[m2{escape}[36m:{escape}[m{escape}[1;31mkind: Service{escape}[m");
```
jdpopkin, line 934: These variable-names aren't really applicable anymore. `working_example` was the one that worked before the changes were made to `get_code_style_sections`, `broken_example` was the one that didn't. That wouldn't actually be a useful distinction going forward; maybe we should rename them?

src/handlers/grep.rs
```diff
@@ -222,14 +222,14 @@ fn get_code_style_sections<'b>(
...
         raw_line,
         match grep.line_number {
-            Some(n) => format!("{}:{}:", grep.path, n).len(),
-            None => grep.path.len() + 1,
+            Some(n) => format!("{}:{}:", grep.path, n).len() - 1,
+            None => grep.path.len(),
         },
     ) {
-        let match_style_sections = ansi::parse_style_sections(&raw_line[raw_code_start..])
+        let match_style_sections = ansi::parse_style_sections(&raw_line[(prefix_end + 1)..])
```
jdpopkin, line 232: If there are no escape-code-sequences separating the prefix-section from the code-section, the old approach would have produced the same value as the new approach. If there _are_ escape-code-sequences between the two, the old approach would have skipped them, which isn't what we want. So I _think_ this is okay?

dandavison:
Hi @jdpopkin, I'm really sorry for being so slow here. I really appreciate you working on this bug.

I did a quick manual test and I'm thinking that there might be a tiny bit more work needed here. Try `git grep match` in the delta repo -- here's what I get:

### On master
image

### On this branch
image

So it looks like your branch is extending the highlighting too far in some cases. Do you agree / see that also?

jdpopkin:
Yeah, I see the same thing on my end - it looks like my branch gets the highlighting wrong for all lines where the first non-whitespace character is `+` or `-`! I think I'll be able to look into this a bit tonight.

jdpopkin:
So it looks like some of the weirdness here comes from `git grep` itself. Like, if I run `GIT_PAGER=less git grep match` and copy the escape codes out for that first bad line, I get this:

```
etc/examples/189-merge-conflict.2.diff^[[36m:^[[m10^[[36m:^[[m^[[32m +        let (style, non_emph_style) = ^[[1;31mmatch^[[m state {^[[m
```

With ANSI codes spelled out in words and some of the filename removed:

```
[...].diff${escape}${foreground=cyan}:${escape}${}10${escape}${foreground=cyan}:${escape}${}${escape}${foreground=green} +        let (style, non_emph_style) = ${escape}${bold=true,foreground=red}match${escape}${} state {${escape}${}
```

That looks normal in my standard green-on-black terminal, but not in this one:


root_cause

So: for some reason, `git grep` is actually marking the start of these lines as _green_. And that's breaking match-detection because delta chooses between `match_style` and `non_match_style` based on whether or not this style-section has a non-default foreground-color. (Without my branch, delta doesn't have this problem because it can't detect a style-section that includes the first character of the code-section.)

We can work around that by checking for _bold_ sections rather than non-default-foreground-color ones. I'm not really sure whether or not that's, like, a _universal_ answer - there might be things other than pattern-matches that get formatted as bold, and maybe it'll run into trouble with users who customize git color config. If it makes sense to commit to default-git-colors-or-else, it could look for bold+red style-sections. If not, maybe the criteria used by `get_code_style_sections` could be configurable?

Anyway, I added a test that repros the scenario you ran into, along with a code change that makes `get_code_style_sections`'s match-detection decide based on boldness rather than non-default foreground-color. It makes the tests pass, and it looked okay in my arbitrary noodling around with `GIT_PAGER=target/release/delta git grep`. I'll push that up in a moment - let me know what you think.

jdpopkin:
> So: for some reason, git grep is actually marking the start of these lines as green.

As it turns out, the files where this is happening are example-files used for testing raw git output. In this case, these lines are green because they literally contain ANSI color codes making them green. And since ANSI color codes are what delta relies on to tell where the match-boundaries are in `git grep` output, it's getting confused. The general problem applies even without this branch's changes, though - the only difference is that the current implementation cuts out color codes if they start at the beginning of the line.

I think checking for bold or maybe red+bold (the [default format for a grep match](https://github.com/git/git/blob/v2.37.1/grep.h#L192)) instead of checking for a non-default foreground color makes sense here. That would fix the existing problems covering lines in `etc/examples` (though we'd still run into problems on lines that specifically included bold red text). I just pushed an `--amend` that checks for redness as well as boldness - what do you think?

dandavison:
Thanks @jdpopkin! The code change is very clean and it seems to work so I think that's great and I've merged it.

(By the way, I agree that `git grep` is useful but I just wanted to point out for anyone reading that delta handles `rg --json` output, for which there should never be any parsing problems.)

Issue #1072, opened 2022-05-09
Unable to find "github" theme / default theme has poor color contrast on default macOS terminal


toolness:
Hello!

Your README mentions a "github" theme with a link to this screenshot:

> https://user-images.githubusercontent.com/52205/81058911-6abcf700-8e9d-11ea-93be-e212824ec03d.png

However, I have been searching for about 10 minutes and cannot seem to find a way to activate this theme.  The [`themes.gitconfig`](https://github.com/dandavison/delta/blob/master/themes.gitconfig) lists no theme called "github" and I can't find reference to it anywhere else in the documentation.

This is unfortunate since the default theme appears to be made for dark terminals, and the result (on my default MacOS terminal) has very poor [contrast and color accessibility](https://webaim.org/articles/contrast/) which makes it more difficult to read than regular `git diff`.

dandavison:
Hi @toolness, use `syntax-theme = GitHub` to select the GitHub syntax-highlighting theme. (You can also specify `light = true` and GitHub will be selected as the default syntax-highlighting theme for light terminal backgrounds). You can use `delta --list-syntax-themes` and `delta --show-syntax-themes` to view the syntax highlighting themes. In general, `delta --help` is the point-of-truth documentation for all available delta configuration.

There's a distinction between "theme" and "syntax-highlighting theme", which is described here: https://dandavison.github.io/delta/custom-themes.html Basically a "theme" is a broader concept; one thing you can specify in a "theme" is the "syntax-highlighting theme".

PR #1073, opened 2022-05-09
Add 'light = false' w/ comment in gitconfig

toolness:
As mentioned in #1072, the macOS default terminal is light and it's challenging to find documentation on how to make delta copacetic with it, so this adds an explicit `light = false` in the `.gitconfig` example, with a comment explaining that it can be changed for light terminals.

(Oops, looks like my indentation is messed up, will fix in a subsequent commit)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@
 
 [delta]
     navigate = true  # use n and N to move between diff sections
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks! That seems helpful.

toolness:
Sure thing, thanks for making this great tool!

Issue #1144, opened 2022-08-02
🐛 truncate_str does not truncate to exact width


SheldonNico:
- [x ] Please include the raw text output from git, so that we can reproduce the problem.
      (You can use `git --no-pager` to produce the raw text output.)
```
diff --git a/bad_width_a.txt b/bad_width_b.txt
index 8b0d958..e69de29 100644
--- a/bad_width_a.txt
+++ b/bad_width_b.txt
@@ -1,5 +0,0 @@
-中文字符长度通常为2
-
-中文字符长度通x常为2
-
-abcdefghijklm j
```
- [ x] A screenshot of Delta's output is often helpful also.
![image](https://user-images.githubusercontent.com/11974052/182276667-f302237a-9ad8-4a68-b213-6f25c2b32bed.png)

`ansi::truncate_str` does not truncate str to exact width and cause bad alignment

PR #1145, opened 2022-08-02
try fix bad alignment in unicode (#1144)

SheldonNico:
Add spaces in `truncate_str` doesn't satisify the function description. So I change the function `pad_panel_line_to_width` (maybe not right)

### DIFF
--- a/src/ansi/console_tests.rs
+++ b/src/ansi/console_tests.rs
@@ -54,10 +54,11 @@ mod tests {
             &truncate_str(&s, 10, "..."),
             &format!("foo {}...", style("bar").red().force_styling(true))
         );
+        // `バ`(width = 2) will be truncate to 1, we use space to fill
         let s = format!("foo {}", style("バー").red().force_styling(true));
         assert_eq!(
             &truncate_str(&s, 5, ""),
-            &format!("foo {}", style("").red().force_styling(true))
+            &format!("foo {}", style(" ").red().force_styling(true))
         );
         let s = format!("foo {}", style("バー").red().force_styling(true));
         assert_eq!(
--- a/src/ansi/mod.rs
+++ b/src/ansi/mod.rs
@@ -52,7 +52,7 @@ pub fn truncate_str<'a, 'b>(s: &'a str, display_width: usize, tail: &'b str) -> 
             for g in t.graphemes(true) {
                 let w = g.width();
                 if used + w > display_width {
-                    // use space to fill the gap left by truncate. Need another 
+                    // use space to fill the gap left by truncate. Need another
                     // option to custom this or just use `WrapOption.wrap_right_prefix_symbol`
                     // in the future.
                     result.push_str(&" ".repeat(display_width.saturating_sub(used)));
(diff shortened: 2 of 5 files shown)


th1000s:
Thank you, great find. Indeed, the length of Chinese characters is usually 2 (do you know any unicode points which map to 3 or more cells?), but here the assumption is <= 1.

Now however `measure_text_width()` is called called twice, the first time in `truncate_str()`. The fill-up logic would be better placed in there. It would also mean that this function now returns something with the _exact_ display length if something had to be cut off, i.e. the rest is filled up. Because placing a space _after_ the `→` would look uneven, it is better to place them before or even use a blue symbol (`·→`, similar to the `↵`) to make it clear that this is not the original input.

And somewhere in the wrapping logic the same miscounting happens which results in truncation instead of wrapping, probably when `line_is_too_long()` assumes that `line.graphemes(true).count()` equals the display width. Then spaces or `·` will also have to be placed before the `WrapConfig::left_symbol` (`↵`).

Save the diff below as 'badwrap' and run:
`for i in {34..22}; do target/debug/delta --width=$i < badwrap | tail -n +4; done`

``` diff
@@ -1,5 +0,0 @@
-一二三四
-1一二三四
-12一二三四
-123一二三四
-123456789
```

SheldonNico:
> do you know any unicode points which map to 3 or more cells?), but here the assumption is <= 1.

I didn't find any Chinese character map to 3 or more. But from the code in [unicode-width](https://github.com/unicode-rs/unicode-width/blob/master/src/tables.rs#L50), 3 means ambiguous width.

 > The fill-up logic would be better placed in there. 

Yes, Add a symbol before is much better. 

I will try in my next commit.

dandavison:
Hi @SheldonNico, thanks very much for working on this. I'm sorry I let it slip and then forgot about it.

SheldonNico:
Add space(no custom symbol for simplicity) in `truncate_str` is easy. 

But change how width calculated in wrapping need too much modifications, and some modifications may not be right since I didn't figure out all function logical. I try to fix broken tests, and it seems to work. Feel free to not accept this PR, if you have plan in the future. 

image

SheldonNico:
> do you know any unicode points which map to 3 or more cells?

Yes, original test find one: [क्षि ](https://github.com/dandavison/delta/blob/master/src/wrapping.rs#L942), and [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=58c299aab7c10063c40c834666cc65fa). It break my text editor. But in browser, it take 2 columns.

th1000s:
Very nice, looks good to me. Also, you may add [this commit](https://github.com/th1000s/delta/commit/e245a8bbf54cb7a02fe792f2001d8e141ec0e3a9) which will correct the tests instead of adapting the code to work with the old tests.

> Add space(no custom symbol for simplicity)

Yes, that would have to become customizable. I will do that later when making the wrapping aware of word boundaries.

SheldonNico:
Done merge. 

As I notice, there are some [updates](https://github.com/dandavison/delta/commit/970000f36886bdab908d1037dbc1148abde52d5e). Do I need do something in my branch?

dandavison:
Thank you very much @SheldonNico and @th1000s. This will be released today.

PR #1158, opened 2022-08-11
Support configurable timestamps in git blame output (#1157) 

### LINKED ISSUE #1157: 🚀 Configurable timestamp output format in git blame
In git blame I noticed this:
image

It would be helpful to know e.g. which change was applied first but this can not be figured out from the humanized timestamp.
https://github.com/dandavison/delta/blob/738c5a141bf86d3e05aa348eb87df78c36a72337/src/handlers/blame.rs#L273

I changed it locally to: `blame.time.format("%H:%M %Y-%m-%d").to_string(),`.

Would it be possible to make this configurable? Perhaps a new option *--blame-**output**-timestamp-format* could be added. If present (and not empty) it would be used as a format string. If not, we will fall back to HumanTime formatting. What do you think?

I can work on a PR if this could be accepted.

mliszcz:
New CLI/config option is introduced: blame-timestamp-output-format.

Fixes #1157.


mliszcz:
@dandavison could you please have a look at clippy output? I believe my code does not trigger any of those:
```
error: you are deriving `PartialEq` and can implement `Eq`
```

I noticed the same errors in two other recent PRs:
https://github.com/dandavison/delta/actions/workflows/ci.yml

I do not see any clippy errors when running locally with:
```
cargo 1.62.1 (a748cf5a3 2022-06-08)
rustc 1.62.1 (e092d0b6b 2022-07-16)
clippy 0.1.62 (e092d0b 2022-07-16)
```

mliszcz:
Okay, it turns out that these clippy errors appear after upgrade to rust 1.63.0. There was a release on Thursday last week and since the CI pipeline is using 'stable' channel it was impacted.

dandavison:
Hi @mliszcz, your changes look great -- sorry for being slow to review properly, I'm a bit swamped with the day job as usual.

And thanks for looking into the clippy errors. You don't need to address them (but of course if you feel like fixing them, in this PR or a separate one, you would be very welcome!).

mliszcz:
Thanks for having a look. I will fix these clippy errors later today in a new commit in this PR. This should be straightforward, I can see that clippy already suggest fixes for most of them.

dandavison:
Thanks very much for this work @mliszcz. I didn't see any improvements to suggest in your blame changes. And thanks very much for fixing (and explaining) the clippy issues.

PR #1165, opened 2022-08-19
Added auto completions for fish shell

exploide:
I added auto completions for the fish shell.

On the next release, it is probably a good idea to take a note for packagers that this file should be delivered as `/usr/share/fish/vendor_completions.d/delta.fish`.


dandavison:
Thanks a lot @exploide! I might try to test this myself in the future but it looks great and I'm going to merge it now so that it's available to every others.

dandavison:
https://github.com/dandavison/delta/pull/1561 makes use of the shell completion autogeneration feature in Rust's [clap](https://github.com/clap-rs/clap/tree/master/clap_complete) library. It's output looks similar to yours, so I'm going to delete all the shell completion scripts that are checked into the repo and instead encourage users to use the output of `delta --generate-completion fish`. Please let me know if you don't think that's the right course. And thanks a lot for doing this in the first place!

PR #1187, opened 2022-09-13
Switch bat to library mode

tranzystorek-io:
This disables bat dependencies that are only required for the `bat` binary, e.g. clap v2. Should cut down a bit on the build time.


dandavison:
Awesome, thanks for this @tranzystorek-io, I see lots of nice red in Cargo.lock.

PR #1191, opened 2022-09-14
Refactoring ansi/iterator

zhiburt:
Hi there
As I was looking through the file and I decided to make it a bit cleaner.

No actual changes are supposed to be done.
Except her it was unsound that we are not doing anything with old element but I decided to make it explicit `None`. 

https://github.com/zhiburt/delta/blob/22ff23d588a9b483b9457f8df6b414f59122b7e1/src/ansi/iterator.rs#L134

Take care

### DIFF
--- a/src/ansi/iterator.rs
+++ b/src/ansi/iterator.rs
@@ -24,6 +24,7 @@ pub struct AnsiElementIterator<'a> {
     pos: usize,
 }
 
+#[derive(Default)]
 struct Performer {
     // Becomes non-None when the parser finishes parsing an ANSI sequence.
     // This is never Element::Text.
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks a lot! I really like the look of this. I'll look properly this evening.

dandavison:
I agree with you that that code needed cleaning up!

Incidentally, there's some `ansi_style` stuff in there which I think may be appropriate for delta but probably we should make it optional in the crate. If it improves benchmarking times to remove it we may want to do that sooner rather than later.

dandavison:
This is much, much nicer than the code it's replacing. Thanks a lot. I believe it also benchmarks slightly faster (3-5%? I was using `make benchmark`). Do you have time to also apply these changes to https://github.com/dandavison/vte-ansi-iterator, so that your authorship is attached to them?

Issue #1176, opened 2022-08-31
🚀 wrap code of `git blame` in block of code only


timsu92:
Currently, there's no wrap to the code. So, if the code is too long in width, it will be wrapped by terminals to the next line, which is in the same place as `--blame-format`.

I hope it can wrap the codes in the right of line numbers.

timsu92:
True.
However, that targets `git diff`, which already has a working solution(with some bug) at least in side-by-side view.

dandavison:
What's the bug you're referring to @timsu92?

timsu92:
I haven't searched for exist issue yet. I'll search for that or raise a new one.

PR #1194, opened 2022-09-16
Add codeberg link parsing

p00f:


dandavison:
OK, merged this, thanks very much!

> (Can the "configuration" for the major ones i.e. sourcehut and codeberg be built-in along with github and gitlab?)

Sure. But, actually, I wonder whether maybe this is more work than it's worth. Can you think of a good configuration interface/API for users to specify this parsing logic? We have `--file-transformation` where users supply a sed-style transformation, but it would need to be more complicated than that. I'm thinking perhaps there are more important things to work on in delta (e.g. line wrapping in the non-side-by-side cases #1176 #299)

p00f:
In #1176 and #299 the issue is that blame and non-side-by-side are not wrapped, correct?

PR #1205, opened 2022-10-05
Ensure tabs always end at a multiple of the tab width

th1000s:
Plus clippy warnings, but this pushes the minimally supported rust version to 1.62 now via [`then_some()`](https://doc.rust-lang.org/std/primitive.bool.html#method.then_some).


dandavison:
Hi @th1000s, nice to see you.

I'm getting a crash on this branch with the diff below, but not when piping explicitly into delta -- only when allowing git to invoke delta. Can you reproduce that? (The diff introduces one tab character on l.16)

```diff
diff --git a/Makefile b/Makefile
index 6176427..a48a901 100644
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ unit-test:
 	cargo test
 
 end-to-end-test: build
-	./tests/test_raw_output_matches_git_on_full_repo_history
+		./tests/test_raw_output_matches_git_on_full_repo_history
 	./tests/test_deprecated_options > /dev/null
 	./tests/test_navigate_less_history_file
 
```

```
thread 'main' panicked at 'String mismatch encountered while superimposing style sections: ' ' vs '.'', src/paint.rs:949:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/panicking.rs:142:14
   4: delta::paint::Painter::paint_lines
   5: delta::paint::paint_minus_and_plus_lines
   6: delta::paint::Painter::paint_buffered_minus_and_plus_lines
   7: delta::handlers::hunk::<impl delta::delta::StateMachine>::handle_hunk_line
   8: delta::delta::delta
   9: delta::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

th1000s:
Not reproducible for me, neither with debug nor release (nor full re-)builds using Rust 1.64.0. Maybe it is part of your `~/.gitconfig` or the repo config? Running it with `HOME=/ git ...` here also works fine.

In `prepare_raw_line()` the order of adding a newline and calling `ansi_preserving_slice()` is flipped, but both result in the same string with a trailing newline. When using this code path via e.g. `git diff --word-diff=color` everything works as expected.

dandavison:
Hi @th1000s, sorry to be slow here. You're right -- it was dependent on gitconfig. The crash can be reproduced by setting

```
[diff]
    colorMovedWS = allow-indentation-change
```

and viewing this diff:

```diff
diff --git a/Makefile b/Makefile
index 6176427..a48a901 100644
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ unit-test:
 	cargo test
 
 end-to-end-test: build
-	./tests/test_raw_output_matches_git_on_full_repo_history
+		./tests/test_raw_output_matches_git_on_full_repo_history
 	./tests/test_deprecated_options > /dev/null
 	./tests/test_navigate_less_history_file
 
```

although, apparently not when piping to delta; only when the delta process is started by git.

th1000s:
Variable tab expansion is quite tricky, and I have to ask: Is this worth it? Most tabs in code will be leading ones which work fine with the fixed expansion method, and tabstops will continuously cause quite a bit of work in the future since they concern two different abstraction layers.

What do you think @nickelc @wescande @tjquillan @MarcoIeni? Do you want to keep maintaining this @dandavison? :)

Also, in some of the grep cases the non-fixed tab expansion can be used. And more tests with raw inputs are needed.

MarcoIeni:
I don't maintain or use delta. Please don't ping me unless you need help with ci, thanks :)

dandavison:
> Is this worth it?

I am finding this hard to answer: I never work in repos with tabs, except for in Makefiles, and there they are always at the beginning of lines. But that's not to say I don't want delta to work well with repos where tabs are used differently.

CharlemagneLasse:
> Is this worth it?

If you ever worked in Linux kernel or related code bases - definitely worth it. Let us count relevant lines in linux-next 20240216:

```
$ git grep '[^'$'\t'']'$'\t' |wc -l
2293788
```

dandavison:
> I see, raw ansi codes produced by colorMovedWS or colorMoved cause this

I'm sorry @th1000s -- I think that you have fixed this bug (right?), but when I read

> Variable tab expansion is quite tricky, and I have to ask: Is this worth it?

I mistakenly concluded that you had not yet fixed it. So is there anything more needed on this branch, or was it ready to merge almost a year ago?

chtenb:
Apart from tab-indented codebased, TSV files is another case this comes up.

PR #1216, opened 2022-10-24
measure_text_width() without constructing a temporary string

th1000s:
The improvement probably can't be measured, but still feels right :)


dandavison:
Agreed, thanks! cc @zhiburt who is using code from this module in [ansitok](https://gitlab.com/zhiburt/ansitok/) and [tabled](https://github.com/zhiburt/tabled) and hence it is being used in display of all tabular data by [nushell](https://github.com/nushell/nushell). (There's been an emphasis on performance in that work.)

Issue #1114, opened 2022-06-28
Doesn't work with git version `2.37.0`


boris-petrov:
Using delta 0.13.0. I have this in my `.gitconfig` file:

```
[interactive]
	diffFilter = delta --color-only
```

Which worked great when using `git add -p`. Since version 2.37.0 of git, it blows up:

```
error: could not parse colored hunk header
```

And then some random stuff.

OJFord:
Workaround:
```ini
[add.interactive]
    useBuiltin = false
```

(Credit to @malsyned in issue linked above)

phil-blain:
I would say it is a Git regression, and should be reported to the Git mailing list: http://git-scm.com/community.

phil-blain:
> I would say it is a Git regression, and should be reported to the Git mailing list: http://git-scm.com/community.

Reported here: https://lore.kernel.org/git/ecf6f5be-22ca-299f-a8f1-bda38e5ca246@gmail.com/T/#u

boris-petrov:
Indeed it is! Thanks!

WhyNotHugo:
Fixed for me too. Thanks for following up!

PR #1228, opened 2022-11-05
Remove Git 2.37 workaround from install docs

adamchainz:
The bug was fixed in Git 2.38, as noted on original issue: https://github.com/dandavison/delta/issues/1114#issuecomment-1265920597

I guess we can remove this from the docs now, so new users don’t switch to the old interactive `add`.

I thought of adding a section “If you’re using Git 2.37...” but that seemed too specific. If anyone encounters the error, hopefully they’ll search for it and see they need to upgrade Git.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,6 @@
 
 [interactive]
     diffFilter = delta --color-only
-[add.interactive]
-    useBuiltin = false # required for git 2.37.0
 
 [delta]
     navigate = true    # use n and N to move between diff sections


adamchainz:
It also seems the workaround was never added to the docs: https://github.com/dandavison/delta/blob/master/manual/src/get-started.md

Perhaps removing the GitHub README and pointing only to the docs site would be a good idea, to reduce redundant redundancy (that is redundant).

dandavison:
Thanks @adamchainz! I agree. I think I do like having the quick start instructions in the Github README landing page so that people who are used to installing and configuring things like this can do it very quickly without any clicking around.

adamchainz:
Thanks for the fast review. Fair enough on keeping on GitHub.

PR #1238, opened 2022-11-16
ci: improve formatting

MarcoIeni:
It looks better like this. What do you think?
It's the same syntax used [here](https://github.com/rust-github/rust-gh-example/blob/main/.github/workflows/cd.yml).

EDIT:
- [x] apply it also in ci.yaml


dandavison:
Agreed, thanks!

PR #1239, opened 2022-11-16
ci: release apple arm binary

MarcoIeni:
Requested [here](https://github.com/dandavison/delta/issues/484#issuecomment-1312790539)

There are no VMs for GitHub actions that support Mac ARM. See https://github.com/actions/runner-images/issues/2187.
So we cannot run tests for MAC ARM. However, we can run `cargo build`, which is what's required to build the binary.

- [x] to be merged after #1238

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -15,6 +15,9 @@ jobs:
           - os: macos-latest
             target: x86_64-apple-darwin
             use-cross: false
+          - os: macos-latest
+            target: aarch64-apple-darwin
+            use-cross: false
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             use-cross: false
@@ -96,7 +99,7 @@ jobs:
         with:
           command: publish
           args: --token ${{ secrets.CARGO_API_KEY }} --allow-dirty
-          
+
   bump-homebrew-formula:
     runs-on: macos-latest
     steps:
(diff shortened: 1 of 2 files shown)


MarcoIeni:
```
Caused by:
  Bad CPU type in executable (os error 86)
```

I need to fix this.

.github/workflows/cd.yml
```diff
@@ -12,19 +12,37 @@ jobs:
...
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-gnu
+            use-cross: true
 
     steps:
       - name: Installing Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+          target: ${{ matrix.job.target }}
```
MarcoIeni, line 45: This installs the target if missing. Useful for `aarch64-apple-darwin` on mac

MarcoIeni:
This is ready now. 👍

dandavison:
Hi @MarcoIeni, thanks for this work!

So basically, we add `aarch64-apple-darwin` to cd but not to ci. Is that an accurate summary?

I've merged your reformatting PR so I'll wait for you to rebase this one.

MarcoIeni:
> So basically, we add aarch64-apple-darwin to cd but not to ci. Is that an accurate summary?

Yes, exactly

dandavison:
Thank you very much @MarcoIeni!

PR #1244, opened 2022-11-25
Highlighting improvements

phillipwood:
While using delta I have noticed a few cases where the highlighting
was not quite as good as it could have been. This PR attempts to
address that. The most significant change is that where possible
insertions and deletions are grouped together to minimize the number of
groups of changes within a line. For example the change
    
    -printf "%s\n" s y y ...
    +test_write_lines s y n ...

is now highlighted as

    -[printf "%s\n"] s y n ...
    +[test_write_lines] s y y ...

rather than

    -[printf "%]s[\n"] [s ]y y ...
    +[test_write_lines ]s y y ...

There are several other changes:
 - Unchanged tokens are less likely to be highlighted when the length of
the line is unchanged.
 - Where possible deletions will now precede insertions.
 - An unchanged space at the end of a line will not be highlighted as
changed when it is preceded by a changed word.

I was unsure whether to squash the last two commits together. The
diffstat of the combined patch is quite a bit smaller than the two
separate patches but I think it is clearer to show the changes
separately.

### DIFF
--- a/src/align.rs
+++ b/src/align.rs
@@ -1,10 +1,6 @@
 use std::cmp::max;
 use std::collections::VecDeque;
 
-// Cost invariants:
-//    SUBSTITUTION_COST < DELETION_COST + INSERTION_COST
-//    DELETION_COST + INSERTION_COST + INITIAL_MISMATCH_PENALITY < 2 * SUBSTITUTION_COST
-const SUBSTITUTION_COST: usize = 3;
 const DELETION_COST: usize = 2;
 const INSERTION_COST: usize = 2;
 // extra cost for starting a new group of changed tokens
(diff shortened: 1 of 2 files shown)


phillipwood:
@dandavison Thanks, I'll try and take a look at the issues this week. I'm not sure what to do with `max-line-distance`. I've been using it unchanged and that seems to give reasonable results though there is some change it what gets highlighted and what does not. The screenshots below show the diff in the output between `master` and this branch when viewing the history of git (reading diffs of diffs takes a bit of getting used to but it is a convenient way to find changes in the output).

A couple of changes that get highlighted when they weren't before
Screenshot from 2022-11-26 15-30-18
image

A couple of examples where we lose the highlighting
image
image

Here are a couple examples of improvements
image
image

This is an improvement but I'm pretty sure it is just luck
image

I've not seen anything terrible but these are slightly worse now. In the future I'd like to play with sliding the highlights to optimize their position in the same way git's diff.indentHeurstic does for diffs.
image
image

nestor-custodio:
I know the v0.15.0 release is a few days old at this point, but the highlighting improvement is *so noticeable* I just had to come back to this issue and say **WOW**. Thank you for making these changes!

PR #1245, opened 2022-11-27
Add .gitattributes rules for rust files

phillipwood:
Setting `diff=rust` will make git show the current function name in the hunk headers.

While delta currently removes git's whitespace error markup from diffs adding the whitespace rules will mean pre-commit hooks that run `git diff --check` will reject changes with whitespace errors.  In order to get git to print the correct position of the error when there is an unwanted tab in the indent following a space we need to turn off `space-before-tab` otherwise the space will be considered to be an error instead of the tab.

The whitespace rule could perhaps be applied more widely but the Makefile and example gitcofig have leading tabs so I restricted them to rust files for now.

[edited to explain `-space-before-tab`]

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.Makefile linguist-language=Makefile
 etc/performance/* linguist-vendored
+*.rs whitespace=tab-in-indent,trailing-space,-space-before-tab eol=lf


dandavison:
Thanks!

I haven't spent much time with this, but `diff=rust` isn't always an improvement is it? E.g. here (before on the left, after on the right): The current version indicates that the function `ingest_line` is a method in the `StateMachine` `impl`, whereas with `diff=rust` it erroneously suggests that the change is in a sibling function above.


image

phillipwood:
Oh dear, I'd forgotten that it could do that (there was a [discussion](https://lore.kernel.org/git/20200923215859.102981-1-rtzoeller@rtzoeller.com/) about fixing this on the git mailing list a while ago but nothing came of it). On the whole I think it is an improvement as it generally shows the function being changed rather than the enclosing struct or mod (see below, before on the left, after on the right) but if you're not keen then lets drop `diff=rust` and just keep the whitespace rules. The case you saw only happens if the real function name ends up in the context lines above the change.
Untitled

phillipwood:
I've removed the diff attribute leaving just the whitespace attributes sorry it has taken me so long to get around to it.

dandavison:
Thanks!

PR #1249, opened 2022-12-03
Explicitly request xz compression

### LINKED ISSUE #1248: 🐛 Debian 11 dpkg package doesn’t support zstd compression.
```sh
sudo dpkg -i /home/user/software/git-delta_0.15.0_amd64.deb 
dpkg-deb: error: archive '/home/user/software/git-delta_0.15.0_amd64.deb' uses unknown compression for member 'control.tar.zst', giving up
dpkg: error processing archive /home/user/software/git-delta_0.15.0_amd64.deb (--install):
 dpkg-deb --control subprocess returned error exit status 2
Errors were encountered while processing:
 /home/user/software/git-delta_0.15.0_amd64.deb
```

dandavison:
Fixes #1248

### DIFF
--- a/etc/ci/before_deploy.sh
+++ b/etc/ci/before_deploy.sh
@@ -162,7 +162,7 @@ Description: Syntax highlighter for git
  detection, and restructured diff output for git on the command line.
 EOF
 
-    fakeroot dpkg-deb --build "$tempdir" "${dpkgname}_${version}_${architecture}.deb"
+    fakeroot dpkg-deb -Z xz --build "$tempdir" "${dpkgname}_${version}_${architecture}.deb"
 }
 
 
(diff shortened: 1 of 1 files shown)


etc/ci/before_deploy.sh
```diff
@@ -162,7 +162,7 @@ Description: Syntax highlighter for git
  detection, and restructured diff output for git on the command line.
 EOF
 
-    fakeroot dpkg-deb --build "$tempdir" "${dpkgname}_${version}_${architecture}.deb"
+    fakeroot dpkg-deb -Z xz --build "$tempdir" "${dpkgname}_${version}_${architecture}.deb"
```
baryluk, line 165: Please use `-Zxz` without a space.

This is what man page says. (`-Z xz` works - I just tested it, but is not explicitly documented).

dandavison: OK, changed! Personally I have always inserted a space after a single-letter flag for every Unix/Linux command I've used in the last 20 years as far as I remember -- I think a Linux command would be pretty broken if it didn't accept a space. Plus one has to remember that Java and Windows people for some reason think it's acceptable to use a single `-` with the long option variants :)

PR #1260, opened 2022-12-14
Recommend zdiff3 merge.conflictStyle

adamchainz:
Compared to diff3, zdiff3 minimizes the conflicted region, so fewer lines will show as conflicted. This normally makes merge conflicts easier to resolve. Docs: https://git-scm.com/docs/git-config#Documentation/git-config.txt-mergeconflictStyle .

I also made `conflictStyle` follow Git’s casing, as other options are documented like that as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@
     # light = true
 
 [merge]
-    conflictstyle = diff3
+    conflictstyle = zdiff3
 ```
 
 Delta has many features and is very customizable; please see the [user manual](https://dandavison.github.io/delta/).
--- a/manual/src/get-started.md
+++ b/manual/src/get-started.md
@@ -17,5 +17,5 @@
     # light = true
 
 [merge]
-    conflictstyle = diff3
+    conflictstyle = zdiff3
 ```
--- a/manual/src/merge-conflicts.md
+++ b/manual/src/merge-conflicts.md
@@ -1,13 +1,14 @@
 # Merge conflicts
 
-Consider setting
+Consider setting [`merge.conflictStyle`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-mergeconflictStyle) to `zdiff3`:
 
 ```gitconfig
 [merge]
-    conflictstyle = diff3
+    conflictStyle = zdiff3
 ```
 
-With that setting, when a merge conflict is encountered, delta will display diffs between the ancestral commit and each of the two merge parents:
+With that setting, when a merge conflict is encountered, Git will display merge conflicts with the contents of the merge base as well.
+delta will then display this as two diffs, from the ancestor to each side of the conflict:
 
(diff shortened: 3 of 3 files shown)


th1000s:
The option zdiff3 is rather new, added in 2.35 from last year. So this change would break systems with older git versions.

adamchainz:
True - Git 2.35 was [released 2022-01-24](https://lore.kernel.org/git/xmqqee4x3pij.fsf@gitster.g/T/#u). But I'm not sure a >1 year old version is such a concern... Users on such old versions also have many unpatched CVEs, since Git doesn't backport security fixes too far 🤷‍♂️

simono:
@dandavison Would you consider to merge this change? Thanks 😄

dandavison:
100% happy to add the recommendation to the manual; no hesitation there (does anyone want to come up with good "before and after" examples/screenshots showing the advantage?).

As for the front-page recommended config, we don't have an explicit policy yet on compatibility with supported Git versions. I think could be viewed as a statistical question: where does the Git version that this requires fall in the distribution of Git versions used by delta users? Which leads to the question: what version of Git do the most common operating systems ship with or in the cases where people get git from package managers what do they get? (I don't think CVEs in Git are relevant; people use whatever version of Git they have.)

simono:
Maybe we can link to [this GitHub blog post](https://github.blog/open-source/git/highlights-from-git-2-35/#:~:text=If%20you%E2%80%99ve%20ever%20dealt%20with%20a%20merge%20conflict), which explains zdiff3 pretty well?

`zdiff3` was introduced in Git 2.35 (Jan 2022). I think it should be shipped with all major operating systems and package managers by now.

Some examples:
macOS has git 2.39 since at least Version 13 (Ventura).
Debian has git 2.39 since Version 12 (Bookworm), with backports available for older systems.
Ubuntu has git 2.40 since Version 23.10 (Mantic Minotaur).

dandavison:
Thanks @simono, I read the blog post, but I admit I was waiting until I could personally started noticing the improvements in my own workflows.

This may be old news to everyone else but for the record, here's an example. It's simpler than the one in the GitHub blog -- in a way there is no conflict, just two adjacent changes. So this one's  bit improvement because under zdiff3 in my Git UI (magit/vscode/lazygit etc) I can just stage it without resolving anything, whereas diff3 would require me to choose a resolution.

### with diff3:
image



### with zdiff3:

image

ccoVeille:
zdiff3 was a game changer for me

dandavison:
@th1000s do you think enough time has passed to put this in the example config that people will blindly copy and paste, or should we just add it as a recommendation for now?

th1000s:
Okay, on an incompatible setting it causes the next git pull/rebase command to fail, even if there was no conflict: `fatal: unknown style 'zdiff3' given for 'merge.conflictstyle'`. For LTS or other users who use older versions the cause should be obvious.

dandavison:
OK, thanks all! Merging this after a 2-year docs PR review process :)

adamchainz:
Yay, thanks!

PR #1291, opened 2023-01-27
Add `gruvmax-fang` theme

MaxFangX:
I spent more time than I intended customizing my own diff output and it turned out really nice so I figured I might as well upstream it. Inspired by [gruvbox](https://github.com/morhetz/gruvbox), `bat`'s default diff, and the existing `delta` themes.

### Split diff

Git show with split diff

### Unified diff

Unified diff

### Git blame

Git blame 3

### Merge conflicts

Merge conflicts


dandavison:
Looks really nice, thanks!

PR #1315, opened 2023-02-27
Reduce allocations during `--show-colors`

nickelc:
Using `BTreeMap` makes the additional allocation for sorting unnecessary.

### DIFF
--- a/src/colors.rs
+++ b/src/colors.rs
@@ -1,7 +1,5 @@
-use std::collections::HashMap;
-
-pub fn color_groups() -> HashMap<String, Vec<(String, String)>> {
-    [
+pub fn color_groups() -> Vec<(&'static str, Vec<(&'static str, &'static str)>)> {
+    vec![
         (
             "Blue",
             vec![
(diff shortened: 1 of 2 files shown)


dandavison:
Thanks! I think perhaps I just wasn't familiar with how to use `&'static str`, hence the ugly conversion to `String`.

Given that we're just iterating over them do we actually need a map data structure at all? Should we just use a Vec?

Also, out of interest, is there a reason the `BTreeMap` can be used with `&'static str` but the HashMap could not?

There are some Clippy failures -- let me know if you'd like to fix them or else I can.

nickelc:
> Given that we're just iterating over them do we actually need a map data structure at all? Should we just use a Vec?

yes you could, since the colors are already sorted.

> Also, out of interest, is there a reason the BTreeMap can be used with &'static str but the HashMap could not?

The `BTreeMap` is a sorted map and keeps the colors ordered like before with `Itertools::sorted`.

> There are some Clippy failures -- let me know if you'd like to fix them or else I can.

I've opened a PR to clippy warnings #1298

dandavison:
> yes you could, since the colors are already sorted.

OK thanks. Would you mind changing it to `Vec` seeing as we're perfecting this area of the code?

> I've opened a PR to clippy warnings https://github.com/dandavison/delta/pull/1298

Oops, sorry I missed that! My day job has been making it hard to work on delta.

dandavison:
Thanks! Merged.

PR #1320, opened 2023-03-01
Fix the`--show-themes` command

nickelc:
The command failed because the two empty arguments for the internal `Opt` parsing actually meant "<binary> <minus_file>" and since `minus_file` is [now parsed with `PathBufValueParser`](https://github.com/dandavison/delta/commit/5cef3872999f1084cb13013b002a6421c4aa3f15) and the value parser rejects empty values.


dandavison:
Thanks!

PR #1253, opened 2022-12-04
Update to clap 4

tjquillan:
This updates delta to `clap` version `4.0.29`. With all the migrations that come along with it. I am leaving this as a draft for now as I still need to make the correct updates to the code. I will list the changes and important parts of `4.0.0` here as I go. I will try to do each change as a single commit so they can be dropped or changed as needed.

Changes:
* Add clap features `help,usage,error-context`
    * [If `default-features = false`, run `cargo add clap -F help,usage,error-context`](https://github.com/clap-rs/clap/blob/35f4597ce2c039f44a98628458196f4d6d67b3f2/CHANGELOG.md?plain=1#L320) 
*  Add clap feature `wrap_help`
    * [Run `cargo add clap -F wrap_help` unless you want to hard code line wraps](https://github.com/clap-rs/clap/blob/35f4597ce2c039f44a98628458196f4d6d67b3f2/CHANGELOG.md?plain=1#L321)
* Remove `DeriveDisplayOrder` as it is now the default
    * [Make `DeriveDisplayOrder` the default and removed the setting.](https://github.com/clap-rs/clap/blob/35f4597ce2c039f44a98628458196f4d6d67b3f2/CHANGELOG.md?plain=1#L389) 
* Update `ValueSource` using statement
    * [`clap::{Indices, OsValues, ValueSource, Values}` to `clap::parser::{Indices, OsValues, ValueSource, Values}`](https://github.com/clap-rs/clap/blob/35f4597ce2c039f44a98628458196f4d6d67b3f2/CHANGELOG.md?plain=1#L635)
* Update `#[clap(...)]` to `#[command(...)]` where appropriate
    * Attribute `#[clap(...)]` has been deprecated in favor of `#[command(...)]` 
* Update `#[clap(...)]` to `#[arg(...)]` where appropriate
    * Attribute `#[clap(...)]` has been deprecated in favor of `#[arg(...)]`
* Update `#[structopt(...)]` to `#[arg(...)]` where appropriate
    * Attribute `#[structopt(...)]` has been deprecated in favor of `#[arg(...)]`
* Remove `value_parser` because it is now default
    * [Changed the default for arguments from `parse` to `value_parser`](https://github.com/clap-rs/clap/blob/35f4597ce2c039f44a98628458196f4d6d67b3f2/CHANGELOG.md?plain=1#L434)

PR #1322, opened 2023-03-02
Update `clap` to 4.1

nickelc:
This PR is a continuation of #1253 

Thanks to @tjquillan

### DIFF
--- a/src/options/set.rs
+++ b/src/options/set.rs
@@ -22,9 +22,9 @@ macro_rules! set_options {
     $opt:expr, $builtin_features:expr, $git_config:expr, $arg_matches:expr, $expected_option_name_map:expr, $check_names:expr) => {
         let mut option_names = HashSet::new();
         $(
-            let kebab_case_field_name = stringify!($field_ident).replace("_", "-");
-            let option_name = &$expected_option_name_map[&kebab_case_field_name];
-            if !$crate::config::user_supplied_option(&kebab_case_field_name, $arg_matches) {
+            let field_name = stringify!($field_ident);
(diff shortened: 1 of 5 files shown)


src/options/set.rs
```diff
@@ -88,7 +89,7 @@ pub fn set_options(
 
     // --color-only is used for interactive.diffFilter (git add -p) and side-by-side cannot be used
     // there (does not emit lines in 1-1 correspondence with raw git output). See #274.
-    if config::user_supplied_option("color-only", arg_matches) {
+    if config::user_supplied_option("color_only", arg_matches) {
```
dandavison, line 92: How sure are we that we've updated all places to snake-case that need updating?

nickelc: All usage of the `ArgMatches` seems to go through the `options::set::set_options` function, the `set_options!` macro and `config::user_supplied_option` function.
And the macro checks the names against the `expected_option_name_map` hash map.

dandavison:
It looks like we've lost the colours from `delta -h` and `delta --help`. I liked them, but there was a drawback which was that the ANSI sequences remain even when the help text was being piped/redirected, so e.g. if redirecting to less, the user had to know to use `less -R`.

WDYT? ANy chance clap v4 has made the colors go away when stdout is not a tty?

nickelc:
> Any chance clap v4 has made the colors go away when stdout is not a tty?

Clap's v4 release blog post explains the reasons for this change and plans for customization.
https://epage.github.io/blog/2022/09/clap4/#polishing-help-output

dandavison:
I see, thanks for explaining. Really useful work here, thanks.

PR #1323, opened 2023-03-02
Update `git2` to 0.16.1

nickelc:
`git2::ConfigEntries` no longer implements `Iterator` since `git2` 0.15.


dandavison:
Thanks you!

PR #1324, opened 2023-03-04
Added config option.

### LINKED ISSUE #1301: 🚀 Separate configuration file and multiple profiles
Hello all!
First of all, thanks for this awesome tool which I love every day.
In very humble opinion there are two features that are missing:

## External configuration file

Rather than storing all options in `.gitconfig`, can we store the configuration inside an external file and just reference that from the .gitconfig?

Like:

```
# .gitconfig
[delta]
  config = /path/to/config
```

and 

```sh
delta --config=/path/to/config
```

I don't have any preference for the config format, it might be the same gitconfig format, TOML or YAML.

## Support multiple profiles

Following the proposal above (but given we use environment variable it is technically independent of it), we might add a `profile` option which specifies which section to load. For instance:

```
# .gitconfig

[delta]
  config = /path/to/config
  profile = main
``` 

and 

```sh
delta --config=/path/to/config --profile=main
```

will look for section `delta.main` rather than simply `delta`.

What do you think about it?

ShogunPanda:
This PR adds two new options:

1. `--config <PATH>`: Can be used to load a config file different from `./.gitconfig`. The loaded file MUST use the same format.
2. `--features-first`: (bool) When enabled, this options gives features priority towards the default delta section. For instance, if we have the following `.gitconfig` file:

    ```
    [delta]
    pager = DEFAULT
   
    [delta "feature"]
    pager = FEATURE
    ```

    With the current delta (or, if this PR lands, by default) the final resolved value for the pager, even if the feature `feature` is enabled) would be `DEFAULT`. With the new option, the final resolved values becomes `FEATURE`.

Fixes #1301.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -55,9 +55,9 @@ A feature name may not contain whitespace. You can activate multiple features:
 
 If more than one feature sets the same option, the last one wins.
 
-If a value is present in the [delta] section, then features are not considered at all.
+If an option is present in the [delta] section, then features are not considered at all.
 
-If you want a value to be fully overridable by a feature and also have a non default value when
+If you want an option to be fully overridable by a feature and also have a non default value when
 no features are used, then you need to define a \"default\" feature and include it in the main
 delta configuration.
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -55,6 +55,23 @@ A feature name may not contain whitespace. You can activate multiple features:
 
 If more than one feature sets the same option, the last one wins.
 
+If a value is present in the [delta] section, then features are not considered at all.
+
+If you want a value to be fully overridable by a feature and also have a non default value when
+no features are used, then you need to define a \"default\" feature and include it in the main
+delta configuration.
+
+For instance:
+
+[delta]
+feature = default-feature
+
+[delta \"default-feature\"]
+width = 123
+
+At this point, you can override features set in the command line or in the environment variables
+and the \"last one wins\" rules will apply as expected.
+
 STYLES
 ------
 
@@ -275,6 +292,10 @@ pub struct Opt {
     /// But color and highlight hunk lines according to your delta configuration. This is mainly
     /// intended for other tools that use delta.
     pub color_only: bool,
+
+    #[arg(long = "config", default_value = "", value_name = "PATH")]
+    /// Load the config file at PATH instead of ~/.gitconfig.
+    pub config: String,
 
     #[arg(
         long = "commit-decoration-style",
@@ -1135,7 +1156,16 @@ impl Opt {
         git_config: Option<GitConfig>,
         assets: HighlightingAssets,
     ) -> Self {
-        Self::from_clap_and_git_config(env, Self::command().get_matches(), git_config, assets)
+        let mut final_config = git_config;
+        let matches = Self::command().get_matches();
+
+        if let Some(path) = matches.get_one::<String>("config") {
+            if !path.is_empty() {
+                final_config = Some(GitConfig::try_create_from_path(&env, path));
+            }
+        }
+
+        Self::from_clap_and_git_config(env, matches, final_config, assets)
     }
 
     pub fn from_iter_and_git_config<I>(
--- a/src/git_config/mod.rs
+++ b/src/git_config/mod.rs
@@ -5,7 +5,6 @@ pub use git_config_entry::{GitConfigEntry, GitRemoteRepo};
 use crate::env::DeltaEnv;
 use regex::Regex;
 use std::collections::HashMap;
-#[cfg(test)]
 use std::path::Path;
 
 use lazy_static::lazy_static;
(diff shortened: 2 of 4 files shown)


src/git_config/mod.rs
```diff
@@ -64,11 +63,40 @@ impl GitConfig {
         }
     }
 
+    #[cfg(not(test))]
+    pub fn try_create_from_path(env: &DeltaEnv, path: &String) -> Option<Self> {
```
dandavison, line 67: WDYT about just returning `Self`? This function doesn't really return an `Option` since it never return `None`.

src/cli.rs
```diff
@@ -1135,6 +1143,20 @@ impl Opt {
...
+                return Self::from_clap_and_git_config(
+                    env,
+                    Self::command().get_matches(),
+                    parsed_config,
+                    assets,
+                );
+            }
+        }
+
         Self::from_clap_and_git_config(env, Self::command().get_matches(), git_config, assets)
```
dandavison, line 1160: Nit: if you can find a nice way of doing it I'd prefer to avoid duplicating the

```
Self::from_clap_and_git_config(env, Self::command().get_matches(), git_config, assets)
```

call and instead compute `git_config` either from the custom file or accepting the passed-in value.

ShogunPanda:
I'll address the code suggestions this evening. Thanks for the hints: I'm not a Rustacean yet but I'm trying to ;)

About the feature-first flag, I might just be me misunderstanding features at all.
When I looked at them I kinda interpreted as a way to "override" the main delta section. So basically the main delta was supposed to behave as a "user default" and a feature might override it.

But apparently the delta section is always the source of truth if present and features can only backfill it.

My new option instead enables the way I initially interpreted. The way I find is useful is when in the gitconfig you set a value in the delta section and eventually you want to overwrite or "unset" (in the case I mean overwrite by setting the original default value) in a feature and you don't want to maintain separate files.

ShogunPanda:
Moreover, one important remark: the new option is a opt-in (and disabled by default ) so existing users are not affected at all.

dandavison:
> When I looked at them I kinda interpreted as a way to "override" the main delta section. So basically the main delta was supposed to behave as a "user default" and a feature might override it.
> But apparently the delta section is always the source of truth if present and features can only backfill it.

I think you're right that behavior here is a little bit surprising, but there is a way of doing what you want that I and others use. Basically, this doesn't work (as you know):

```gitconfig
[delta]
file-added-label = from default section

[delta "my-feature"]
file-added-label = from my feature section
```

But this works:

```gitconfig
[delta]
features = my-defaults-feature

[delta "my-defaults-feature"]
file-added-label = from my-defaults-feature

[delta "my-feature"]
file-added-label = from my-feature
```

e.g. with (using your new option)

```
delta --config /tmp/delta.config --features=my-feature --show-config
```

or with

```
export DELTA_FEATURES = '+my-feature'
```

ShogunPanda:
Oh, I see.
So, in other words, everything that I want to be "overwritable" from a feature must be extract off the main delta in a feature and eventually that feature can eventually be loaded by default in the main section. Am I right?

In that case you are right, my use case is redundant. I'll remove from the PR. Eventually, I'll just double check the docs and add this use case explicitly so that other people don't get in my same mistake.

dandavison:
Yes that's right. Improving the docs would be great, and I'm open to code changes that make the behavior more natural/coherent/intuitive if it doesn't change things too radically and feels like it's fixing a bug. But I'd prefer not to add additional options parameterizing the config resolution behavior since it will get very niche!

ShogunPanda:
I agree, maybe supporting both idea is too much.

Not sure which is the more "natural". TBH the current one didn't sound natural to me, but it might just be me. 

Probably since this is only relevant when using features, and as probably nobody can use (or get to know about them) features without reading the docs, just improving docs might be enough.

As you have seen, changing behavior it's only a matter to move an if block.

I'll remove the new option anyway. Just let me know if you want me to change the behavior or update the docs.

src/cli.rs
ShogunPanda: Done. What about now?

src/git_config/mod.rs
ShogunPanda: Good point. Fixed.

ShogunPanda:
As promised, I've handled the two code issues.
I also removed the `features-first` option and improved the documentation for the use case that confused me.

@dandavison Do you think are we good now?

src/cli.rs
```diff
@@ -55,6 +55,23 @@ A feature name may not contain whitespace. You can activate multiple features:
 
 If more than one feature sets the same option, the last one wins.
 
+If a value is present in the [delta] section, then features are not considered at all.
+
+If you want a value to be fully overridable by a feature and also have a non default value when
+no features are used, then you need to define a \"default\" feature and include it in the main
+delta configuration.
```
dandavison, line 62: I think the first two uses of the word "value" here are synonymous with "option" in earlier lines?

src/cli.rs
```diff
@@ -55,6 +55,23 @@ A feature name may not contain whitespace. You can activate multiple features:
 
 If more than one feature sets the same option, the last one wins.
 
+If a value is present in the [delta] section, then features are not considered at all.
+
+If you want a value to be fully overridable by a feature and also have a non default value when
+no features are used, then you need to define a \"default\" feature and include it in the main
```
dandavison, line 61:
```suggestion
If an option is present in the [delta] section, then features are not considered at all.

If you want an option to be fully overridable by a feature and also have a non default value when
no features are used, then you need to define a \"default\" feature and include it in the main
```

src/cli.rs
dandavison: Don't worry I just made the minor edit I was going on about and will merge now.

dandavison:
Awesome, thanks! I think this will be useful.

PR #1325, opened 2023-03-05
Update dependencies & switch back to the `dirs` crate

nickelc:
The `dirs`/`dirs-sys` repositories are unarchived and the crates are
maintained again, while `dirs-next` crates aren't.

The crate is also used by the `xdg` crate.


dandavison:
LGTM!

PR #1328, opened 2023-03-05
Return the `GitRemoteRepo` type directly

nickelc:
The type is unnecessarily wrapped and it's the only use for `GitConfigEntry::GitRemote` enum variant.


dandavison:
LGTM, thanks!

PR #1329, opened 2023-03-05
Get git's minus/plus style from config instead of copying it around

nickelc:
After removing the `GitConfigEntry::GitRemote` variant, the enum can be removed and the git config can be accessed directly.


dandavison:
Hi @nickelc, it's not immediately clear to me what this PR does and whether there's test coverage for the changes it's making. Can you add a more extensive description to the PR with sections like

- What this is aiming to do
- How it does it
- How do we know this is correct

Thanks!

PR #1337, opened 2023-03-09
Rename `git_config_entry` module to `remote`

nickelc:
The `GitConfigEntry` enum has been removed from the module and now only contains the `GitRemoteRepo` type.


dandavison:
LGTM

PR #1340, opened 2023-03-10
Set language specifiers for code blocks and fix links in manual

nickelc:
Plain links need to be enclosed in `<>` to linkify them.

The language specifiers fixes the rustdoc errors for `mdbook test`.


dandavison:
Awesome, thanks a lot, I checked and can see the improvements you made in the rendered Markdown.

PR #1341, opened 2023-03-10
Derive the default for enums with `#[derive(Default)]`/`#[default]`

nickelc:
This feature was stabilized with 1.62 and clippy now complains about it.


dandavison:
Nice, thanks, I saw the clippy complaints on your other PR.

PR #1342, opened 2023-03-10
Merge the different `GitConfig` constructors for a config file

nickelc:
The `try_create_from_path` function and the `from_path` function for tests can be merged into a single function.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,6 +1,6 @@
 use std::collections::{HashMap, HashSet};
 use std::ffi::OsString;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use bat::assets::HighlightingAssets;
 use clap::{ColorChoice, CommandFactory, FromArgMatches, Parser};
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -1142,7 +1142,8 @@ impl Opt {
 
         if let Some(path) = matches.get_one::<String>("config") {
             if !path.is_empty() {
-                final_config = Some(GitConfig::try_create_from_path(&env, path));
+                let path = Path::new(path);
+                final_config = Some(GitConfig::from_path(&env, path, true));
```
dandavison, line 1146: As a matter of style in languages like Rust with convenient enums/ADTs, do you think there's an argument for creating an enum so that this call site looks something like

```rust
GitConfig::from_path(&env, path, HonorGitConfigParametersEnvVar.Yes)
```

making it clear what the semantics of the boolean is?

nickelc: I think more idiomatic would be a builder.
```rust
GitConfig::from_path(&env, path).ignore_git_config_parameters(true);
```
Personally, I don't mind a boolean parameter because of my intellisense setup that shows the signature, doc comments and autocompletes this function to `GitConfig::from_path(&env, path, honor_env_var)`.

dandavison: Yep, fair enough! Feel free to introduce builders.

PR #1343, opened 2023-03-11
Set MSRV to 1.64.0 in package metadata

nickelc:
clippy respects the rust version set in the package metadata and wouldn't break the build job for newer rust features.

The current clap version in use dictates 1.64.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ categories = ["command-line-utilities", "development-tools"]
 description = "A syntax-highlighting pager for git"
 documentation = "https://github.com/dandavison/delta"
 edition = "2018"
+rust-version = "1.64.0"
 homepage = "https://github.com/dandavison/delta"
 license = "MIT"
 repository = "https://github.com/dandavison/delta"


dandavison:
Hi @nickelc I trust your changes but would you mind spelling out for me exactly what problem this change is solving?

nickelc:
First it documents the MSRV for readers and then it limits clippy to only issue warnings for lints that are compatible with the MSRV.

This reduces the possibility of a broken `clippy` build job and allows the version requirements to be changed deliberately.

`delta` could also be checked against this MSRV with a CI job.

nickelc:
> `delta` could also be checked against this MSRV with a CI job.

This is how the CI job could look like.
https://github.com/dandavison/delta/blob/2c48e356c8b5be5ddd0ca1b7a4e5b98f46f95bdd/.github/workflows/ci.yml#L111-L124

nickelc:
> How will I know when to update the MSRV?

A build job would allow to check PRs against the MSRV https://github.com/dandavison/delta/pull/1343#issuecomment-1464964115 and then decide to deliberately bump the MSRV.

dandavison:
It might well be my lack of familiarity with standard Rust practices but my worry is that this is more moving parts to maintain. When I'm in doubt I sometimes look at how bat and ripgrep do things, as examples of Rust CLI projects, so I'll try to do that.

PR #1345, opened 2023-03-11
Remove unneeded `extern crate` statements

nickelc:
With edition 2018, `extern crate` is no longer needed in most circumstances.


dandavison:
Thanks!

PR #1361, opened 2023-03-19
Update `bitflags` to 2.2

nickelc:
The `Clone`, `Copy`, `Debug` and `PartialEq` traits are no longer derived automatically by the macro.


dandavison:
Thanks!

PR #1402, opened 2023-04-25
Clarify how to use delta with `tmux`

isak102:
I recently spent a lot of time trying to make delta work with tmux and I was very confused since my neovim colors worked properly but not delta. In the end I noticed that I had to explicitly enable true color as the default was `auto`. I thought this information might be useful to other people too


isak102:
Yeah that worked aswell

dandavison:
OK great, thanks. I might add that as another thing people could check (my understanding is terminal emulators are supposed to set it themselves).

PR #1404, opened 2023-04-26
Improve manual build job

nickelc:
- Only trigger the build if the book has changed.
- Split the job into `build` and `deploy` with specific permissions.
- Don't build documentation for dependencies.
- Deploy to GitHub Pages directly without going through the `gh-pages` branch.

The third point fixes the current problem building the docs of the `rustix` crate with Rust 1.69.

### DIFF
--- a/.github/workflows/manual.yml
+++ b/.github/workflows/manual.yml
@@ -4,29 +4,53 @@ on:
   push:
     branches:
       - master
+    paths:
+      - "manual/**"
+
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks, these changes all look like improvements.

nickelc:
The repository settings must be adjusted for the manual build.
I thought `actions/configure-pages` would take care of it.

image

PR #1405, opened 2023-04-26
Replace deprecated `error_chain` crate with `anyhow`

nickelc:
The `error_chain` crate is now deprecated for a long time and [`anyhow`](https://crates.io/crates/anyhow) has proven to be a popular replacement for applications.

This also improves the current error messages for panics.

```
PAGER='"less' git show
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: \
Error(Msg("Could not parse pager command."), State { next_error: Some(ParseError), \
backtrace: InternalBacktrace })', src/main.rs:136:88
```

```
PAGER='"less' git show
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Could not parse pager command.

Caused by:
    missing closing quote', src/main.rs:125:88
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -45,6 +45,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.70"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7de8ce5e0f9f8d88245311066a578d72b7af3e7088f32783804676302df237e4"
+
+[[package]]
 name = "approx"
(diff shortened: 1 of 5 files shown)


dandavison:
Thanks a lot. This is something I was vaguely aware of but had never got around to (and hadn't understood the libraries involved.)

nickelc:
Basically, it's `anyhow` for applications and `thiserror` for libraries (codegen for `Display` & `From` trait impls).

PR #1413, opened 2023-05-08
Fix submodule diff parsing

th1000s:
Added support for an optional -dirty hash suffix which is present in `git diff` output when files were changed in a submodule.


dandavison:
Hi Thomas! Thanks.

PR #1430, opened 2023-06-03
Fix typo in publishing build job & restore cross-compilation

nickelc:
The typo prevented the cross-compilation targets from using the `cross`
binary for the compilation.

The script evaluated with `${{ matrix.job.cross }}` to:
```
  if [[ "" == "true" ]]; then
    cross build --release --target x86_64-unknown-linux-musl
  else
    cargo build --release --target x86_64-unknown-linux-musl
  fi
```

The correct result for `${{ matrix.job.use-cross }}` is:
```
  if [[ "true" == "true" ]]; then
    cross build --release --target x86_64-unknown-linux-musl
  else
    cargo build --release --target x86_64-unknown-linux-musl
  fi
```

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -24,6 +24,18 @@ jobs:
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
             use-cross: false
+          - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dandavison:
Thanks!

PR #1440, opened 2023-06-05
wrap --help output, fix rust 1.80 build

th1000s:
Set clap option `max_term_width`, so the help output is by default as wide as the terminal or this max value. Then manually wrap the `after_long_help()` text (only on demand) to the same width using `textwrap`. Also use matching ansi codes in this section.

----

Another try to improve deltas --help output, and I finally can use textwrap! But now `--version` and `--help` is handled manually. Another possibility would be to manually construct the pieces of claps `StyledStr`  (...which is returned from `render_help()` and according to `dbg!()` different from the simple String wrapper? `render_help() = StyledStr {   pieces: [ (None, "A viewer for git and diff output",), (..)] }` ).

CC @tjquillan @nickelc @FGasper

### DIFF
--- a/manual/src/full---help-output.md
+++ b/manual/src/full---help-output.md
@@ -9,7 +9,7 @@ Arguments:
   [MINUS_FILE]
           First file to be compared when delta is being used in diff mode
 
-          `delta file_1 file_2` is equivalent to `diff -u file_1 file_2 | delta`.
+          `delta file1 file2` is equivalent to `diff -u file1 file2 | delta`.
 
   [PLUS_FILE]
           Second file to be compared when delta is being used in diff mode
@@ -18,24 +18,39 @@ Options:
       --blame-code-style <STYLE>
           Style string for the code section of a git blame line.
 
-          By default the code will be syntax-highlighted with the same background color as the blame format section of the line (the background color is determined by blame-palette). E.g. setting this option to 'syntax' will syntax-highlight the code with no background color.
+          By default the code will be syntax-highlighted with the same
+          background color as the blame format section of the line (the
+          background color is determined by blame-palette). E.g. setting this
+          option to 'syntax' will syntax-highlight the code with no
+          background color.
 
       --blame-format <FMT>
           Format string for git blame commit metadata.
 
-          Available placeholders are "{timestamp}", "{author}", and "{commit}".
+          Available placeholders are "{timestamp}", "{author}", and
+          "{commit}".
 
           [default: "{timestamp:<15} {author:<15.14} {commit:<8}"]
 
       --blame-palette <COLORS>
-          Background colors used for git blame lines (space-separated string).
+          Background colors used for git blame lines (space-separated
+          string).
 
-          Lines added by the same commit are painted with the same color; colors are recycled as needed.
+          Lines added by the same commit are painted with the same color;
+          colors are recycled as needed.
 
       --blame-separator-format <FMT>
-          Separator between the blame format and the code section of a git blame line.
+          Separator between the blame format and the code section of a git
+          blame line.
 
-          Contains the line number by default. Possible values are "none" to disable line numbers or a format string. This may contain one "{n:}" placeholder and will display the line number on every line. A type may be added after all other format specifiers and can be separated by '_': If type is set to 'block' (e.g. "{n:^4_block}") the line number will only be shown when a new blame block starts; or if it is set to 'every-N' the line will be show with every block and every N-th (modulo) line.
+          Contains the line number by default. Possible values are "none" to
+          disable line numbers or a format string. This may contain one
+          "{n:}" placeholder and will display the line number on every line.
+          A type may be added after all other format specifiers and can be
+          separated by '_': If type is set to 'block' (e.g. "{n:^4_block}")
+          the line number will only be shown when a new blame block starts;
+          or if it is set to 'every-N' the line will be show with every block
+          and every N-th (modulo) line.
 
           [default: │{n:^4}│]
 
@@ -50,14 +65,20 @@ Options:
       --blame-timestamp-output-format <FMT>
           Format string for git blame timestamp output.
 
-          This string is used for formatting the timestamps in git blame output. It must follow the `strftime` format syntax specification. If it is not present, the timestamps will be formatted in a human-friendly but possibly less accurate form.
+          This string is used for formatting the timestamps in git blame
+          output. It must follow the `strftime` format syntax specification.
+          If it is not present, the timestamps will be formatted in a
+          human-friendly but possibly less accurate form.
 
-          See: (https://docs.rs/chrono/latest/chrono/format/strftime/index.html)
+          See:
+          <https://docs.rs/chrono/latest/chrono/format/strftime/index.html>
 
       --color-only
           Do not alter the input structurally in any way.
 
-          But color and highlight hunk lines according to your delta configuration. This is mainly intended for other tools that use delta.
+          But color and highlight hunk lines according to your delta
+          configuration. This is mainly intended for other tools that use
+          delta.
 
       --config <PATH>
           Load the config file at PATH instead of ~/.gitconfig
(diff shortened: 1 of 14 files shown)


dandavison:
Thanks! I tried this out and it seems like a good improvement to me.

(Tangential: Ultimately I feel that the lengthy man-page style output should be sent into a pager.)

th1000s:
Now that `--help` is handled manually, piping the output into `less -R` is easy. And clap doesn't seen to support a pager currently.

dandavison:
(There's a formatter error here.)

th1000s:
The fmt is fixed, but sending the output into a pager via `OutputType::from_mode` (now using a lighter `PagerCfg`) doesn't work, the default pager `less` just quits after the first page. `PagingMode::Always` doesn't help, and a default `DeltaEnv::init()` works fine for the normal delta. However `DELTA_PAGER="cat"` does display the entire output. strace shows less is started as `"/usr/bin/less", ["less", "--RAW-CONTROL-CHARS"]` as expected, what am I missing?

dandavison:
Hm. Is it the `LESS` environment variable supplying additional arguments?

src/cli.rs
```diff
@@ -1117,13 +964,57 @@ pub enum InspectRawLines {
...
+                    &PagerCfg::default(),
+                );
+
+                // works fine:
+                // print!("{}", help.ansi());
+
+                let mut writer = output_type.unwrap();
+                let mut writer = writer.handle().unwrap();
+                {
+                    write!(&mut writer, "{}", help.ansi()).unwrap();
```
th1000s, line 998: It wasn't the writer drop impl/destructor but the one of output_type!

src/cli.rs
```diff
@@ -1117,13 +964,57 @@ pub enum InspectRawLines {
...
+                    None,
+                    &PagerCfg::default(),
+                );
+
+                // works fine:
+                // print!("{}", help.ansi());
+
+                let mut writer = output_type.unwrap();
+                let mut writer = writer.handle().unwrap();
+                {
```
th1000s, line 997: It wasn't the writer drop impl/destructor but the one of output_type!

src/cli.rs
```diff
@@ -1117,13 +964,57 @@ pub enum InspectRawLines {
...
+                    write!(&mut writer, "{}", help.ansi()).unwrap();
+                    // dtor?
+                }
```
th1000s, line 1000: It wasn't the writer drop impl/destructor but the one of output_type!

th1000s:
Now with pager support, even listens to `PAGER` etc. env vars for configuration (but not `--paging never`).

The `full---help-output.md` markdown is a bit out of sync, but could be generated from then ansi `--help` output itself. However the one tool for that, [term2md](https://github.com/jonm/term2md), currently removes leading whitespace.

Oh and I found `--show-themes` is/was already broken.

dandavison:
This one has conflicts for us to resolve. Sorry for letting it drag on.

th1000s:
The full help output has to be regenerated (some ansi2md util would help), and I have to check that the big moved textblock in `cli.rs` is still correct.

src/cli.rs
```diff
@@ -1129,6 +976,196 @@ pub struct Opt {
...
+
+Use '<' for left-align, '^' for center-align, and '>' for right-align.
+
+
+{H_}Support{_H}
+
+If something isn't working correctly, or you have a feature request, please open an issue at https://github.com/dandavison/delta/issues.
+
+For a short help summary, please use delta -h.
+"##
```
dandavison, line 1163: The result of all this looks really nice.

src/utils/bat/output.rs
```diff
@@ -13,6 +13,28 @@ use crate::env::DeltaEnv;
 use crate::fatal;
 use crate::features::navigate;
 
+#[derive(Debug, Default)]
+pub struct PagerCfg {
+    pub navigate: bool,
+    pub show_themes: bool,
```
dandavison, line 19: I should have added a comment on this code...

```rust
                if config.show_themes {
                    p.arg("+n");
                }
```                

If you happen to know or discover what the logic is do feel free to add a comment!

src/cli.rs
```diff
@@ -1168,9 +1205,70 @@ pub enum DetectDarkLight {
...
+impl<A> Call<A> {
+    fn try_convert<B>(self) -> Option<Call<B>> {
+        use Call::*;
+        match self {
+            Delta(_) => None,
+            Help(help) => Some(Help(help)),
+            Version(ver) => Some(Version(ver)),
+        }
+    }
+}
```
dandavison, line 1227: This all looks like a nice new organization of the command processing.

dandavison (approved):
LGTM! This will be a big improvement. The paging and new styling of the lengthy sections are really nice.


> The full help output has to be regenerated

Just redirecting `--help` into that file and fixing the enclosing backticks, right -- or is something more needed?

manual/src/full---help-output.md
```diff
@@ -9,7 +9,8 @@ Arguments:
   [MINUS_FILE]
           First file to be compared when delta is being used in diff mode
 
-          `delta file_1 file_2` is equivalent to `diff -u file_1 file_2 | delta`.
+          `delta file_1 file_2` is equivalent to `diff -u file_1 file_2 |
+          delta`.
```
th1000s, line 14: Now that is annoying, doesn't fit in 80 lines!

manual/src/full---help-output.md
```diff
@@ -632,165 +855,234 @@ Options:
...
   -V, --version
           Print version
 
-GIT CONFIG
-----------
 
-By default, delta takes settings from a section named "delta" in git config files, if one is present. The git config file to use for delta options will usually be ~/.gitconfig, but delta follows the rules given in https://git-scm.com/docs/git-config#FILES. Most delta options can be given in a git config file, using the usual option names but without the initial '--'. An example is
+Git config
+
+  By default, delta takes settings from a section named "delta" in git config
```
th1000s, line 869: This mentions the git config, but should also mention how to set `pager = delta`. This way `--help` is a stand-alone manual, at least for the basics.

And this quick start section should be right after `Arguments:`, but that will require some more `--help` surgery.

th1000s (commented):
Okay, am redirecting `--help` into the markdown.

Also, I wrote... another wrap function 🙂. This time for the help output because I needed variable indentation.

manual/src/full---help-output.md
dandavison: I'd be happy to lose the underscores; I think I'd debated them once or twice with myself.

th1000s: Nice, now it fits!

src/utils/bat/output.rs
th1000s: `+cmd`, "specified cmd to be executed", so on start is it as if the user pressed 'n' to continue the search.

th1000s:
Also fixed the rust 1.80 build, note the Cargo.* changes. FTR, the build for the time crate was:

```
  --> /Users/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.31/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++

```

manual/src/full---help-output.md
dandavison: > This mentions the git config, but should also mention how to set pager = delta. This way --help is a stand-alone manual, at least for the basics.

Agreed. Very happy for you to add this but let me know if you want me to do anything.

th1000s: This PR is long enough, lets improve the help content in a different PR :)

src/utils/bat/output.rs
dandavison: Right thanks, I'll add that as a comment.

dandavison:
I've approved this one already; please merge at will!

dandavison:
I think this PR accidentally caused `delta -h` lines to be wrapped; I've opened https://github.com/dandavison/delta/issues/1795.

ofalvai:
@dandavison would it be possible to cut a new release with this Rust 1.80 fix? It would make it possible to upgrade delta in nixpkgs (https://github.com/NixOS/nixpkgs/issues/332957)

dandavison:
@ofalvai will do! I'll ping here when it's done.

PR #1450, opened 2023-06-19
Do not panic on superimposition mismatch in release builds

th1000s:
Prints: **!!<delta bug 1450>** .... **!please report!** which should be eye-catching enough.

Lines which would trigger a panic and early exit with the message "String mismatch encountered while superimposing style sections" now try to print the line as well as possible plus a brief error message.

In debug builds and when running from the target/ dir (such as when core.pager = "target/release/delta" is set) or when the env var DELTA_NO_WORKAROUNDS is set this mismatch panics as before.


dandavison:
Good call -- I agree that the nature of delta is such that it is generally more helpful to just keep going.

th1000s:
So, if this error shows up in the side-by-side view it can get truncated, all I can do is trim the message a bit. Well, still better than crashing...

Remind me, why is the superimposition done so late, couldn't it be done after `Painter::update_diff_style_sections()`? This is e.g. why wrap has to be called for the syntax and the diff part separately.

dandavison:
> Remind me, why is the superimposition done so late, couldn't it be done after Painter::update_diff_style_sections()? This is e.g. why wrap has to be called for the syntax and the diff part separately.

Interesting, I hadn't noticed that this was a bit ugly and that there seems to be the possibility that you point out. I looked at the code and I agree it looks superficially at least as if it could be done after `Painter::update_diff_style_sections()`

PR #1472, opened 2023-07-17
Bump clap to v4.3.14 and regenerate Cargo.lock

th1000s:
Mostly get an up to date rustix crate which fixes [bytecodealliance/rustix#575: rustix keeps rebuilding in my dependency tree](https://github.com/bytecodealliance/rustix/issues/575)


dandavison:
LGTM thanks

Issue #447, opened 2020-12-15
🚀 Use light/dark theme base on terminal theme or system variable.


tshu-w:
Hi, I have set my terminal to auto change themes based on macOS appearance. It would be much appreciated if delta can use light/dark theme base on terminal theme or system variable.

lourenci:
I second that. I'm struggling with the same issue. 

I don't care about the `delta` theme at all, I wish `delta` could just alternate between `--light` and `--dark` colors according to the terminal colors automatically.

Out of the curiosity, what terminal emulator are you using @tshu-w? Does it support this out-of-the-box?

Thank you.

tshu-w:
@lourenci Hi, I'm using iTerm2 on macOS, unlike the terminal that comes with the system, it requires a [script](https://gist.github.com/jamesmacfie/2061023e5365e8b6bfbbc20792ac90f8) to make the automatic switch possible.

dandavison:
Hi @tshu-w and @lourenci, yes let's figure out a good way to do this. What do you think of this proposal:

- There are two parts to this: (1) detecting the relevant system state (e.g. OS dark mode, or current terminal color theme etc) and (2) getting delta to respond to the relevant system state.

- I suggest we address those separately. Specifically, I suggest that we say that (1) is not Delta's responsibility. Instead, Delta will expose an environment variable named `DELTA_FEATURES`. This environment variable does exactly the same thing as the existing `--features` flag. (see [custom features](https://github.com/dandavison/delta#custom-features) and [`--help` output](https://github.com/dandavison/delta#full---help-output))

- Thus it is the user's responsibility to populate the `DELTA_FEATURES` environment variable with the names of the custom features they want to be active, in response to changes in their system state. One way to do this might be to use shell prompt code (i.e. shell code that is executed every time the prompt is displayed).

- So, for example, one might define custom features something like below, and then arrange for `DELTA_FEATURES` to contain either `"my-light-mode"` or `"my-dark-mode"` as appropriate.
```gitconfig
[delta "my-light-mode"]
    light = true
    file-style = magenta # or whatever other settings you want; this is just an example

[delta "my-dark-mode"]
    dark = true
    file-style = green # or whatever other settings you want; this is just an example
```

If you're able to [build delta from source](https://github.com/dandavison/delta#build-delta-from-source) then you can play around with this now as I've added support for the `DELTA_FEATURES` env var. cc @ulwlu ref https://github.com/dandavison/delta/pull/340#issuecomment-744257639

[...]

PR #1493, opened 2023-07-29
Try to obtain default dark/light state from the terminal

aykevl:
Use the termbg crate to try to read the current terminal background theme (light or dark) and use it as the default if no `--light` or `--dark` was specified (or `light=true` in .gitconfig).

This doesn't work in all cases: delta is frequently used with a pipe, in which case it's not really possible to obtain the background color using OSC 11. The fallback is the environment variable COLORFGBG which is supported by most terminals, but it doesn't immediately update when the system theme changes (it only applies to newly opened terminals). Still, I think this is better than assuming a dark background.

See https://github.com/dandavison/delta/issues/447 for discussion.

---

This PR is not ready yet. There are two problems:

1. termbg doesn't check whether stdin/stderr are connected to a terminal. This is fixed in https://github.com/dalance/termbg/pull/19 but isn't merged yet.
2. Lots of tests fail with this PR. I think I'm going to need a bit of help here. The issue is that the tests also try to read from the current terminal, while they shouldn't. Honestly the only place that should be querying the terminal should be somewhere in main, certainly not in a test. One way to fix this would be to provide some sort of mock terminal background read function, but my Rust knowledge is rather limited and I'm not sure how this is best implemented.


aykevl:
I have been investigating this problem some more and it looks like it's fundamentally impossible to request the background color from the terminal when the output is piped to less.

The main call here is "OSC 11". I don't know where it is documented, but [this](https://stackoverflow.com/questions/2507337/how-to-determine-a-terminals-background-color) is one place that describes it.
To be able to read the response from the terminal, two flags need to be disabled: `ECHO` and `ICANON` (see [`man tcsetattr`](https://linux.die.net/man/3/tcsetattr)). This avoids showing the response in the terminal and makes it possible to read it (without waiting for a newline first). The termbg crate sets the terminal to raw mode, this is a bit overkill but also removes those two flags (among many others).
The problem here is that less also changes the terminal state: it changes these flags manually but basically it also sets it to raw mode, including `ECHO` and `ICANON`. So what happens is that there is essentially a race condition between termbg and less. For example, termbg sets the terminal to raw mode, then less sets it to raw mode, and then termbg disables raw mode (making less unusable).
I've been thinking about a solution to this, but I don't see a way to avoid the race condition entirely.

---

What would work, is reading the `COLORFGBG` environment variable. This is set by a number of terminals, apparently including Konsole, iTerm2, and the built-in macOS terminal. The problem with that one is that it doesn't respond to changes in the color scheme, but it will be correct for newly created terminals. Perhaps that would be better than assuming a dark background in all cases?

Issue #1490, opened 2023-07-27
Delta appending ^[[OK at the end of each modified line ( color related i think )


ippsav:
![20230727_09h31m05s_grim](https://github.com/dandavison/delta/assets/69125922/84d3029d-49aa-4a79-a340-56a6050a919c)

Delta version: `delta 0.16.5`

dandavison:
Hi @ippsav, this is likely to be a problem with your environment (terminal emulator) since it hasn't been reported by other users.

- Which terminal emulator are you using?
- Are you using tmux or screen?
- What OS are you using?

Please try a different terminal emulator and see if you see the same problem.

ippsav:
Terminal: alacritty
Session manager: tmux
OS: Arch linux

sophiabrandt:
I'm encountering the same issue when I use the fish shell.

It seems to make no difference if I use tmux or not.

Emulator: kitty
OS: macOs M1

It works when I use zsh with iterm2.
It also works with zsh and kitty.

I've been searching through the fish shell github, but haven't found an obvious solution.

Are you using fish shell as well @ippsav ?

dandavison:
Hi @sophiabrandt, @ippsav. Hm, you probably already know this but for the record that code is ANSI CSI code "Erase in Line" (https://en.wikipedia.org/wiki/ANSI_escape_code). Delta uses it to cause the background colour to extend to the right-hand edge of the terminal window.

FWIW I can't reproduce it with fish 3.6.1, Alacritty, MacOS M2, with or without tmux.

ippsav:
No i am using zsh

sophiabrandt:
@dandavison in any case, many thanks for your work on delta. ❤️

michenriksen:
@ippsav @sophiabrandt do you use [`most`](https://www.jedsoft.org/most/) or maybe some other fancy pager? (`echo $PAGER`)

I saw the same strange rendering problem, and I finally narrowed it down to `export PAGER='most'` in my `.zshrc` file.

I fixed it by configuring delta to use less in my `.gitconfig`:

```gitconfig
[delta]
  pager = less
```

@dandavison I found this in the [README for bat](https://github.com/sharkdp/bat#using-a-different-pager):

> _Note: If PAGER is more or most, bat will silently use less instead to ensure support for colors._

Maybe it would make sense for delta to do the same?

sophiabrandt:
@michenriksen yes! That's it! Thank you. Never would have thought of the pager... 😅

[...]

PR #1494, opened 2023-07-29
Fix issue where pager would be set to `more` or `most`

ippsav:
### Bug
Colors wouldn't render properly in case the `$PAGER` is set to `more` or `most`.(fix for #1490 )
20230729_23h00m14s_grim
### Solution
Handle the case where `$PAGER` is set to `more` or `most` by using `bat::config::get_pager_executable` which sets the pager to `less` if that's the case.

Unit tests:

> env.rs:
>   - test_env_parsing_with_pager_set_to_bat
>   - test_env_parsing_with_pager_set_to_more
>   - test_env_parsing_with_pager_set_to_most

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -39,8 +39,8 @@ impl DeltaEnv {
         let pagers = (
             env::var(DELTA_PAGER).ok(),
             // We're using `bat::config::get_pager_executable` here instead of just returning
-            // the pager from the environment variables, because we want to make sure 
-            // that the pager is a valid pager from env and handle the case of 
+            // the pager from the environment variables, because we want to make sure
+            // that the pager is a valid pager from env and handle the case of
             // the PAGER being set to something invalid like "most" and "more".
             bat::config::get_pager_executable(None),
         );
--- a/src/env.rs
+++ b/src/env.rs
@@ -41,7 +41,7 @@ impl DeltaEnv {
         let pagers = (
             env::var(DELTA_PAGER).ok(),
             env::var(BAT_PAGER).ok(),
-            // We're using `bar::config::get_pager_executable` here instead of just returning
+            // We're using `bat::config::get_pager_executable` here instead of just returning
             // the pager from the environment variables, because we want to make sure 
             // that the pager is a valid pager from env and handle the case of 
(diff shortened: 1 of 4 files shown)


src/env.rs
```diff
@@ -41,7 +41,11 @@ impl DeltaEnv {
         let pagers = (
             env::var(DELTA_PAGER).ok(),
             env::var(BAT_PAGER).ok(),
```
dandavison, line 43: Would you mind getting rid of all explicit mention of `BAT_PAGER`. `get_pager_executable` will look at `BAT_PAGER` itself.

src/env.rs
```diff
@@ -10,7 +10,7 @@ const DELTA_EXPERIMENTAL_MAX_LINE_DISTANCE_FOR_NAIVELY_PAIRED_LINES: &str =
     "DELTA_EXPERIMENTAL_MAX_LINE_DISTANCE_FOR_NAIVELY_PAIRED_LINES";
 const DELTA_PAGER: &str = "DELTA_PAGER";
 const BAT_PAGER: &str = "BAT_PAGER";
-const PAGER: &str = "PAGER";
+// const PAGER: &str = "PAGER";
```
dandavison, line 13: Could you remove this commented line please.

src/env.rs
```diff
@@ -41,7 +41,11 @@ impl DeltaEnv {
         let pagers = (
             env::var(DELTA_PAGER).ok(),
             env::var(BAT_PAGER).ok(),
-            env::var(PAGER).ok(),
+            // We're using `bar::config::get_pager_executable` here instead of just returning
```
dandavison, line 44:
```suggestion
            // We're using `bat::config::get_pager_executable` here instead of just returning
```

dandavison (changes requested):
This is awesome, thanks very much! I've made a couple of small requests for changes to the diff.

ippsav:
@dandavison all fixed ! that's true now that i read the source code of bat, no need for handling `BAT_PAGER`, thanks !

dandavison:
Perfect, thanks for doing this!

ippsav:
You're welcome ! it was a fun thing to do ! and thanks for making such a useful app !

Issue #1487, opened 2023-07-24
🐛 `is out of bounds of` due to URL in Json from ripgrep not being encoded


tbm:
I am trying ripgrep in combination with delta based on your recent post in the ripgrep thread.

I get this:

```
thread 'main' panicked at 'byte index 63 is out of bounds of `https://en.wikipedia.org/w/api.php?action=help&modules=query

`', src/handlers/grep.rs:358:38
stack backtrace:
```

Reproduce:

```
cat x | delta
```

x

```
{"type":"match","data":{"path":{"text":"TODO"},"lines":{"text":"https://en.wikipedia.org/w/api.php?action=help&modules=query\n\n\n"},"line_number":4,"absolute_offset":148,"submatches":[{"match":{"text":"\n\n\n"},"start":60,"end":63}]}}
```

If I encode the `&` in the URL, it works.

I'm not sure if the bug is in delta or grip.

This is with delta 0.16.5.

ippsav:
I've been tinkering with this issue for a couple of minutes, i ll be making a pr for this soon to handle this issue

ippsav:
@dandavison pr ready for review ! #1496

dandavison:
@tbm @ippsav what's an example of making `rg` product output with multiple trailing newlines like this:

```
{"text":"https://en.wikipedia.org/w/api.php?action=help&modules=query\n\n\n"}
```

Does it require using the `--multiline` flag (`-U`)?

For example with input file

```
1

2
```

I can reproduce this:

```
rg --multiline --json '1\n\n' /tmp/in
```

and that crashes delta.

@ippsav is it usage of `--multiline` that your fix is addressing? I'm really sorry I've been so slow here; I didn't find this issue easy to think about and was low on time.

tbm:
> Does it require using the --multiline flag (-U)?

Sorry, I forgot to mention it.  Yes, it requires `-U`.

Example:

```
65569:tbm@jirafa: ~/tmp/wiktionary] rg -U "\n\n"  xt
hread 'main' panicked at 'byte index 47 is out of bounds of `https://dumps.wikimedia.org/backup-index.html
`', src/handlers/grep.rs:358:38
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
x 
zsh: exit 101
65570:tbm@jirafa: ~/tmp/wiktionary] cat x

https://dumps.wikimedia.org/backup-index.html

```

dandavison:
Thanks!

[...]

PR #1496, opened 2023-07-30
Fix ripgrep trailing newline

ippsav:
### Description
Fix trailing newline issue in code highlights (#1487 )

The ripgrep JSON output includes a trailing newline at the end of each match. However, the code highlighting in delta does not expect a trailing newline, which was causing an issue when the match contains a newline at the end. 

To fix this, We check if the trailing newline is part of the actual match text and it's at the end of the string. If not, the trailing newline is removed.

### Unit tests:
  `ripgrep_json.rs`:
- test_match_text_ends_with_newline
- test_match_text_without_newline

### DIFF
--- a/src/handlers/ripgrep_json.rs
+++ b/src/handlers/ripgrep_json.rs
@@ -127,6 +127,20 @@ mod tests {
     use super::*;
 
     #[test]
+    fn test_match_text_ends_with_newline() {
+        let line = r#"{"type":"match","data":{"path":{"text":"src/cli.rs"},"lines":{"text":"    fn from_clap_and_git_config(config\n"},"line_number":null,"absolute_offset":35837,"submatches":[{"match":{"text":"config\n"},"start":4,"end":7}]}}"#;
+        let grep_line = parse_line(line).unwrap();
+        assert_eq!(grep_line.code.chars().last(), Some('\n'));
+    }
+
+    #[test]
+    fn test_match_text_without_newline() {
+        let line = r#"{"type":"match","data":{"path":{"text":"src/cli.rs"},"lines":{"text":"    fn from_clap_and_git_config(\n"},"line_number":null,"absolute_offset":35837,"submatches":[{"match":{"text":"config"},"start":4,"end":6}]}}"#;
+        let grep_line = parse_line(line).unwrap();
+        assert_eq!(grep_line.code.chars().last(), Some('('));
+    }
+
+    #[test]
     fn test_deserialize() {
         let line = r#"{"type":"match","data":{"path":{"text":"src/cli.rs"},"lines":{"text":"    fn from_clap_and_git_config(\n"},"line_number":null,"absolute_offset":35837,"submatches":[{"match":{"text":"fn"},"start":4,"end":6}]}}"#;
         let ripgrep_line: RipGrepLine = serde_json::from_str(line).unwrap();
--- a/src/handlers/ripgrep_json.rs
+++ b/src/handlers/ripgrep_json.rs
@@ -15,7 +15,14 @@ pub fn parse_line(line: &str) -> Option<grep::GrepLine> {
             // A real line of rg --json output, i.e. either of type "match" or
             // "context".
             let mut code = ripgrep_line.data.lines.text;
-            if code.ends_with('\n') {
+            // ripgrep --json emits a trailing newline for each match, but
+            // delta's code highlighting does not expect a trailing newline.
+            // Therefore, remove the trailing newline if it is not part of the
+            // match.
+            let newline_match_end = ripgrep_line.data.submatches.iter().any(|m| {
+                code.ends_with(&m._match.text) && m._match.text.ends_with('\n')
+            });
+            if !newline_match_end && code.ends_with('\n') {
                 code.truncate(code.len() - 1);
                 if code.ends_with('\r') {
                     code.truncate(code.len() - 1);


src/handlers/ripgrep_json.rs
```diff
@@ -15,7 +15,16 @@ pub fn parse_line(line: &str) -> Option<grep::GrepLine> {
...
-            if code.ends_with('\n') {
+            // ripgrep --json emits a trailing newline for each match, but
+            // delta's code highlighting does not expect a trailing newline.
+            // Therefore, remove the trailing newline if it is not part of the
+            // match.
+            let newline_match_end = ripgrep_line
+                .data
+                .submatches
+                .iter()
+                .any(|m| code.ends_with(&m._match.text) && m._match.text.ends_with('\n'));
```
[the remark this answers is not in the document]
ippsav: He might have multiple matches and it wouldn't always be the last match or am i missing something ?
```
[{"match":{"text":"\n\n\n"},"start":60,"end":63}}, {"match":{"text":"config"},"start":50,"end":55}}
```
This is not an actual example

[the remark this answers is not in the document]
dandavison: OK sorry I'll try to find time to think through this properly! Your PR looks great, the only reason I've been slow is because I was slightly confused about exactly what the condition should be for us stripping newline.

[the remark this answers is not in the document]
ippsav: No worries ! in any case keep me updated ! thanks

dandavison:
Could you give an example of a toy input file and a ripgrep command that crashes delta without this fix and behaves correctly with this fix?

ippsav:
Here is an example @dandavison :

example file: 
```json
{"type":"match","data":{"path":{"text":"src/cli.rs"},"lines":{"text":"    fn from_clap_and_git_config(config\n"},"line_number":null,"absolute_offset":35837,"submatches":[{"match":{"text":"config\n"},"start":33,"end":39}]}}
```
Output of running the command with delta:
20230806_14h35m27s_grim

Output of running the same command with the fix:
20230806_14h36m57s_grim

I ve just realised the test cases are not really testing what it should be testing i ll be fixing it right now, my bad on that !

dandavison:
Can you give the input file and ripgrep command that you're using to generate the example? I think if you include that command as a comment by the test case that would be helpful.

If you want to use delta's source code as the example input file you could make it reproducible via a command that specifies a master commit like

```
git show 0608765f40a193d16651e1ac0c547bd2f9280bf9:src/cli.rs | rg --json config
```

ippsav:
here is a simpler example that relates to the example i mentioned above:
`raw.txt`:
```
    fn from_clap_and_git_config(config

```
There is a new line at the end make sure you press enter.

The output of running `cat raw | rg -U --json "config\n"`
```json
{"type":"begin","data":{"path":{"text":"<stdin>"}}}
{"type":"match","data":{"path":{"text":"<stdin>"},"lines":{"text":"    fn from_clap_and_git_config(config\n"},"line_number":1,"absolute_offset":0,"submatches":[{"match":{"text":"config\n"},"start":32,"end":39}]}}
{"type":"end","data":{"path":{"text":"<stdin>"},"binary_offset":null,"stats":{"elapsed":{"secs":0,"nanos":14720,"human":"0.000015s"},"searches":1,"searches_with_match":1,"bytes_searched":40,"bytes_printed":265,"matched_lines":1,"matches":1}}}
{"data":{"elapsed_total":{"human":"0.000646s","nanos":646020,"secs":0},"stats":{"bytes_printed":265,"bytes_searched":40,"elapsed":{"human":"0.000015s","nanos":14720,"secs":0},"matched_lines":1,"matches":1,"searches":1,"searches_with_match":1}},"type":"summary"}
```

src/handlers/ripgrep_json.rs
```diff
@@ -15,7 +15,16 @@ pub fn parse_line(line: &str) -> Option<grep::GrepLine> {
             // A real line of rg --json output, i.e. either of type "match" or
             // "context".
             let mut code = ripgrep_line.data.lines.text;
-            if code.ends_with('\n') {
+            // ripgrep --json emits a trailing newline for each match, but
+            // delta's code highlighting does not expect a trailing newline.
+            // Therefore, remove the trailing newline if it is not part of the
+            // match.
```
dandavison, line 21: If delta needs to strip the `\n` in the usual case, then that must be because some bad thing X happens if it does not do it. So why is it OK to not strip the `\n` when it is part of the match? Why does X not happen?

ippsav: Because once it will be printed, the \n will be stripped since it's included in the match, unless im missing something here

dandavison: I'm sorry, I'm not following. Would you mind explaining in more detail? I'm not understanding the relationship between whether or not something is in the match, and the stripping newlines for the highlighter. I think it's me that's missing something!

ippsav: I need some time, I've been trying with the example you mentioned in the #1496  that's failing, i should look more into it, specially that now i am thinking about it i don't think i handled it well it's a bit confusing

tbm:
@ippsav do you have any time to update your PR with the comments from @dandavison ?  I'm the guy who reported this issue and would like to see your PR in. Thank you for working on this!

ippsav:
@tbm there have been a lot of work lately, hence I was not very active, I ll take a look at it very soon this week, sorry for the inconvenience !

Issue #96, opened 2020-02-12
Improve binary file added output


dandavison:
Ref #94. It might be better for the output to say something like

```
added: foo (binary file)
```

than the current

```
Binary files /dev/null and b/foo differ
```

anchnk:
Has this been implemented yet? Otherwise I am down to work on it as it seems a good fit to get started contributing to the repository

dandavison:
Hey @anchnk, great! Go for it. I have not started on it.

anchnk:
I had a quick look and I will dive deeper in a few but it seems that the project structure drastically changed. I could replay all the commits since then but I thought it might be faster to ask where is the relevant test located now?

dandavison:
Right, the codebase changed a lot around March/April this year. The test named `test_binary_files_differ` is in `src/tests/test_example_diffs.rs` now.

anchnk:
Thank you very much for your help. I was able to identify the test and where's the condition that identifies binary files modification is (thx to #94 ). Now I am trying to figuring out how to implement the new output. Should we have 3 different outputs for each scenario? Such as:

* `added: foo (binary file)`
* `modified: foo (binary file)`
* `removed: foo (binary file)`

Also, where's the delta output built? I am still grasping the code to see how it works but breaking this task in actionable items would greatly help to grasp the codebase.

Sondeyy:
Hi, is this still open? I would like to help

dandavison:
Hi @Sondeyy, this is still open. Go for it!

dandavison:
I was able to reproduce it with

```
echo a > myfile
gzip myfile
git add myfile.gz
git commit -m 'Commit binary file'
git show | delta --no-gitconfig
```

Sondeyy:
Okay, I'll check and get back to you, thanks.

Sondeyy:
Hi @dandavison, I opened #1502, let me know what you think and if the format and everything is okay.

PR #1502, opened 2023-08-05
Improve binary file added output 

Sondeyy:
Targeting: #96 

If the binary file changed, it will output the `Binary files a/foo and b/foo differ` as before.
If the binary file got added or deleted, it will append ` (binary file)` to the filename, print the diff header and skip the body. 

Tests: 
* binary file added 
* binary file removed
* binary file changed


dandavison:
LGTM! Thanks very much for this work. (I tested it manually using https://github.com/dandavison/delta/issues/96#issuecomment-1665808161)

PR #1503, opened 2023-08-07
Remove the navigation setting from themes.

Qwerty-133:
The mantis-shrimp and mantis-shrimp-lite themes set `navigate` to true, however themes should only tweak visual settings.


dandavison:
Thanks @Qwerty-133! I agree.

Issue #1513, opened 2023-08-18
🐛 `rg --json | delta` output has different colour depending on filename ("ini")


tbm:
I don't know if this is a bug or intentional or something but it seems strange to me.  I used ripgrep on some files and noticed that one file had a different colour.  The file "ini" has different colours to any other files. (Maybe "init" is special cased?)

Reproduce:

```
$ echo "# [[liver]] {{gloss|organ of the body}}" > a
$ cp a ini
$ rg --json "\{" a | delta
$ rg --json "\{" ini | delta
```

![delta](https://github.com/dandavison/delta/assets/1556037/cb98c97a-6919-4685-90b5-81264a403aa0)

tbm:
Original example where I noticed this:

![delta2](https://github.com/dandavison/delta/assets/1556037/e5abad83-b09e-4ef3-86b0-094c34aad57c)

It looks weird.

th1000s:
Fixed with #1549, thank you for the report!

PR #1549, opened 2023-10-03
Repair --default-language, and highlight using full filename

th1000s:
Repair --default-language, and highlight using full filename

Fixed that the "txt" fallback was used instead of --default-language

And when looking for syntax highlighting, keep the filename around
as long as possible.

Using simple custom logic (filename > 4) a Makefile can be highlight with
make syntax, but a cc or ini file does not get treated like a .cc or .ini file.

Currently the underlying highlighting lib syntect and the sublime syntex
definitions can not make this distinction
(http://www.sublimetext.com/docs/syntax.html).

----

Plus clippy warnings

### DIFF
--- a/src/handlers/diff_header.rs
+++ b/src/handlers/diff_header.rs
@@ -95,10 +95,10 @@ impl<'a> StateMachine<'a> {
         if self.source == Source::DiffUnified {
             self.state = State::DiffHeader(DiffType::Unified);
             self.painter
-                .set_syntax(get_file_extension_from_marker_line(&self.line));
+                .set_syntax(get_filename_from_marker_line(&self.line));
         } else {
             self.painter
-                .set_syntax(get_file_extension_from_diff_header_line_file_path(
+                .set_syntax(get_filename_from_diff_header_line_file_path(
                     &self.minus_file,
                 ));
         }
@@ -129,7 +129,7 @@ impl<'a> StateMachine<'a> {
             .unwrap_or(path_or_mode);
         self.plus_file_event = file_event;
         self.painter
-            .set_syntax(get_file_extension_from_diff_header_line_file_path(
+            .set_syntax(get_filename_from_diff_header_line_file_path(
                 &self.plus_file,
             ));
         self.current_file_pair = Some((self.minus_file.clone(), self.plus_file.clone()));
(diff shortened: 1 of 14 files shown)


th1000s:
The filename cutoff it 4, not 3 (from actual DOS 8.3 filenames) because of newer extensions like `.toml` and `.flac`. And I could not think of file formats without extensions which are just 4 short and have a meaning by itself, like `Makefile`. Maybe this could even be extended to 5.

This should fix #1513

th1000s:
The filename cutoff is 4, not 3 (from actual DOS 8.3 filenames) because of newer extensions like `.toml` and `.flac`. And I could not think of file formats without extensions which are just 4 short and have a meaning by itself, like `Makefile`. Maybe this could even be extended to 5.

This should fix #1513

src/handlers/diff_header.rs
```diff
@@ -479,49 +472,48 @@ mod tests {
...
-                "--- src/one.rs	2019-11-20 06:47:56.000000000 +0100"
-            ),
-            Some("rs")
+            get_filename_from_marker_line("--- src/one.rs	2019-11-20 06:47:56.000000000 +0100"),
+            Some("one.rs")
         );
     }
 
     #[test]
     fn test_get_file_extension_from_diff_header_line() {
```
dandavison, line 481: Looks like this test name needs to be updated

src/handlers/grep.rs
```diff
@@ -116,12 +116,8 @@ impl<'a> StateMachine<'a> {
...
                 );
                 if new_path {
-                    if let Some(lang) = handlers::diff_header::get_extension(&grep_line.path)
-                        .or(self.config.default_language.as_deref())
-                    {
-                        self.painter.set_syntax(Some(lang));
-                        self.painter.set_highlighter();
-                    }
+                    self.painter.set_syntax(Some(grep_line.path.as_ref()));
+                    self.painter.set_highlighter();
```
dandavison, line 120: That's nicer.

src/handlers/diff_header.rs
```diff
@@ -479,49 +472,48 @@ mod tests {
...
-            Some("rs")
+            get_filename_from_diff_header_line_file_path("wat hello.rs"),
+            Some("wat hello.rs")
         );
         assert_eq!(
-            get_file_extension_from_diff_header_line_file_path("/dev/null"),
+            get_filename_from_diff_header_line_file_path("/dev/null"),
             None
         );
+        // TODO, handle this:
```
dandavison, line 502: Can you expand comment to clarify in what sense this is not handled?

dandavison (changes requested):
Thanks @th1000s! Just that one question about the code comment in the test: in what sense is Makefile etc not handled.

tbm:
@th1000s I just saw you submitted a PR for an issue I reported. Thank you! do you by any chance to incorporate the feedback comments from @dandavison ?  I'd love to see your fix applied. Thanks!

src/handlers/diff_header.rs
th1000s: Done.

src/handlers/diff_header.rs
th1000s: Oh, that is handled now :) - Left over from writing the failing tests first.

dandavison:
Thanks @th1000s!

(The MacOS build failure occurs on `main`; I am not sure why yet)

PR #1550, opened 2023-10-18
Update installation instructions for openSUSE

sorairolake:
In addition to the Tumbleweed, since openSUSE Leap 15.5, git-delta is available in the official repository.[^1]

[^1]: <https://build.opensuse.org/package/show/openSUSE:Leap:15.5/git-delta>


dandavison:
Thanks!

PR #1560, opened 2023-11-09
use wildcard in suggested true color fix for tmux

### LINKED ISSUE #1559: update intsructions for tmux color issues
I lost some time figuring out that I need to change `set -ga terminal-overrides ",xterm-256color:Tc"` to `set -ga terminal-overrides ",*-256color:Tc"`. The wildcard will probably work for more people.

jcaplan:
It took me a few extra minutes to realize I had `TERM=screen-256color`. Update the docs so the suggested fix works on more configurations.

Fixes #1559


dandavison:
Thank you @jcaplan!

PR #1597, opened 2024-01-05
Upgrade bytelines to latest, without default features

whitfin:
Hi there! I found this project via an issue opened here: https://github.com/whitfin/bytelines/pull/7 (totally a mistake on my part).

I've finally resolved this temporarily, by adding a feature for Tokio. It's still enabled by default to avoid a breaking change, but this means that you can now disable default features to exclude Tokio + relevant dependencies. Figured I'd file it here so you no longer have to specifically pin a version going forward.

At some point (time permitting) there will be a bytelines v3 which does this more correctly, but for now it'll be one less thing to think about on your end! I'm not sure if you want the `Cargo.lock` changes committed, but I can drop those out as necessary. 

Thank you!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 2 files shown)


dandavison:
Thanks @whitfin and thanks for bytelines!

PR #1615, opened 2024-01-23
Detect Dark/Light Mode from Terminal

bash:
Use the [`terminal-colorsaurus`](https://docs.rs/terminal-colorsaurus) crate to detect light/dark mode if neither `--light` nor `--dark` is specified.

I was unsatisfied with existing crates (such as `termbg`) and so I went down the rabbit hole of making my own crate for this. It uses `OSC 10` / `OSC 11` on Unix ~~and winapi calls on Windows.~~ Edit: I have since removed Windows support for the reasons outlined [here](https://github.com/bash/terminal-colorsaurus/blob/main/doc/windows.md)

My crate has a few advantages over existing crates that do this:
* Compares the perceived lightness of background and foreground color (which means it correctly detects the theme in low constrast cases).
* Works if stdin, stdout and stderr are redirected.
* Safely restores the terminal from raw mode even if the library errors.
* No dependency on an async runtime.

To avoid race conditions when piped into a pager as [mentioned in a comment in this previous attempt](https://github.com/dandavison/delta/pull/1493#issuecomment-1656805343) color detection is disabled when stdout is attached to a pipe.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -3,7 +3,7 @@ use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 
 use bat::assets::HighlightingAssets;
-use clap::{ColorChoice, CommandFactory, FromArgMatches, Parser, ValueHint};
+use clap::{ColorChoice, CommandFactory, FromArgMatches, Parser, ValueEnum, ValueHint};
 use clap_complete::Shell;
 use lazy_static::lazy_static;
 use syntect::highlighting::Theme as SyntaxTheme;
@@ -293,6 +293,12 @@ pub struct Opt {
     /// Used when the language cannot be inferred from a filename. It will typically make sense to
     /// set this in per-repository git config (.git/config)
     pub default_language: Option<String>,
+
+    /// Detect whether or not the terminal is dark or light by querying for its colors.
+    ///
+    /// Ignored if either `--dark` or `--light` is specified.
+    #[arg(long = "detect-dark-light", value_enum, default_value_t = DetectDarkLight::default())]
+    pub detect_dark_light: DetectDarkLight,
 
     #[arg(long = "diff-highlight")]
     /// Emulate diff-highlight.
@@ -1122,6 +1128,17 @@ pub enum InspectRawLines {
     True,
     #[default]
     False,
+}
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, ValueEnum)]
+pub enum DetectDarkLight {
+    /// Only query the terminal for its colors if the output is not redirected.
+    #[default]
+    Auto,
+    /// Always query the terminal for its colors.
+    Always,
+    /// Never query the terminal for its colors.
+    Never,
 }
 
 impl Opt {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -297,6 +297,23 @@ pub struct Opt {
     /// Detect whether or not the terminal is dark or light by querying for its colors.
     ///
     /// Ignored if either `--dark` or `--light` is specified.
+    ///
+    /// Querying the terminal for its colors requires "exclusive" access
+    /// since delta reads/writes from the terminal and enables/disables raw mode.
+    /// This causes race conditions with pagers such as less when they are attached to the
+    /// same terminal as delta.
+    ///
+    /// This is usually only an issue when the output is manually piped to a pager.
+    /// For example: `git diff | delta | less`.
+    /// Otherwise, if delta starts the pager itself, then there's no race condition
+    /// since the pager is started *after* the color is detected.
+    ///
+    /// `auto` tries to account for these situations by testing if the output is redirected.
+    ///
+    /// The `--color-only` option is treated as an indicator that delta is used
+    /// as `interactive.diffFilter`. In this case the color is queried from the terminal even
+    /// though the output is redirected.
+    ///
     #[arg(long = "detect-dark-light", value_enum, default_value_t = DetectDarkLight::default())]
     pub detect_dark_light: DetectDarkLight,
 
--- a/src/options/theme.rs
+++ b/src/options/theme.rs
@@ -1,3 +1,5 @@
+use std::io::{stdout, IsTerminal};
+
 /// Delta doesn't have a formal concept of a "theme". What it has is
 /// (a) the choice of syntax-highlighting theme
 /// (b) the choice of light-background-mode vs dark-background-mode, which determine certain
@@ -8,9 +10,8 @@
 /// default is selected.
 use bat;
 use bat::assets::HighlightingAssets;
-use terminal_colorsaurus::Preconditions;
 
-use crate::cli;
+use crate::cli::{self, DetectDarkLight};
 
 #[allow(non_snake_case)]
 pub fn set__is_light_mode__syntax_theme__syntax_set(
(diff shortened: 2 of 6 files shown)


dandavison:
Hi @bash, this looks really interesting!

I don't think I quite have it working in my set up. It is correctly detecting my light-background terminal (i.e. I have removed all env vars and git config settings that would influence it, and on your branch it is correctly using GitHub theme for a light background whereas on master it's using the dark theme despite my terminal being light). However, when I comment out my alacritty color settings so that my terminal has alacritty's default dark background, it continues to use the GitHub theme. Sorry if I've messed up my testing which is possible.

Do you have any concerns about added startup latency? (Do you have any timings?) This is the main/only thing I want us to be sure about regarding this work. Obviously delta ideally wants to always have imperceptible start up time, even on not so high-end machines.

As a tangential comment / for follow up, I think that delta does a few disk I/O things synchronously at start up. I wonder whether there would be latency wins in doing more things concurrently.

I haven't followed the technical details of terminal color detection closely. I did look into it at one point and couldn't quite see / was confused about how delta could do it given that it was accepting stdin from git and outputting stdout to less. In practice, does your comment 

> Only query the terminal if we're not piped to a pager

point to any meaningful limitations? To start with the basic/dumb question, I believe it still works even when delta's output is going to less, right? Can you give me a little bit of hand-holding regarding how this is working at a technical level? Is the point that you're writing to / reading from the tty before delta has started redirecting its output to the pager?

dandavison:
Would you consider opening a PR against bat also (would fix https://github.com/sharkdp/bat/issues/1746) so that both projects benefit and reviewing expertise / problem ironing-out is shared? cc @eth-p

bash:
Thank you @dandavison for testing my changes :)

Since opening this PR, I did some more testing and decided to change a few things:
* I completely removed Windows support from `terminal-colorsaurus` as the state of detecting the terminal's color on Windows is [quite broken](https://github.com/bash/terminal-colorsaurus/blob/main/doc/windows.md).
* Similar to how there's often a `--color=auto/never/always` flag I thought it was a good idea to include a similar flag for the dark/light detection for cases when the heuristic is wrong.
* I noticed during my daily driving that my heuristic (output not redirected) is wrong in the case of `git add --patch`. I decided to piggy-back on the `--color-only` flag that is usually set in that case. (Is this a good idea?)

Prompted by your comment, I decided to collect [some measurements](https://github.com/bash/terminal-colorsaurus/blob/main/doc/latency.md) regarding terminal latency.
As you can see, there is a wide range with some terminals below 100 µs and some that usually take up to 30 ms.

> Can you give me a little bit of hand-holding regarding how this is working at a technical level?

I will happily try to do that. Please note that I have learned a lot of these things over the past few weeks by reading man pages and forum answers, so take everything I say with a grain of salt :)

Querying the terminal for its colors involves communicating with the *controlling terminal*. This is a bidirectional stream that can be acquired by opening `/dev/tty`.
Running a command from a shell without redirecting anything means stdin, stdout and stderr all point to this terminal.

I think this is where the confusion might come from: In the case where all standard streams point to the terminal, one can write to stdout and read the terminal's response from stdin.
However, that's not the goal: We want to read/write to the terminal. In this case stdin/stdout just *happen* to point to the terminal.

Where does the race with the pager come from then? Usually, delta is the one to start the pager *after* processing its settings / querying the terminal for its colors. In that case we don't get a race.

The race occurs when the user manually pipes delta's output to a pager, e.g. by running `git diff | delta | less`.
In this case, both delta and less run simultaneously and there are two ways to cause a race:
* less reads from the terminal to detect if the user has pressed `q` to exit. So any response that the terminal tries to send to delta might get partially consumed by less instead!
* Both less and `terminal-colorsaurus` enable / disable raw mode on the terminal. `terminal-colorsaurus` might disable raw mode while less is still running and depending on it being enabled.

dandavison:
Hi @bash, thanks a lot for the explanation (and sorry for the high-latency conversation). I'm still finding this:

> I don't think I quite have it working in my set up. It is correctly detecting my light-background terminal (i.e. I have removed all env vars and git config settings that would influence it, and on your branch it is correctly using GitHub theme for a light background whereas on master it's using the dark theme despite my terminal being light). However, when I comment out my alacritty color settings so that my terminal has alacritty's default dark background, it continues to use the GitHub theme.

(Would you mind rebasing / merging `main`?)

bash:
Hi @dandavison

Sure thing, I have rebased the PR :)

Regarding your setup: Are you using tmux by any chance? In that case tmux is the one to answer the color query. tmux in turn queries the terminal but iirc caches the result. So if the terminal changes its colors during runtime tmux will unfortunately not know about that.

dandavison:
Ah-ha, that's right, I'm using tmux. Do you think they might be amenable to a PR allowing that caching to be optionally disabled? tmux isn't released often but I assume lots of people (like me) install it from source -- we have to for hyperlink support currently -- so having a patch to enable your terminal color detection would very useful. Also, I think the bat project would benefit from your work just as much as delta (and their architecture w.r.t. the interaction between the Rust app and the pager etc is essentially identical to delta's because delta copied what they do). https://github.com/sharkdp/bat/issues/1746 is their issue that your work would seem to solve.

src/options/theme.rs
```diff
@@ -98,15 +100,91 @@ fn get_is_light_mode_and_syntax_theme_name(
...
+
+fn should_detect_dark_light(opt: &cli::Opt) -> bool {
+    match opt.detect_dark_light {
+        // Querying the terminal for its colors requires "exclusive" access
+        // to the terminal since we read/write from the terminal and enable/disable raw mode.
+        // This causes race conditions with pagers such as less when they are attached to the
+        // same terminal as us.
+        //
+        // This is just an issue when the user manually pipes the output to less, e.g. `git diff | delta | less`.
+        // If we're the ones to start less, there's no race condition since less is started *after* we're done here.
```
dandavison, line 127: Would you mind moving this very helpful comment into the `--help` output? It answers the inevitable question "why wouldn't I want `always`?". I'm happy for the `--help` output to be long -- it's essentially a man page, and we have abbreviated output via the short form `-h`.

src/cli.rs
```diff
@@ -294,6 +294,12 @@ pub struct Opt {
     /// set this in per-repository git config (.git/config)
     pub default_language: Option<String>,
 
+    /// Detect whether or not the terminal is dark or light by querying for its colors.
+    ///
+    /// Ignored if either `--dark` or `--light` is specified.
+    #[arg(long = "detect-dark-light", value_enum, default_value_t = DetectDarkLight::default())]
+    pub detect_dark_light: DetectDarkLight,
```
dandavison, line 301: Nice, I like the name. My first thought would have been something-something-terminal-something, but this increases coherency by tying it to the existing `light` and `dark` options.

bash:
> Do you think they might be amenable to a PR allowing that caching to be optionally disabled?

I think they already react to color changes, but the only terminal that I know of with a mechanism to notify processes of color changes is iTerm 2 [^1]. [Here](https://github.com/tmux/tmux/issues/3582)'s the tmux issue for reference.

> Also, I think the bat project would benefit from your work just as much as delta [...]

Thanks for the heads-up :) That's a good idea, I'll try to prepare a PR for bat too.

[^1]: A few weeks ago I submitted [a patch](https://gitlab.gnome.org/GNOME/vte/-/issues/2740) to vte (the terminal library that GNOME's terminal(s) uses) that adds support for the same mechanism that iTerm 2 uses. They didn't seem happy about the specific mechanism used and I haven't had time yet to consider the alternatives. I have no idea what the stance of other terminal implementors on this.

src/options/theme.rs
bash: Sure thing! I did a tiny bit of rewriting (mainly replaced us/we with delta).

> we have abbreviated output via the short form `-h`.

TIL 👀

dandavison:
Hi @bash, I see in your crate the code comment says
> Windows is [not supported][windows_unsupported].

and above you wrote

> It uses OSC 10 / OSC 11 on Unix and winapi calls on Windows.

How will this PR behave for people using delta on Windows?

bash:
I see that the tests are failing on Windows, was this already the case before or did I break them?

dandavison:
I've merged this, thanks a lot @bash for the work! (including all the clear explanations and code comments).

(The Windows test failure isn't your fault; there's a race condition in some tests that access env vars but are running concurrently that hasn't been fixed yet)

dandavison:
Nice, thanks @aykevl -- it's really useful to have some extra confirmation on different setups for this sort of feature.

ccoVeille:
thank you everyone

PR #1629, opened 2024-02-18
Fix headers of modified binary files

### LINKED ISSUE #1621: 🐛 Paths not updated for a diff of modified binary file
I have a commit in which a binary file is modified, and a new text file is added.

If I diff only the binary file, everything is good:

```diff
diff --git a/meta/archive.zip b/meta/archive.zip
index 00de669..d47cd84 100644
Binary files a/meta/archive.zip and b/meta/archive.zip differ
```

image

But if I diff it together with the text file, the label for the modified binary file becomes incorrect:

```diff
diff --git a/meta/added.txt b/meta/added.txt
new file mode 100644
index 0000000..b5eab54
--- /dev/null
+++ b/meta/added.txt
@@ -0,0 +1 @@
+This file was not here before.
diff --git a/meta/archive.zip b/meta/archive.zip
index 00de669..d47cd84 100644
Binary files a/meta/archive.zip and b/meta/archive.zip differ
```

image

I have built `delta` from the source, commit e208f4e, but tested also for tag `0.16.5` (03f1569) with the same outcome.

May be related to #96 and #1502.

imbrish:
All tests pass, but as this is my first time working with Rust and my first contribution here, it would be great if someone more familiar with the codebase could take a closer look 😉

The diff:

```diff
diff --git a/foo b/bar
similarity index 100%
rename from foo
rename to bar
diff --git a/other b/other
index 00de669..d47cd84 100644
Binary files a/other and b/other differ
```

Now:

image

When merged:

image

Fixes #1621.

### DIFF
--- a/src/handlers/diff_header.rs
+++ b/src/handlers/diff_header.rs
@@ -357,7 +357,7 @@ fn parse_diff_header_line(line: &str, git_diff_name: bool) -> (String, FileEvent
 
 /// Given input like "diff --git a/src/my file.rs b/src/my file.rs"
 /// return Some("src/my file.rs")
-fn get_repeated_file_path_from_diff_line(line: &str) -> Option<String> {
+pub fn get_repeated_file_path_from_diff_line(line: &str) -> Option<String> {
     if let Some(line) = line.strip_prefix("diff --git ") {
         let line: Vec<&str> = line.graphemes(true).collect();
         let midpoint = line.len() / 2;
--- a/src/handlers/diff_header_diff.rs
+++ b/src/handlers/diff_header_diff.rs
@@ -1,4 +1,5 @@
 use crate::delta::{DiffType, InMergeConflict, MergeParents, State, StateMachine};
+use crate::handlers::diff_header::{get_repeated_file_path_from_diff_line, FileEvent};
 
 impl<'a> StateMachine<'a> {
     #[inline]
@@ -25,6 +26,19 @@ impl<'a> StateMachine<'a> {
         self.handle_pending_line_with_diff_name()?;
         self.handled_diff_header_header_line_file_pair = None;
         self.diff_line = self.line.clone();
+
+        // Pre-fill header fields from the diff line. For added, removed or renamed files
+        // these are updated precisely on actual header minus and header plus lines.
+        // But for modified binary files which are not added, removed or renamed, there
+        // are no minus and plus lines. Without the code below, in such cases the file names
+        // would remain unchanged from the previous diff, or empty for the very first diff.
+        let name = get_repeated_file_path_from_diff_line(&self.line).unwrap_or_default();
+        self.minus_file = name.clone();
+        self.plus_file = name.clone();
+        self.minus_file_event = FileEvent::Change;
+        self.plus_file_event = FileEvent::Change;
+        self.current_file_pair = Some((self.minus_file.clone(), self.plus_file.clone()));
+
         if !self.should_skip_line() {
             self.emit_line_unchanged()?;
         }
--- a/src/handlers/diff_header_misc.rs
+++ b/src/handlers/diff_header_misc.rs
@@ -12,25 +12,26 @@ impl<'a> StateMachine<'a> {
     }
 
     pub fn handle_diff_header_misc_line(&mut self) -> std::io::Result<bool> {
-        let is_binary: bool = self.test_diff_is_binary();
-        let file_missing: bool = self.test_diff_file_missing();
-
-        if !file_missing && !is_binary {
+        if !self.test_diff_file_missing() && !self.test_diff_is_binary() {
             return Ok(false);
         }
 
-        if is_binary {
-            match (self.minus_file.as_str(), self.plus_file.as_str()) {
-                ("", "") => {
-                    return self.handle_additional_cases(match self.state {
-                        State::DiffHeader(_) => self.state.clone(),
-                        _ => State::DiffHeader(DiffType::Unified),
-                    });
-                }
-                ("/dev/null", _) => self.plus_file.push_str(" (binary file)"),
-                (_, "/dev/null") => self.minus_file.push_str(" (binary file)"),
-                (_, _) => (),
-            };
+        if self.test_diff_is_binary() {
+            // Print the "Binary files" line verbatim, if there was no "diff" line, or it
+            // listed different files but was not followed by header minus and plus lines.
+            // This can happen in output of standalone diff or git diff --no-index.
+            if self.minus_file.is_empty() && self.plus_file.is_empty() {
+                self.emit_line_unchanged()?;
+                self.handled_diff_header_header_line_file_pair = self.current_file_pair.clone();
+                return Ok(true);
+            }
+
+            if self.minus_file != "/dev/null" {
+                self.minus_file.push_str(" (binary file)");
+            }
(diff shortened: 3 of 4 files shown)


src/handlers/diff_header_diff.rs
```diff
@@ -25,6 +26,15 @@ impl<'a> StateMachine<'a> {
...
         self.handled_diff_header_header_line_file_pair = None;
         self.diff_line = self.line.clone();
+
+        // Fill header information from the "diff" line, may be overwritten later.
+        let name = get_repeated_file_path_from_diff_line(&self.diff_line).unwrap_or_default();
+        self.minus_file = name.clone();
+        self.plus_file = name.clone();
+        self.minus_file_event = FileEvent::Change;
+        self.plus_file_event = FileEvent::Change;
+        self.current_file_pair = Some((self.minus_file.clone(), self.plus_file.clone()));
```
dandavison, line 36: Hi @imbrish, thanks for working on this! I'll admit to being rusty on the details of how this all works. Would you (while it's fresh in your head) be able to add some explanatory comments to this PR regarding the implementation? For example here -- my first reaction as maintainer is "Hm, OK so why are we setting all these to implement a feature regarding binary files and yet these changes are not behind any sort of `is_binary` flag".

imbrish: Sure. This part is the crux of the fix. Diff of modified binary files does not have the `---`/`+++` or `deleted file mode`/`new file mode` lines, and thus the `minus_file` and `plus_file` were never updated for such diffs, which caused the path from the previous diff to be shown again (as in the example). As for why setting all these... I am not sure, I have just seen the same in https://github.com/dandavison/delta/blob/main/src/handlers/diff_header.rs#L168-L172 and re-used that. I have just realized that the code there could possibly be removed, as now we do the same thing here upfront. I will try that later and explain the reasoning in the comment. I am just unsure, why it wasn't done like that in the first place. Do you suspect what could have been the reason for the current complexity?

src/handlers/diff_header.rs
```diff
@@ -357,7 +357,7 @@ fn parse_diff_header_line(line: &str, git_diff_name: bool) -> (String, FileEvent
 
 /// Given input like "diff --git a/src/my file.rs b/src/my file.rs"
 /// return Some("src/my file.rs")
-fn get_repeated_file_path_from_diff_line(line: &str) -> Option<String> {
+pub fn get_repeated_file_path_from_diff_line(line: &str) -> Option<String> {
```
imbrish, line 360: I made this public because I use this code in another file. Possibly could be moved there altogether.

src/handlers/diff_header_misc.rs
```diff
@@ -13,24 +13,18 @@ impl<'a> StateMachine<'a> {
...
-                (_, "/dev/null") => self.minus_file.push_str(" (binary file)"),
-                (_, _) => (),
-            };
+        if is_binary && !(self.minus_file.is_empty() && self.plus_file.is_empty()) {
+            if self.minus_file != "/dev/null" {
+                self.minus_file.push_str(" (binary file)");
+            }
+            if self.plus_file != "/dev/null" {
+                self.plus_file.push_str(" (binary file)");
+            }
```
imbrish, line 27: Here I have originally added the `(binary file)` for the `(_, _)` case for completeness (not strictly related) but then realized the whole `match` could be rewritten in a simpler way and decided to go ahead with that, as it took me a while to understand what this was doing (at first I thought something was wrong here).

src/tests/test_example_diffs.rs
```diff
@@ -240,7 +240,7 @@ mod tests {
         let config = integration_test_utils::make_config_from_args(&[]);
         let output = integration_test_utils::run_delta(BINARY_FILES_DIFFER, &config);
         let output = strip_ansi_codes(&output);
-        assert!(output.contains("Binary files a/foo and b/foo differ\n"));
+        assert!(output.contains("foo (binary file)\n"));
```
imbrish, line 243: Aligned output for modified binary files with the added and removed cases.

src/handlers/diff_header_diff.rs
dandavison: Thanks!

> I am just unsure, why it wasn't done like that in the first place. Do you suspect what could have been the reason for the current complexity?

Hm, I don't have it all paged in but it looks like this PR added those lines: https://github.com/dandavison/delta/pull/945

imbrish: I have now tried removing the lines below, but realized they are still needed.

https://github.com/dandavison/delta/blob/084a6d19b2054f8778c375862f8e4ed3d9145120/src/handlers/diff_header.rs#L163-L182

This is because for both added and removed files, the diff line will be `diff --git a/foo b/foo` so these must be recognized by looking at the `new file mode` or `deleted file mode` lines.

imbrish:
Hey @dandavison,

I have reviewed both #945 and #1502 which touched the same code, made slight modifications, and squashed the commits.

The main fix was to pre-fill the header fields (minus and plus files etc.) already when at the `diff` line in [diff_header_diff.rs](https://github.com/dandavison/delta/blob/main/src/handlers/diff_header_diff.rs) These fields are later updated precisely on actual header minus and header plus lines, but it fixes the header for modified binary files, which do not have the minus and plus lines. This is now also described in the code comment.

Another change was a simplification of [diff_header_misc.rs](https://github.com/dandavison/delta/blob/main/src/handlers/diff_header_misc.rs), which now correctly adds the `(binary file)` suffix also for the modified binaries. I presume the abovementioned omission was the reason it bailed when both filenames were empty, and this is no longer necessary.

I have also updated the tests to include the handled cases and made some slight changes to make them more reliable.

One caveat is that the `get_repeated_file_path_from_diff_line` function returns the path only when the `a/` and `b/` are identical. But I do not think that will be a problem, because in my tests this is always the case apart from renames, which come with their own `rename from` and `rename to` lines.

This is now ready to merge and will close #1621.

imbrish:
I have now realized the reason for a special treatment of a case with both minus and plus files empty in [diff_header_misc.rs](https://github.com/dandavison/delta/pull/1629/files#diff-a04c4675f03bc9bb9fda375d2ab1592ebdba4734585309b4bd3037622a5e2ba9). It seems to be parsing of the output of standalone diff, but this was missing tests. I have decided to handle it slightly differently and pass through the `Binary files ... differ` line verbatim, to align output with the `Files ... are identical` line.

Before:

image

After:

image

This is now ready to merge and will close #1621.

imbrish:
@dandavison noticed you started the checks, but I have found an omission when using `git diff --no-index`, will fix it in a minute.

imbrish:
Done! Here is what was fixed.

Calling `git diff --no-index --no-prefix "foo bar" "sub dir/foo bar baz"` generates a following diff:

```
diff --git foo bar sub dir/foo bar baz
index 329fbf5..481817c 100644
Binary files foo bar and sub dir/foo bar baz differ
```

There is no reliable way to determine the minus and plus files in such cases, and thus I have decided to align the output here with that of standalone diff as described above:

image

When the `a/` and `b/` markers are present the files could be recognized in `diff_header_diff.rs`, but as this is such a minor case, let us leave that for another day, and not delay this pull request any further 👍

dandavison:
Excellent! Thanks very much for this work @imbrish.

Issue #1083, opened 2022-05-20
🐛 Incorrect output when grepping in a branch with "/" and "-" in the name


cffswb:
## Context

OS: Linux
Delta revision: 427c9aa971f3

## Description

Tested in `delta` repository.

Run `git grep` alone:
`$ GIT_CONFIG_GLOBAL=/dev/null git grep theme origin/gh-pages`
![image](https://user-images.githubusercontent.com/6070469/169503081-b74fc203-6ef6-44c6-9153-8fcdf769a50f.png)

Run `git grep` + `delta`:
`$ GIT_CONFIG_GLOBAL=/dev/null git -c pager.grep='delta --no-gitconfig' grep theme origin/gh-pages`
![image](https://user-images.githubusercontent.com/6070469/169503181-5b4370e7-47ed-44da-9a59-e663d12d4aeb.png)
The `gh-pages`es become `gh:pages`es, and the `theme`s aren't highlighted.

dandavison:
Thanks @cffswb, will look into this.

By the way, until this is fixed, I just want to mention that piping `rg --json` into delta should never have parsing bugs (since delta receives a JSON format rather than having to parse grep output, which can't be parsed unambiguously, although I don't think we should be confusing an internal `-` character in a file name if that's what's happening here)

mvf:
This seems to also affect paths whose element just above the file has a `-` in it. Here's a shot of `git grep hostmakedepends` from the root of https://www.github.com/void-linux/void-packages. `cyrus-sasl` and `d-feet` are mangled:

![delta2](https://user-images.githubusercontent.com/304760/195181464-cc017e2f-ed04-470e-9199-93905e47ff85.png)

The same command from the `srcpkgs` directory leaves `d-feet` intact, but the highlighting isn't applied anymore:

![delta](https://user-images.githubusercontent.com/304760/195180245-2c36496f-8d3f-4b31-9e38-9559de8bcf46.png)

This seems to happen when the path element with the `-` comes first and only has a single character before the first `-`.

dandavison:
Hi all, currently the best answer here is to use `rg --json` to avoid any parse errors. Fundamentally the issue here is that traditional grep output is not unambiguously parseable. It would be nice if someone added JSON output to `git grep`. Copying from the other dup issues #1631 and #1259:

The delta manual contains an entry addressing this:

[...]

PR #1634, opened 2024-02-24
Parse filename unambiguously using color escape sequences

joshtriplett:
Parse filename unambiguously using color escape sequences, to handle cases where separators like `-` appear in the filename.

`git grep`, by default, emits ANSI color escape sequences around the filename, separator, and line number. Parse these if available.

Add tests for filenames that previously failed to parse correctly.

Fixes: #1083

---

Before:
delta-before

After:
delta-after

---

This currently assumes the default colors, and will fall back to the previous parsing if any of the colors have been changed. This is an issue elsewhere in delta as well; for instance, if `color.grep.match` has been changed then delta won't highlight matches. I think this would be easy to fix, by parsing git's color configuration (which I implemented in [`anstyle-git`](https://crates.io/crates/anstyle-git)), rendering those color sequences to ANSI escapes, and using those in the regex rather than hardcoded ones. However, I think that should be a separate PR.

### DIFF
--- a/src/handlers/grep.rs
+++ b/src/handlers/grep.rs
@@ -95,9 +95,18 @@ impl<'a> StateMachine<'a> {
             return Ok(false);
         }
 
-        let line = self.line.clone(); // TODO: avoid clone
-        let Some(grep_line) = parse_grep_line(&line) else {
-            return Ok(false);
+        // Try parse_raw_grep_line on raw_line, and fall back to parse_grep_line
+        let raw_line = self.raw_line.clone(); // TODO: avoid clone
+        let line;
+        let grep_line = if let Some(grep_line) = parse_raw_grep_line(&raw_line) {
+            grep_line
+        } else {
+            line = self.line.clone(); // TODO: avoid clone
+            if let Some(grep_line) = parse_grep_line(&line) {
+                grep_line
+            } else {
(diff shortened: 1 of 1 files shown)


dandavison:
Brilliant! This is a great advance. I had switched to using `rg` exclusively but I know people will want to use `git grep` and other greps. (And the `git grep -W` option is interesting; delta tries to be helpful with it in combination with `--navigate`, allowing jumping between search hits with n/N)

imbrish:
Hey @joshtriplett, thanks for this.

I was also thinking about reading custom `git grep` colors from `.gitconfig` to parse the output correctly.

Do you plan a follow up pull request for that?

dandavison:
If anyone does work on reading colors from `.gitconfig`, I just want to point out this hack that's been in the codebase for a long time. What it's doing is checking whether git has added any colors other than the default added/removed colors; if the answer is "yes" then we output a raw line e.g. because this might be git outputting "color-moved" colors. The hack is that the added/removed colors are hardcoded as red/green. Just mentioning in case it can also be made more robust (although I don't think anyone's ever complained).

https://github.com/dandavison/delta/blob/dcae5bcc2428d1fb6f5ff7b9cddd7f268d9a3735/src/style.rs#L359-L366

PR #1638, opened 2024-02-25
Switch from vte to anstyle-parse (already used in dependencies)

joshtriplett:
delta already depends indirectly on anstyle-parse through clap. Switch
from vte to anstyle-parse to eliminate a few dependencies.


dandavison:
Thanks! The anstyle project seems like a good development. I think I had heard of it a while ago but hadn't got around to seeing how we should be using it in delta.

PR #1652, opened 2024-03-10
Add `corvus` theme

evilwaveforms:
Hey, heres a dark theme without syntax highlighting that works well with black terminal bg :bird:

`git log -p`


dandavison:
Thanks! I like this.

PR #1661, opened 2024-03-18
Option to detect the dark theme from the OS instead of terminal

kornelski:
I've added an alternative light/dark theme detection mechanism for macOS. It checks the global theme setting, which doesn't require any special terminal support, assuming the terminal application follows the OS theme too.


kornelski:
The reason is that it's completely terminal-independent, so it can work when piping to a non-terminal, or when the terminal itself doesn't support detection.

PR #1672, opened 2024-04-12
Update syntect to 5.2.0

timhillgit:
Created with `cargo update syntect`. I'd like for delta to use the newest version of syntect so it can parse the `gutterSettings` property of themes: https://github.com/trishume/syntect/issues/469


dandavison:
Thanks!

PR #1681, opened 2024-05-01
tests: prevent parallel env var access

th1000s:
`env::set_var` is not-yet-unsafe, and here tests fail because a var is already set by a second test while the first one is still running.

`cargo test -- test_env` with sufficient (auto detected) parallelism triggers this.


dandavison:
Thanks!

Issue #1644, opened 2024-03-01
allow to completely omit zero ("same") lines (like `diff -U0` would do)


calestyo:
### Discussed in https://github.com/dandavison/delta/discussions/1620

<div type='discussions-op-text'>

<sup>Originally posted by **calestyo** February  8, 2024</sup>
Hey.

I found no way to to disable the "surrounding" unmodified lines. `omit` doesn’t seem to work with `--zero-style` and there doesn’t seem to be a counterpart to `diff`’s `-U0`.

Is this somehow possible?

Thanks,
Chris.</div>

dandavison:
Hi Chris, ok I'm sorry to say something that sounds silly, but isn't the answer `-U0`?? As in `git show -U0`? Or is the issue that you want this when doing `delta file_1 file_2`?

calestyo:
Hey Dan.

Oh yes, I was thinking about the `delta file_1 file_2` case.

But your answer made me realise that I could at least get the behaviour via `diff -U0 a b | delta`.<br>
Nevertheless, would be nicer if it would work straight from delta.

dandavison:
Right, my feeling has always been that it would be a mess to try to add the various diff options (actually git diff options) into delta's command-line parsing. `delta a b` is actually[1] a shorthand for `git diff --no-index a b` (a rather little-known command that allows `git diff` to be used on any arbitrary pair of files) and so my stock answer here is to say "would you mind just using `git diff --no-index a b ... | delta` ?" and admitting that perhaps I should never have implemented the `delta a b` shorthand! 


[1] of course, this may have started out true but is not quite true any longer... we implemented support for things like `delta <(echo a) <(echo b)` which is an improvement over
```
$ git diff --no-index <(echo a) <(echo b)
error: /dev/fd/11: unsupported file type
fatal: cannot hash /dev/fd/11
```

dandavison:
To justify "a mess" -- it's that delta at heart is really a tool that takes in diff / grep / blame input and outputs unparseable stuff to look at. And it has a pretty large collection of options for doing that! So one hesitates before adding options for _generating_ the diff output.

calestyo:
> Right, my feeling has always been that it would be a mess to try to add the various diff options (actually git diff options) into delta's command-line parsing.

One way would perhaps be to add a single option, that allows passing any options to the actually used `diff` backend, be it `git diff` (which probably then also uses `diff`) or as in my example above `diff` itself.<br>
[...]

PR #1697, opened 2024-05-07
Implement --diff-args

dandavison:
Allow extra arguments to `git diff` to be passed from the delta commands line, as suggested by @calestyo in https://github.com/dandavison/delta/issues/1644#issuecomment-1974396731

E.g.

```
delta --diff-args=-U99 /tmp/a /tmp/b
delta -@=-U99 /tmp/a /tmp/b
```

Note that the `=` is required in this case, due to the `-` that comes next.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -166,6 +166,21 @@ pub struct Opt {
     #[arg(long = "detect-dark-light", value_enum, default_value_t = DetectDarkLight::default())]
     pub detect_dark_light: DetectDarkLight,
 
+    #[arg(
+        long = "diff-args",
+        short = '@',
+        default_value = "",
+        value_name = "STRING"
+    )]
+    /// Arguments to pass to `git diff` when using delta to diff two files.
+    ///
+    /// E.g. `delta --diff-args=-U999 file_1 file_2` is equivalent to
+    /// `git diff -U999 --no-index --color file_1 file_2 | delta`.
+    ///
+    /// However, if you use process substitution instead of real file paths, it falls back to `diff -u` instead of `git
+    /// diff`.
+    pub diff_args: String,
+
     #[arg(long = "diff-highlight")]
     /// Emulate diff-highlight.
     ///
@@ -960,12 +975,12 @@ pub struct Opt {
     /// Deprecated: use --true-color.
     pub _24_bit_color: Option<String>,
 
-    /// First file to be compared when delta is being used in diff mode
+    /// First file to be compared when delta is being used to diff two files.
     ///
     /// `delta file1 file2` is equivalent to `diff -u file1 file2 | delta`.
     pub minus_file: Option<PathBuf>,
 
-    /// Second file to be compared when delta is being used in diff mode.
+    /// Second file to be compared when delta is being used to diff two files.
     pub plus_file: Option<PathBuf>,
 
     #[arg(skip)]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -172,13 +172,13 @@ pub struct Opt {
         default_value = "",
         value_name = "STRING"
     )]
-    /// Arguments to pass to `git diff` when using delta to diff two files.
+    /// Extra arguments to pass to `git diff` when using delta to diff two files.
     ///
     /// E.g. `delta --diff-args=-U999 file_1 file_2` is equivalent to
-    /// `git diff -U999 --no-index --color file_1 file_2 | delta`.
+    /// `git diff --no-index --color -U999 file_1 file_2 | delta`.
     ///
-    /// However, if you use process substitution instead of real file paths, it falls back to `diff -u` instead of `git
-    /// diff`.
+    /// However, if you use process substitution (`delta <(command_1) <(command_2)`)
+    /// instead of real file paths, it falls back to `diff -U3` instead of `git diff`.
     pub diff_args: String,
 
     #[arg(long = "diff-highlight")]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -177,8 +177,8 @@ pub struct Opt {
     /// E.g. `delta --diff-args=-U999 file_1 file_2` is equivalent to
     /// `git diff --no-index --color -U999 file_1 file_2 | delta`.
     ///
-    /// However, if you use process substitution (`delta <(command_1) <(command_2)`)
-    /// instead of real file paths, it falls back to `diff -U3` instead of `git diff`.
+    /// If you use process substitution (`delta <(command_1) <(command_2)`) and your git version
+    /// doesn't support it, then delta will fall back to `diff` instead of `git diff`.
     pub diff_args: String,
 
     #[arg(long = "diff-highlight")]
--- a/src/subcommands/diff.rs
+++ b/src/subcommands/diff.rs
@@ -24,12 +24,14 @@ pub fn diff(
         |f: &Path| f.starts_with("/proc/self/fd/") || f.starts_with("/dev/fd/");
 
     let diff_cmd = if via_process_substitution(minus_file) || via_process_substitution(plus_file) {
-        ["diff", "-u", "--"].as_slice()
+        format!("diff -u {} --", config.diff_args)
     } else {
-        ["git", "diff", "--no-index", "--color", "--"].as_slice()
+        format!("git diff --no-index --color {} --", config.diff_args)
     };
 
-    let diff_bin = diff_cmd[0];
+    let mut diff_cmd = diff_cmd.split_whitespace();
+    let diff_bin = diff_cmd.next().unwrap();
+
     let diff_path = match grep_cli::resolve_binary(PathBuf::from(diff_bin)) {
         Ok(path) => path,
         Err(err) => {
@@ -39,7 +41,7 @@ pub fn diff(
     };
 
     let diff_process = process::Command::new(diff_path)
-        .args(&diff_cmd[1..])
+        .args(diff_cmd)
         .args([minus_file, plus_file])
         .stdout(process::Stdio::piped())
         .spawn();
--- a/src/subcommands/diff.rs
+++ b/src/subcommands/diff.rs
@@ -7,7 +7,8 @@ use bytelines::ByteLinesReader;
 use crate::config::{self, delta_unreachable};
 use crate::delta;
 
-/// Run `git diff` on the files provided on the command line and display the output.
+/// Run `git diff` on the files provided on the command line and display the output. Fall back to
+/// `diff` if the supplied "files" use process substitution.
 pub fn diff(
     minus_file: &Path,
     plus_file: &Path,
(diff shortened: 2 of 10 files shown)


src/cli.rs
```diff
@@ -317,6 +317,21 @@ pub struct Opt {
...
+    )]
+    /// Arguments to pass to `git diff` when using delta to diff two files.
+    ///
+    /// E.g. `delta --diff-args=-U999 file_1 file_2` is equivalent to
+    /// `git diff -U999 --no-index --color file_1 file_2 | delta`.
+    ///
+    /// However, if you use process substitution instead of real file paths, it falls back to `diff -u` instead of `git
+    /// diff`.
+    pub diff_args: String,
+
```
th1000s, line 334: That's actually a good use of the unique `-@` argument.

Wording: "Arguments" -> "Extra arguments",  `-U999` comes after `--color`, and give a brief example of process substitution:

... process substitution (`delta "$(grep a file_1)" "$(cmd_2)"`) instead of ...

src/subcommands/diff.rs
```diff
@@ -24,12 +24,14 @@ pub fn diff(
         |f: &Path| f.starts_with("/proc/self/fd/") || f.starts_with("/dev/fd/");
 
     let diff_cmd = if via_process_substitution(minus_file) || via_process_substitution(plus_file) {
-        ["diff", "-u", "--"].as_slice()
+        format!("diff -u {} --", config.diff_args)
     } else {
```
th1000s, line 28: diff's `-u` (defaults to 3 context) will override a _following_ `-U<N>` (GNU diffutils at least), so replace with `-U3` which can be overwritten (and adapt help output above).

src/subcommands/diff.rs
```diff
@@ -39,7 +41,7 @@ pub fn diff(
     };
 
     let diff_process = process::Command::new(diff_path)
-        .args(&diff_cmd[1..])
+        .args(diff_cmd)
         .args([minus_file, plus_file])
         .stdout(process::Stdio::piped())
         .spawn();
```
th1000s, line 47: `git diff --no-index --nonsense` will exit with 129 and first print `error: unknown option 'nonsense'` and then the _full_ `--help` output. Maybe catch such cases (`diff` will just exit with 2) and print the actual (quoted, in case `split_whitespace()` was the culprit) command `delta` called.

src/subcommands/diff.rs
dandavison: OK, I will change this. For the record, I couldn't reproduce that on my Mac with `diff (GNU diffutils) 3.10` from homebrew. With that (and the `diff` that ships with Mac) `diff -u -U99` gives the same result as `diff -U3 -U99`.

th1000s: You're right, `-U3 -U4` works fine. Upon further inspection it seems that any previously set unified context can only be _increased_, so `-u` (3) then `-U4` works fine, but `-U3 -U1` or `-u -U1` will leave the context at 3, not set it to 1. The diff args will have to be scanned for any options that set the unified context :face_with_diagonal_mouth:

dandavison:
One issue is that I see that git _does_ now support process substitution https://stackoverflow.com/questions/22706714/why-does-git-diff-not-work-with-process-substitution Which is great, but now it would be nice to improve the fallback logic. We could detect the git version as we do for less, or we could try `git diff` and fall back to `diff` in the case of a certain error exit code and `via_process_substitution(file)` returning true. 

An example of an error:
```
$ delta -@-U-1 <(cat b.rs) src/paint.rs
diff: invalid context length '-1'
diff: Try 'diff --help' for more information.
'diff' process failed with exit status 2. Command was: diff -U3 -U-1 -- /dev/fd/11 src/paint.rs
```

src/subcommands/diff.rs
dandavison: I haven't done this yet. This could also be where we catch people using git <2.43 with process substitution and fallback to diff?

src/subcommands/diff.rs
```diff
@@ -76,17 +99,97 @@ pub fn diff(
...
+            && code == 128
+            && (via_process_substitution(minus_file) || via_process_substitution(plus_file))
+        {
+            // https://stackoverflow.com/questions/22706714/why-does-git-diff-not-work-with-process-substitution
+            // When called as `delta <(echo foo) <(echo bar)`, then git < 2.42 just prints the diff of the
+            // filenames which were created by the process substitution and does not read their content.
+
+            // It looks like `git diff` failed due to lack of process substitution (version <2.42);
+            // try again with `diff`.
+            diff(minus_file, plus_file, Differ::Diff, config, writer)
```
dandavison, line 118: @th1000s I think this PR is almost there. I would like the following to be true:

> if a user has `git >= 2.42` then delta will use `git diff`, even if they use process substitution.

(So that their git config is honored and they can use all the `git diff` options).

Instead of trying to detect the user's git version, the current state of this PR always tries `git diff` and falls back to `diff` if it seems to have failed due to process substitution. After all, this problem should become rarer and rarer as people upgrade git.

This seems to be working on the two Mac laptops I tested on (one with old git and one new), since with the old git I am getting an _error_ (not merely printing of filenames):

```
$ git diff --no-index <(cat [b.rs](http://b.rs/)) src/[paint.rs](http://paint.rs/)
error: /dev/fd/11: unsupported file type
fatal: cannot hash /dev/fd/11
$ echo $?
128
$ git --version
git version 2.39.3 (Apple Git-146)
```

However it sounds like from your comment that you encountered incorrect output with a success (< 2) exit code:

> When called as `delta <(echo foo) <(echo bar)`, then git < 2.42 just prints the diff of the filenames which were created by the process substitution and does not read their content.

If so, then I think my PR needs to do a bit more.

Any thoughts? (and any chance you can test out this PR??)

src/subcommands/diff.rs
dandavison: Done, see https://github.com/dandavison/delta/pull/1697/files#r1703205451

src/cli.rs
dandavison: Thanks, done.

src/subcommands/diff.rs
dandavison: Added detection of whether diff args imply that context is set by user and added tests.

src/subcommands/diff.rs
```diff
@@ -76,17 +99,97 @@ pub fn diff(
...
             eprintln!("'{diff_bin}' process terminated without exit status.");
             config.error_exit_code
-        })
+        });
+
+    if code >= 2 {
+        let via_process_substitution =
+            |f: &Path| f.starts_with("/proc/self/fd/") || f.starts_with("/dev/fd/");
+        let is_git_diff = matches!(differ, Differ::GitDiff);
+        if is_git_diff
```
th1000s, line 108: Why not `#[derive(PartialEq)]` (and `Debug` for good measure)? Then just `==` is possible.

src/subcommands/diff.rs
th1000s: Indeed, on Linux git sees e.g. `/proc/self/fd/16`, then reads and diffs nonsense like `pipe:[15699060]`. But I would just document that this requires git v2.42 and not call and parse `git --version` first.

dandavison: > Indeed, on Linux git sees e.g. `/proc/self/fd/16`, then reads and diffs nonsense like `pipe:[15699060]`. But I would just document that this requires git v2.42 and not call and parse `git --version` first.

Just to make sure I understand, are you saying we should 

1. just not bother calling diff at all, now that git supports process substitution? If the user is trying process substitution then they should use a recent git that supports it.

or

2. that we should now always invoke git twice, i.e. the first time call git to parse the version and the second time call git or diff to compute the diff?

th1000s: I'd be okay with 1. because this will work everywhere.. eventually. But if 2. is not too much work now (and I remember that parsing `less --version` takes just 4ms) that would be nicer of course.

dandavison:
This isn't very well tested, the excuse being the dependency on the presence and installed versions of external software.

Some commands I've used in local testing

```
git show HEAD~100:src/paint.rs > b.rs
```

```
delta --diff-args='-U1 --no-color' <(cat b.rs) src/paint.rs
delta <(cat b.rs) src/paint.rs
```

```
# This works on a machine with git >= 2.42, since it calls git diff,
# which tolerates a negative value
delta --diff-args='-U-1 --no-color' <(cat b.rs) src/paint.rs

# But on a machine with 2.39.3 it yields the following,
# because the process substitution cause it to use diff, which does not tolerate
# the negative value

'diff' process failed with exit status 2. Command was: diff -U3 -U-1 --no-color -- /dev/fd/11 src/[paint.rs](http://paint.rs/) 

(exit code 2)
```

```
$ delta -@='--no-index --nonsense' <(cat b.rs) src/paint.rs
error: unknown option `nonsense'
'git' process failed with exit status 129. Command was: git diff --no-index --color --no-index --nonsense -- /dev/fd/11 src/paint.rs
```

src/utils/git.rs
```diff
@@ -0,0 +1,35 @@
+use std::process::Command;
+
+pub fn retrieve_git_version() -> Option<(usize, usize)> {
+    if let Ok(git_path) = grep_cli::resolve_binary("git") {
+        let cmd = Command::new(git_path).arg("--version").output().ok()?;
+        parse_git_version(&cmd.stdout)
+    } else {
+        None
+    }
+}
```
dandavison, line 10: FWIW, this takes 6.5ms on an M2 Mac (hyperfine's estimate based on 1000 calls in a loop), so slightly slower than the 4ms you found for `less`. (But, if neither file uses process substitution then it's not done at all)

src/subcommands/diff.rs
dandavison: Done, thanks.

src/subcommands/diff.rs
dandavison: Switched to parsing the git version. I think this PR is ready now.

th1000s:
Looks good. One mistake I made when testing was using `-@U1` instead of `-@-U1`, maybe the former can be silently converted, all (git) diff arguments other than the files start with a dash anyhow.

```sh
$ delta -@U1 <(cat a) b
diff: extra operand 'b'
diff: Try 'diff --help' for more information.
'diff' process failed with exit status 2. Command was: diff -U3 U1 -- /proc/self/fd/16 b
```

dandavison:
> One mistake I made when testing was using `-@U1` instead of `-@-U1`, maybe the former can be silently converted, all (git) diff arguments other than the files start with a dash anyhow.

Good call, that seems like a nice usability improvement. Done. (For the first argument only)

dandavison:
There's still this minor mystery, as was the case 3 years ago: https://github.com/dandavison/delta/pull/546#issuecomment-835852373

We have to skip this case

```
#[case("/dev/null", "/dev/null", ExpectDiff::No)]
```

Because it fails on `(ubuntu-latest, i686-unknown-linux-gnu, true)`, but passes on all other builds.

I.e., apparently, on `(ubuntu-latest, i686-unknown-linux-gnu, true)` our test sees `diff /dev/null /dev/null` returning exit status 1.

dandavison:
One final change: it now works if `git` isn't available. This means we're doing `git` version lookup in all code paths. (Incidentally, I was misled by the name of `grep_cli::resolve_binary` -- it does nothing on non-Windows).

th1000s:
.. all code paths _when diffing_, that seems okay. And we will just leave the `/dev/null` mystery for later :)

dandavison:
Great, thanks for all the reviewing and discussion @th1000s.

PR #1728, opened 2024-06-23
Don't read git files when --no-gitconfig is given + unused variables

th1000s:
* Don't read git files when `--no-gitconfig` is given
    
    Despite starting delta with `--no-gitconfig`, strace still shows
    that `.git/config` etc. files are accessed. By no longer doing
    that it is very clear that no config options are read from there.

*    Fix unused variables


ccoVeille:
I don't think many people expect you, and I and a few other crazy people are using `strace -e file -ff` to check what is used.

But cheers, I like it!

dandavison:
Thanks @th1000s! (Not sure what's up with the `macos-latest` unit tests)

PR #1737, opened 2024-07-04
Fix panic when blame-palette is empty

thorio:
Currently, when `blame-palette` is set to an empty string, the application panics:
```
$ git blame ARCHITECTURE.md | delta --blame-palette ""
thread 'main' panicked at src/handlers/blame.rs:175:47:
attempt to calculate the remainder with a divisor of zero
```

This patch replaces this with a nicer error message:
```
$ git blame ARCHITECTURE.md | delta --blame-palette ""
Option 'blame-palette' must not be empty.
```

### DIFF
--- a/src/handlers/blame.rs
+++ b/src/handlers/blame.rs
@@ -173,10 +173,6 @@ impl<'a> StateMachine<'a> {
         let n_keys = self.blame_key_colors.len();
         let n_colors = self.config.blame_palette.len();
 
-        if n_colors == 0 {
-            fatal("Option 'blame-palette' must not be empty.")
-        }
-
         let color = self.config.blame_palette[n_keys % n_colors].clone();
         if Some(color.as_str()) != other_than_color {
             color
--- a/src/handlers/blame.rs
+++ b/src/handlers/blame.rs
@@ -172,7 +172,6 @@ impl<'a> StateMachine<'a> {
     fn get_next_color(&self, other_than_color: Option<&str>) -> String {
         let n_keys = self.blame_key_colors.len();
         let n_colors = self.config.blame_palette.len();
-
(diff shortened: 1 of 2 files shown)


src/handlers/blame.rs
```diff
@@ -172,6 +172,11 @@ impl<'a> StateMachine<'a> {
     fn get_next_color(&self, other_than_color: Option<&str>) -> String {
         let n_keys = self.blame_key_colors.len();
         let n_colors = self.config.blame_palette.len();
+
+        if n_colors == 0 {
+            fatal("Option 'blame-palette' must not be empty.")
+        }
```
dandavison, line 178: Thanks @thorio! I'm thinking that this should go earlier since it is essentially user input validation. Perhaps in `src/config.rs` where we first parse the argument. (Or maybe `clap` has a "value parsing" feature that we can use?)

thorio: I thought about that, but since the actual color is only parsed _after_ this point, I just stuck the check here for simplicity.

Ideally clap would parse the color and then store it in a struct or enum guaranteeing its validity, but that would be a larger change.

As for just moving it earlier in the execution, I think right after [this line](https://github.com/dandavison/delta/blob/d7a22a7cf9cd5692928d41856662fc2abf1549b0/src/config.rs#L215) would be a good spot.

src/handlers/blame.rs
```diff
@@ -172,6 +172,7 @@ impl<'a> StateMachine<'a> {
     fn get_next_color(&self, other_than_color: Option<&str>) -> String {
         let n_keys = self.blame_key_colors.len();
         let n_colors = self.config.blame_palette.len();
+
```
dandavison, line 175:
```suggestion
```

dandavison:
Thanks!

PR #1739, opened 2024-07-06
tests: add insta for snapshot testing

th1000s:
Insta makes writing new tests or performing changes which update the expected value(s) easier. The new reference data can be reviewed and inserted/updated automatically by `cargo insta review` (after `cargo install cargo-insta`, but that tool is optional).

The snapshots are stored inline using the `@""` syntax and not in separate files because the delta test output is small and designed to be human readabe.

See https://insta.rs/#hello-snapshot-testing
and https://docs.rs/insta/1.39.0/insta/

---

Example workflow:

image

*No, the diff there is not produced by delta :(*

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -515,6 +515,7 @@ dependencies = [
  "dirs",
  "git2",
  "grep-cli",
+ "insta",
  "itertools",
  "lazy_static",
  "palette",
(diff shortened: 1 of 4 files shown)


dandavison:
Great, LGTM! Will merge when builds finish.

Issue #1738, opened 2024-07-05
🐛 `relative-paths` option doesn't work for binary files


stefan-muc:
Add two files (one text, one binary) in the root of a git repo (`/d/temp`) to git cache, create a subdirectory, `cd` there and do a `git diff --cached`, you will get this output:
```text
$ pwd
/d/temp/subdir
$ git diff --cached

added: README.md
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

─────┐
• 1: │
─────┘
Hello World

added: image.png (binary file)
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

Now set `git config delta.relative-paths true` and you will get this output:
```text
$ git diff --cached

Δ \dev\null ⟶   ..\README.md
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

─────┐
• 1: │
─────┘
Hello World

added: image.png (binary file)
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

Directory structure
```text
./image.png
./README.md
./subdir <-- cwd
```

`\dev\null` gets inserted because of #1538
`..\README.md` is correct
`image.png` is wrong, should be `..\image.png` - delta uses absolute path here, but should use relative path!

Delta version `0.17.0`, can reproduce on Windows 11 and Linux

th1000s:
Indeed, the paths were not updated in case of binary files. Fixed in #1740

Issue #1538, opened 2023-09-05
🐛 \dev\null shown for new file when relative paths are enabled


Destroy666x:
- [x] Please include the raw text output from git, so that we can reproduce the problem:
```
Δ \dev\null ➔ ..\test.md
```
- [x] A screenshot of Delta's output is often helpful also:
![image](https://github.com/dandavison/delta/assets/3606072/817a1bd1-3e1e-429e-b4a4-6d0b26ff0349)

If you enable `git config --global delta.relative-paths true`, the there's unnecessary `/dev/null` and arrow to the left of added files, at least for `git diff --cached` on Windows 10 when the path is relative.

PR #1740, opened 2024-07-06
Make relative-paths work with binary files

th1000s:
*  Make relative-paths work with binary files
       
    `relativize_path_maybe()` was not called in this case. Added
    test and converted a few existing ones to insta.

* Make relativize_path_maybe directly update the path argument
    
    This simplifies the call sites.
    Also skip updating if the path is already absolute, and handle
    '/dev/null' on Windows so it is no longer converted to '\dev\null'.

*  gitignore: add pending insta snapshots

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ performance/index.html
 src/junk.rs
 **/*.rs.bk
 /target
+
+# insta: unreviewed inline data, and separate snapshot files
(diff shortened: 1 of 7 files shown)


th1000s:
Fixes #1738 and #1538

dandavison:
Great, thanks!

PR #1741, opened 2024-07-07
tests: set terminal width to 43

th1000s:
Make test more deterministic by not allowing them
to "see" the current terminal width.


th1000s:
@bash I think something similar might be helpful for the terminal color detection, this also "sees" the current terminal (`test_full_filename_used_to_detect_language` fails because of that). Otherwise `DetectLightModeOverride` has to be used a lot more.

dandavison:
LGTM!

PR #1745, opened 2024-07-09
Update `catppuccin-latte` casing in `LIGHT_SYNTAX_THEMES` list

injust:
Also sort `LIGHT_SYNTAX_THEMES`


dandavison:
Thanks! Maybe we should make theme logic more case-insensitive? (non-blocking comment, orthogonal to this PR)

injust:
That sounds reasonable.

I didn't look at the code, but is it viable to infer whether a theme is light/dark from the `.gitconfig` value, i.e. `catppuccin.gitconfig` in this case?

PR #1761, opened 2024-07-16
Add truncate_str_short() which always returns a prefix of the input

th1000s:
Document that the input of `truncate_str()` may be cut on a double width char ("fullwidth") - then the output is no longer a prefix of the input.

`truncate_str_short()` could be faster by not constructing the result but only returning an index, but it will be used in rare edge cases so reusing existing logic is good enough.

Common code moved to `truncate_str_impl()`.

Assume (with fallback) that graphemes are at most 2 wide.
Fewer allocations.
Tests.

### DIFF
--- a/src/ansi/mod.rs
+++ b/src/ansi/mod.rs
@@ -25,24 +25,19 @@ pub fn measure_text_width(s: &str) -> usize {
     })
 }
 
-/// Truncate string such that `tail` is present as a suffix, preceded by as much of `s` as can be
-/// displayed in the requested width.
-// Return string constructed as follows:
-// 1. `display_width` characters are available. If the string fits, return it.
-//
-// 2. Contribute graphemes and ANSI escape sequences from `tail` until either (1) `tail` is
-//    exhausted, or (2) the display width of the result would exceed `display_width`.
-//
-// 3. If tail was exhausted, then contribute graphemes and ANSI escape sequences from `s` until the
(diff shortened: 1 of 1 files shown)


src/ansi/mod.rs
```diff
@@ -51,20 +46,60 @@ pub fn truncate_str<'a>(s: &'a str, display_width: usize, tail: &str) -> Cow<'a,
...
+pub fn truncate_str<'a>(s: &'a str, display_width: usize, tail: &str) -> Cow<'a, str> {
+    truncate_str_impl(s, display_width, tail, Some(' '))
+}
+
+/// Truncate string `s` so it fits into `display_width`, ignoring any ANSI escape sequences when
+/// calculating the width. If a double-width ("fullwidth") grapheme has to be cut, it is omitted and
+/// the resulting string is *shorter* than `display_width`. But this way the result is always a
+/// prefix of the input `s`.
+pub fn truncate_str_short(s: &str, display_width: usize) -> Cow<str> {
+    truncate_str_impl(s, display_width, "", None)
```
dandavison, line 102: clippy wants `#[allow(dead_code)]` here

th1000s: I'll just merge it with the  --max-syntax-highlighting-length PR :)

th1000s:
Now contained in PR #1746

PR #1762, opened 2024-07-16
Make less version >= 633 behave like previous versions wrt. Nerd Fonts

### LINKED ISSUE #1616: 🐛 delta causes fancy unicode symbols such as Nerd Font Icons to be displayed as escaped in angle brackets
```bash
$ git clone https://github.com/vorburger/vorburger-dotfiles-bin-etc.git
$ git checkout 975d4c977d41d8dd096067fa6d0588edee2c9fc7
$ ./fonts-install.sh
$ git --no-pager log | head -12
```

Screenshot from 2024-01-27 20-07-41

Note the funky "font symbol" I've used in those 2 commit messages (the fancy 'A' - see it?), this is from https://github.com/ryanoasis/nerd-fonts/, which my `./fonts-install.sh` that's invoked above installed (if you then also appropriately configure to use that [patched!] *Fira Code **Nerd** Font* in your Terminal).

Now check out what `delta` seems to do:

```bash
$ git --no-pager log | delta

commit 975d4c977d41d8dd096067fa6d0588edee2c9fc7 (HEAD -> main, origin/main, origin/HEAD)
Author: Michael Vorburger <SPAM@vorburger.ch>
Date:   9 minutes ago

    Fonts: Install <U+F031> Fonts in *-gui.sh instead of *install.sh

commit 7f9b65a35974881c0712f1fab386bb8edf9a6f8d
Author: Michael Vorburger <SPAM@vorburger.ch>
Date:   32 minutes ago

    Fonts: Fira Code != Fira Code NERD <U+F031>
```

So that  "fancy" 'A' got replaced by `<U+F031>`- even though my (appropriately configured) Terminal could display it.

Is there a way to configure `delta` to pass it through? For reference, check out this out for good ol' `less`:

```bash
$ git --no-pager log | less
commit 975d4c977d41d8dd096067fa6d0588edee2c9fc7 (HEAD -> main, origin/main, origin/HEAD)
Author: Michael Vorburger <SPAM@vorburger.ch>
Date:   11 minutes ago

    Fonts: Install <U+F031> Fonts in *-gui.sh instead of *install.sh

commit 7f9b65a35974881c0712f1fab386bb8edf9a6f8d
Author: Michael Vorburger <SPAM@vorburger.ch>
Date:   34 minutes ago

    Fonts: Fira Code != Fira Code NERD <U+F031>
```

so same, it also does this (unwanted) "escaping", but `less --raw-control-chars` does the trick:

```bash
$ git --no-pager log | less --raw-control-chars
```

Screenshot from 2024-01-27 20-14-21

Of course, one would want to have the cake and eat it too, and still have fancy "ANSI coloring" as well... 😄

th1000s:
Sets `LESSUTFCHARDEF` (unless already present), a new env var introduced by this less version to always print characters from the Private Use Area of Unicode as-is, even if the terminal does not understand them.

---

Fixes #1616

Better valus for `LESSUTFCHARDEF` may be provided later, see [nerd-fonts/ryanoasis#1337](https://github.com/ryanoasis/nerd-fonts/issues/1337).

### DIFF
--- a/src/utils/bat/output.rs
+++ b/src/utils/bat/output.rs
@@ -21,6 +21,7 @@ pub enum PagingMode {
     #[default]
     Never,
 }
+const LESSUTFCHARDEF: &str = "LESSUTFCHARDEF";
 use crate::errors::*;
 
 pub enum OutputType {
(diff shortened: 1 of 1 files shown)


dandavison:
LGTM. The fix here is in code originally vendored from bat. cc @keith-hall, @Enselic in case it's useful to bat also.

PR #1766, opened 2024-07-17
avoid negative repeat_count fixes #1760

### LINKED ISSUE #1760: 🐛 Crash while doing git diff HEAD..FETCH_HEAD on latest delta code running on MacOS
- [ ] Please include the raw text output from git, so that we can reproduce the problem.
      (You can use `git --no-pager` to produce the raw text output.)
- [ ] A screenshot of Delta's output is often helpful also.

Thanks for filing a Delta bug report!
`
thread 'main' panicked at library/alloc/src/raw_vec.rs:25:5:
capacity overflow
stack backtrace:
   0:        0x100acedb0 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h7fda799bfa435f0f
   1:        0x100adb2d0 - core::fmt::write::h149c43f8c649a127
   2:        0x100abf880 - std::io::Write::write_fmt::hc658f220ee6401ee
   3:        0x100acec04 - std::sys_common::backtrace::print::hd320cb51e23d85e7
   4:        0x100aca6f8 - std::panicking::default_hook::{{closure}}::h2a019217d59e304e
   5:        0x100aca414 - std::panicking::default_hook::h7115745ac0da2b42
   6:        0x100acaa94 - std::panicking::rust_panic_with_hook::hcaae50c6a8446a46
   7:        0x100acf160 - std::panicking::begin_panic_handler::{{closure}}::h3511e01278c0eee1
   8:        0x100acefdc - std::sys_common::backtrace::__rust_end_short_backtrace::hb40ac48f98b88c7f
   9:        0x100aca7dc - _rust_begin_unwind
  10:        0x100af9a5c - core::panicking::panic_fmt::hd81673e1b116ffee
  11:        0x100ada570 - alloc::raw_vec::capacity_overflow::he177c4fad30c3064
  12:        0x100af99f4 - alloc::raw_vec::handle_error::h814061ff56cfdcc0
  13:        0x100ada494 - alloc::slice::<impl [T]>::repeat::heea357c6bbc4f473
  14:        0x1008b47fc - delta::paint::Painter::paint_lines::h2270367d410b0326
  15:        0x1008b439c - delta::paint::Painter::paint_zero_line::h6c548bfff5cbc41f
  16:        0x1008da244 - delta::handlers::hunk::<impl delta::delta::StateMachine>::handle_hunk_line::h2d0a5765982fdc3b
  17:        0x1008d4100 - delta::delta::delta::hab6a5759a13c6797
  18:        0x1008b9d78 - delta::main::h5909ee6f2ee73102
  19:        0x100909d88 - std::sys_common::backtrace::__rust_begin_short_backtrace::h7d1b12fed2fc7b08
  20:        0x1008df6dc - std::rt::lang_start::{{closure}}::h92cf623c35598587
  21:        0x100aca730 - std::panicking::try::hd72586cf9ddb09aa
  22:        0x100ab809c - std::rt::lang_start_internal::ha2b1bf612bcbfcfb
  23:        0x1008bae70 - _main
`

lrfurtado:


src/paint.rs
```diff
@@ -258,10 +258,14 @@ impl<'p> Painter<'p> {
...
+                let mut repeat_count:usize = 0;
+                if text_width < config.available_terminal_width {
+                    repeat_count  = config.available_terminal_width - text_width;
+                }
                 line.push_str(
                     #[allow(clippy::unnecessary_to_owned)]
                     &fill_style
-                        .paint(" ".repeat(config.available_terminal_width - text_width))
+                        .paint(" ".repeat(repeat_count))
                         .to_string(),
```
th1000s, line 269: You can use [saturating_sub()](https://doc.rust-lang.org/std/primitive.i16.html#method.saturating_sub) directly there.

th1000s:
Nice find, interesting that this unsigned underflow was never triggered before.

Can you provide a minimal example diff which triggers this so it can be called with `cat diff.txt | delta --no-gitconfig --width=X`? I assume the current terminal width is a contributing factor, so `--width` should make it independent of that. Then a test with insta's `assert_snapshot!` can be written to ensure that never happens again.

lrfurtado:
```
diff --git a/foobar b/foobar
index 185a33cd5b..123f893935 100755
--- a/foobar
+++ b/foobar
@@ -14,6 +14,11 @@ function UNDERFLOW_EXEC() {
    XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX
 }

+
+echo "=== VALIDATE UNDERFLOW"
+
 # this is an example of the underflow error
 echo "=== UNDERFLOW_EXEC"
 UNDERFLOW_EXEC -i foobar.yaml


```

lrfurtado:
@th1000s  are my latest changes what you had in mind ?

th1000s:
Yes, however when keeping the test and reverting your fix, then running `cargo test test_long_line_underflow -- --nocapture` fails, but does not trigger the crash. But this code path _is_ quite hard to it, I could not do it after trying a few parameters.

I see `XX↵`, so it seems side-by-side with wrapping is involved. Can you reproduce this when piping your example input into the delta (debug build) binary? And can you reproduce this when specifying `--no-gitconfig` and then adding the settings from your git config manually, such as `--side-by-side`? And then, what is your terminal width: `echo $COLUMNS`.

lrfurtado:
without the saturating_sub the test now crashes 
image

lrfurtado:
> @th1000s I think it looks better now

Found something interesting @th1000s --width 111 is not fully insulating the unitest from terminal sizes differences, if I increase or decrease the size of the font on my terminal. The tests start breaking.

src/tests/test_example_diffs.rs
```diff
@@ -2137,6 +2137,30 @@ src/align.rs:71: impl<'a> Alignment<'a> { │
         "###);
     }
 
+    #[test]
+    fn test_long_line_underflow() {
+        let output = DeltaTest::with_args(&["--raw",  "--width" , "111", "--zero-style", "syntax #FFFFFF"]).with_input(GIT_DIFF_LONG_LINE).skip_header();
```
th1000s, line 2142: `--raw` is not needed here (documented as "intended for testing")

src/tests/test_example_diffs.rs
```diff
@@ -3125,4 +3149,30 @@ index 53f98b6..14d6caa 100644
...
++
++
++echo "=== VALIDATE UNDERFLOW"
++/usr/local/bin/check_underflow
++
+ # this is an example of the underflow error
+ echo "=== UNDERFLOW_EXEC"
+ UNDERFLOW_EXEC -i foobar.yaml
+
+"#;
```
th1000s, line 3177: You can leave out almost all of these +- lines, as you found out the context before the changed lines was the cause.

th1000s:
Nice, thank you for the investigation, looks like styling zero hunks is not tested enough. And because of this other bug you found (but is now nicely reproducible) let's just add `#[ignore]` to the `#[test]`.

And when you set the expected output in `assert_snapshot!(..)` try setting your terminal to 111 so it is as it should be. Also, `cargo fmt` will take care of the code formatting.

src/tests/test_example_diffs.rs
lrfurtado: any concerns leaving as is ?

th1000s: Well, it makes the minimal reproducer diff larger than it has to be, and the text sort of indicates that it's these lines. As you have to rebase anyhow to make it build with rust 1.80, maybe you can also trim the added lines down to one or two.

PR #1773, opened 2024-07-28
chore: cleanup brew formula file in favor of core tap

chenrui333:
cleanup the formula file (which is in 0.15.0) in favor of the core one (0.17.0), https://formulae.brew.sh/formula/git-delta


dandavison:
Thanks! I'd been meaning to do that for years, but was vaguely thinking I should deprecate the one in the repo to give people using it a graceful path to using the one in homebrew core. And never got around to it. I agree we can just delete it now!

chenrui333:
Thanks @dandavison!

PR #1779, opened 2024-08-01
chore: update to build with rust 1.80

chenrui333:
seeing some time related failure 

```
error[E0282]: type annotations needed for `Box<_>`
  --> /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.31/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++
```

has been addressed with 0.3.36, https://github.com/time-rs/time/issues/693

also fixed the path borrow check issue

```
error[E0596]: cannot borrow `inner_relativize` as mutable, as it is not declared as mutable
  --> src/utils/path.rs:46:17
   |
41 |             *path = relative_path.to_string_lossy().into_owned();
   |             ----- calling `inner_relativize` requires mutable binding due to mutable borrow of `path`
...
46 |         let _ = inner_relativize();
  1 pick 644ca1f fix borrow check error for `path`¬
  7 # This is a combination of 2 commits.¬
   |                 ^^^^^^^^^^^^^^^^ cannot borrow as mutable
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,9 +10,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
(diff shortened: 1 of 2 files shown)


dandavison:
Thanks a lot @chenrui333! Both you and @th1000s fixed this, and https://github.com/dandavison/delta/commit/f441d4d0839abcc8822e8f2963746c4e641886a4 is already in main. But thanks a lot for the fix.

chenrui333:
Nice, good to know. Thanks!

PR #1787, opened 2024-08-04
Handle ambiguous diff header, '--- ' can also be present in a minus hunk

th1000s:
`diff -u file1 file2` output starts with '--- file1', a '-- 12' line being removed results in '--- 12', which was interpreted as the start of a new diff. Now the number of removed lines announced in the header as e.g. '@@ -1,4 +1,3 @@' is taken into account for this specific diff input.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -1,5 +1,6 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
+use std::convert::TryInto;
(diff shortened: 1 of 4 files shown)


dandavison:
Very nice. Delta probably should have made use of the hunk coordinates in the header a long time ago. For example, this PR brings to mind the suggestion made by @fdcds here in 2019: https://github.com/dandavison/delta/issues/60#issuecomment-557485242

PR #1805, opened 2024-08-14
cd: build binaries targeting GNU libc on ubuntu-20.04, not latest

th1000s:
This improves compatibility, as `latest` is 24.04 currently. The statically linked musl build still uses `ubuntu-latest`

---

The binaries built on ubuntu-latest/ubuntu-24.04 might not run on systems with an older libc, so debian stable and maybe even fedora 40 can't run these releases.

But I am not sure how to quickly test this, merge just before making a new release and see? :)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -21,19 +21,19 @@ jobs:
           - os: windows-latest
             target: x86_64-pc-windows-msvc
             use-cross: false
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
             target: x86_64-unknown-linux-gnu
             use-cross: false
           - os: ubuntu-latest
             target: x86_64-unknown-linux-musl
             use-cross: true
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
             target: i686-unknown-linux-gnu
             use-cross: true
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
             target: arm-unknown-linux-gnueabihf
             use-cross: true
-          - os: ubuntu-latest
+          - os: ubuntu-20.04
             target: aarch64-unknown-linux-gnu
             use-cross: true
 


dandavison:
Let me check whether I understand, since this isn't a particularly strong suit of mine:

1. `ubuntu-latest` is `24.04` currently
2. Other than `*-musl`, our `ubuntu-*` builds build dynamically linked binaries, meaning the binaries will link at runtime to whatever libc is on the system where the binary is installed.
3. A dynamically linked binary built on `24.04` will, when installed on an older system such as Debian Stable, fail.
4. Therefore you're proposing to build all our dynamically built binaries on an older ubuntu: `20.04`


Is (3) because the dynamically linked binary has encoded into it a desire for a specific (new) version of libc, i.e. will not be satisfied by an older libc?

Do many projects "pin" their (ubuntu) builds at a more conservative release for this reason? Is it something it would have made sense for us to do in delta years ago?

th1000s:
Yes, it doesn't link against many libraries (libz.so.1, libgcc_s.so.1, libm.so.6 and libc.so.6), but libc always adds a versioned symbol with each release. For example on Debian oldstable delta 0.17 fails to start with the following error. And yes, that is why older Linux systems are usually used to build releases (I have seen ancient CentOS 5/6 for that). Pinning this version would (have been :) be a good idea, especially since GH regularly updates what the `-latest` label points to (see the macOS x86/arm switchover).

```
$ delta
delta: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by delta)
delta: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by delta)
delta: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by delta)
```

dandavison:
> But I am not sure how to quickly test this, merge just before making a new release and see? :)

I note the "just before" there. Is the idea that since we cannot easily test, to minimize the risk to people building from `main`, we merge shortly before a release which is an event that will likely provide complaints if there are problems?

I think we are approaching release time, so shall we merge now?

th1000s:
Those building from main won't use the GH cd pipeline :) - It is rather so that a failing cd isn't caused by a then half forgotten commit.

Issue #1810, opened 2024-08-16
🚀 Automatically update binary URLs in installation section of manual


dandavison:
We are sometimes forgetting on release to update the binary URLs in `manual/src/installation.md`.
See (constantly work-in-progress) `etc/release.Makefile`.

PR #1832, opened 2024-08-28
Update binary URLs in installation section of manual (#1810)

dvermd:
Calling `bump-version-in-documentation` before `create-github-release`  should fix the issue


dvermd:
The documentation will indeed be updated before the release but it is only locally before `create-github-release` wil create the tag and push the commit and push the tag.

What was your thought on addressing this update?

PR #1843, opened 2024-09-04
Allow `--dark` to override dark/light detected from syntax theme

### LINKED ISSUE #1713: 🐛 Dark theme doesn't seem to be working
- [ ] Please include the raw text output from git, so that we can reproduce the problem.
```
❯ git --no-pager blame smallestmultiple/src/main.rs
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  1) // Done using formula a * b = lcm(a*b)*gcd(a*b)
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  2) 
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  3) fn gcd(mut a: i64, mut b: i64) -> i64 {
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  4)     while b != 0 {
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  5)         let temp = b;
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  6)         b = a % b;
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  7)         a = temp;
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  8)     }
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530  9)     a
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 10) }
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 11) 
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 12) fn main() {
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 13)     let mut lcm: i64 = 1;
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 14)     for i in 1..=20 {
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 15)         lcm = lcm * i / gcd(lcm, i);
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 16)     }
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 17)     println!("{}", lcm);
47d46703 (Rahil Prakash 2024-05-21 00:15:37 +0530 18) }
```
image

Delta doesn't seem to be respecting the manually set theme. I've tried this both on tmux and base kitty.
gitconfig for reference
```
[user]
	name = Rahil Prakash
	email = 54015030+Radrahil@users.noreply.github.com

[init]
	defaultBranch = main

[pull]
	rebase = false

[core]
	editor = nvim
  pager = delta

[interactive]

[delta]
    dark = true
    navigate = true

[merge]
    conflictstyle = diff3

[diff]
    colorMoved = default

```

bash:
(Probably) Resolves #1713

Currently a light syntax theme (like `GitHub`) will always set the decorations to light mode too, even if `--dark` is passed.

I took the chance to add missing test cases in `test_syntax_theme_selection` and clean up the match a bit :)

Note that since dark/light detection is done outside `get_is_light_mode_and_syntax_theme_name` a terminal detected to be dark will also win against a light syntax theme. I'm not sure if this is the behavior we want, but it's already the case for light terminal vs dark syntax theme.

### DIFF
--- a/src/options/theme.rs
+++ b/src/options/theme.rs
@@ -162,7 +162,12 @@ mod tests {
             (Some("Nord"), None, "Nord", Dark),
             (None, Some(Dark), DEFAULT_DARK_SYNTAX_THEME, Dark),
             (None, Some(Light), DEFAULT_LIGHT_SYNTAX_THEME, Light),
+            (Some("GitHub"), Some(Light), "GitHub", Light),
+            (Some("GitHub"), Some(Dark), "GitHub", Light), // TODO: This should be Dark.
+            (Some("Nord"), Some(Light), "Nord", Light),
+            (Some("Nord"), Some(Dark), "Nord", Dark),
             (Some("none"), None, "none", Dark),
+            (Some("none"), Some(Dark), "none", Dark),
             (Some("None"), Some(Light), "none", Light),
         ] {
             let mut args = vec![];
--- a/src/options/theme.rs
+++ b/src/options/theme.rs
@@ -159,6 +159,7 @@ mod tests {
         ) in vec![
             (None, None, DEFAULT_DARK_SYNTAX_THEME, Dark),
             (Some("GitHub"), None, "GitHub", Light),
+            (Some("Nord"), None, "Nord", Dark),
             (None, Some(Dark), DEFAULT_DARK_SYNTAX_THEME, Dark),
             (None, Some(Light), DEFAULT_LIGHT_SYNTAX_THEME, Light),
             (Some("none"), None, "none", Dark),
--- a/src/options/theme.rs
+++ b/src/options/theme.rs
@@ -162,11 +162,8 @@ mod tests {
         ) in vec![
             (None, None, DEFAULT_DARK_SYNTAX_THEME, Dark),
             (Some("GitHub"), None, "GitHub", Light),
-            (Some("GitHub"), None, "GitHub", Light),
-            (None, Some(Light), DEFAULT_LIGHT_SYNTAX_THEME, Light),
             (None, Some(Dark), DEFAULT_DARK_SYNTAX_THEME, Dark),
             (None, Some(Light), DEFAULT_LIGHT_SYNTAX_THEME, Light),
-            (None, Some(Light), "GitHub", Light),
             (Some("none"), None, "none", Dark),
             (Some("None"), Some(Light), "none", Light),
         ] {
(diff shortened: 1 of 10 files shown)


src/options/theme.rs
```diff
@@ -86,24 +86,23 @@ fn is_no_syntax_highlighting_syntax_theme_name(theme_name: &str) -> bool {
...
-    get_is_light_opt(opt)
-        .or_else(|| should_detect_dark_light(opt).then(detect_light_mode))
-        .unwrap_or_default()
+fn get_is_light(opt: &cli::Opt) -> Option<bool> {
+    get_is_light_opt(opt).or_else(|| {
+        should_detect_dark_light(opt)
+            .then(detect_light_mode)
+            .flatten()
+    })
 }
```
th1000s, line 106: In my opinion true/false is clear, but a tri-state should get a name, in this case it doesn't even need `PartialEq`. Also `get_is_light_opt()` can be moved into the function.

src/options/theme.rs
```diff
@@ -126,19 +125,18 @@ fn should_detect_dark_light(opt: &cli::Opt) -> bool {
...
 
 #[cfg(not(test))]
-fn detect_light_mode() -> bool {
+fn detect_light_mode() -> Option<bool> {
     use terminal_colorsaurus::{color_scheme, ColorScheme, QueryOptions};
-    color_scheme(QueryOptions::default()).unwrap_or_default() == ColorScheme::Light
+    color_scheme(QueryOptions::default())
```
th1000s, line 130: Because this PR has such a nicely thought out commit list/stack, could you also add a bonus commit to move this `use` up and then do whatever so the test cfg is happy? This non-test crate should not hide here :)

src/options/theme.rs
bash: What would you think about an `enum ColorMode { Dark, Light }`? This could be used in place of the bool everywhere `is_light` appears. I think `Option<ColorMode>` is then pretty clear as well.

bash: I pushed [b33459b](https://github.com/dandavison/delta/pull/1843/commits/b33459b70399d49ce1cddfb145dac5a803cffb84) that implements an enum and replaces the "is light mode" bool everywhere. It's easy to revert if you disagree with the change :)

dandavison: The enum feels like an improvement to me. I think it is a bool mainly because I was very new to Rust when I added it.

th1000s (approved):

PR #1851, opened 2024-09-06
Fix clippy warnings

bash:
I noticed in #1843 that main had some clippy warnings. `src/utils/helpwrap.rs` interestingly didn't show up when I ran `cargo clippy` but got fixed with `cargo clippy --fix` 🤷🏼


dandavison:
Thanks @bash!

PR #1856, opened 2024-09-10
Do not double panic in FakeParentArgs::drop

dvermd:
Check if we're already panicking before calling error which will also panic.

### DIFF
--- a/src/utils/process.rs
+++ b/src/utils/process.rs
@@ -598,6 +598,7 @@ pub mod tests {
         With(usize, Rc<Vec<T>>),
         None,
         Invalid,
+        ErrorAlreadyHandled,
     }
 
     // When calling `FakeParentArgs::get()`, it can return `Some(values)` which were set earlier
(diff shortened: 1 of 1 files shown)


src/utils/process.rs
```diff
@@ -854,9 +859,8 @@ pub mod tests {
     fn test_process_testing_assert_never_used() {
         let _args = FakeParentArgs::once("never used");
 
-        // causes a panic while panicking, so can't test:
-        // let _args = FakeParentArgs::for_scope(&"never used");
-        // let _args = FakeParentArgs::once(&"never used");
+        let _args = FakeParentArgs::for_scope(&"never used");
+        let _args = FakeParentArgs::once(&"never used");
     }
```
th1000s, line 865: Nice. Now that this is testable, this test needs to be split into separate cases. `once()` should panic on drop, even stand-alone as before, because it is never used.

But calling `for_scope` twice should panic in the constructor because the previous state (in the thread local variable) was not cleared out properly (and previously the drop also panicked because it encountered bad state as well).

Actually, couldn't `FakeParentArgs::error` just set the state so that drop doesn't panic again? And to ensure the correct panic is hit maybe the more precise [#[should_panic(expected = "msg here")]](https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic) can be used.

dvermd: I think all `FakeParentArgs` will go through `drop` before `error` being called in this construct.

```
            if FAKE_ARGS.with(|a| a.replace(initial(string_vec))) != TlsState::None {
                Self::error(from_);
            }
```
Thus, we'll only get panics from the `drop` method and none from the constructors

src/utils/process.rs
```diff
@@ -648,24 +649,37 @@ pub mod tests {
...
-            panic!(
-                "test logic error (in {}): wrong FakeParentArgs scope?",
-                where_
-            );
+        fn error(where_: &str) {
+            FAKE_ARGS.with(|a| {
+                let old_value = a.replace(TlsState::Error);
+
+                match old_value {
+                    TlsState::Error => (),
```
th1000s, line 674: Just one final "naming things": I think `ErrorAlreadyHandled` or something like that would be a better name, also see `Drop` where another `Error => nothing` line is.

dvermd: I didn't changed any `ErrorAlreadyHandled => nothing` lines because the only case where FAKE_ARGS is set to `ErrorAlreadyHandled` is in the `error` function just before panicking. 
When we're already in `ErrorAlreadyHandled` state and hit the `ErrorAlreadyHandled => nothing` lines, there's nothing more to do because we're already panicking. 

Did I miss a case ?

th1000s: Looks good, thank you!

PR #1892, opened 2024-11-09
Redundant Option Checks, Unwrap Safety

sharma-shray:
Redundant Option Checks, unwrap Safety, unnecessary Lifetimes, Rust often infers lifetimes automatically


src/parse_styles.rs
```diff
@@ -28,8 +28,14 @@ pub fn parse_styles(opt: &cli::Opt) -> HashMap<String, Style> {
...
-    resolved_styles.get_mut("minus-emph-style").unwrap().is_emph = true;
-    resolved_styles.get_mut("plus-emph-style").unwrap().is_emph = true;
+    resolved_styles
+        .get_mut("minus-emph-style")
+        .expect("minus-emph-style not found in resolved styles")
+        .is_emph = true;
+    resolved_styles
+        .get_mut("plus-emph-style")
+        .expect("plus-emph-style not found in resolved styles")
+        .is_emph = true;
```
dandavison, line 38: The project has a (admittedly idiosyncratic) style of using `unwrap_or_else(|| panic(...))` instead of `expect(..)` because of the confusion regarding how to read and write `expect` messages (e.g. what you wrote vs "minus-emph-style should be in resolved styles"). Would you mind switching to use `unwrap_or_else`?

src/style.rs
```diff
@@ -307,17 +307,11 @@ impl fmt::Debug for AnsiTermStyleEqualityKey {
...
-            if a == b {
-                true
-            } else {
-                ansi_term_16_color_equality(a, b) || ansi_term_16_color_equality(b, a)
-            }
+    match a.zip(b) {
+        Some((a, b)) => {
+            a == b || ansi_term_16_color_equality(a, b) || ansi_term_16_color_equality(b, a)
         }
+        None => a.is_none() && b.is_none(),
```
dandavison, line 314: Ah, nice! I actually hadn't considered that one could zip `Option` values.

src/parse_styles.rs
sharma-shray: hello! sure, corrected

dandavison (approved):
Thanks, LGTM! Will merge.

PR #1896, opened 2024-11-11
Add console commands for git configuration

harmtemolder:
For those of us who find running commands in the console easier than editing files

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,15 @@
     conflictstyle = zdiff3
 ```
 
+Or run:
+
+```sh
+git config --global core.pager delta
+git config --global interactive.diffFilter 'delta --color-only'
+git config --global delta.navigate true
+git config --global merge.conflictStyle zdiff3
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,15 @@
...
 ```
 
+Or run:
+
+```sh
+$ git config --global core.pager delta
+$ git config --global interactive.diffFilter 'delta --color-only'
+$ git config --global delta.navigate true
+$ git config --global merge.conflictStyle zdiff3
+```
```
dandavison, line 45: Hi, thanks for this. Could you remove the pseudo-prompt prefixes so that it's pasteable?

harmtemolder: Sure thing. I actually prefer without as well, but was working on something else this morning where they were included, so guess that rubbed off

dandavison:
Thanks! I agree this makes sense. Even though I'm trying to keep the README minimal, you're right, for lots of people the current README contents would have led to "err, where is my git config?"

PR #1903, opened 2024-11-13
Honor pager path when checking less version

dandavison:


src/utils/bat/less.rs
```diff
@@ -1,12 +1,9 @@
+use std::path::PathBuf;
 use std::process::Command;
 
-pub fn retrieve_less_version() -> Option<usize> {
-    if let Ok(less_path) = grep_cli::resolve_binary("less") {
```
dandavison, line 4: @th1000s do you agree that it appears to be incorrect that here we use a hardcoded `"less"` as the name of the executable to resolve, despite one frame up the stack having obtained `less_path` via `resolve_binary` on the user's passed-in `pager_path`?

th1000s: Yes, your eyes do not deceive you :)

dandavison: Good!

PR #1905, opened 2024-11-13
Unnecessarily Cloning, structuring

sharma-shray:
Hello, minor updates, unnecessarily cloning, structuring


src/options/set.rs
```diff
@@ -79,7 +79,7 @@ pub fn set_options(
     }
     opt.navigate = opt.navigate || opt.env.navigate.is_some();
     if opt.syntax_theme.is_none() {
-        opt.syntax_theme.clone_from(&opt.env.bat_theme);
+        opt.syntax_theme = opt.env.bat_theme.take();
```
dandavison, line 82: While there might be an efficiency saving here, it makes the data inconsistent, so I think we should only make changes like that for demonstrated performance wins.

src/git_config/mod.rs
```diff
@@ -156,21 +156,15 @@ fn parse_config_from_env_var_value(s: &str) -> HashMap<String, String> {
...
-            }
+            let key = captures
+                .get(1)
+                .or_else(|| captures.get(3))
+                .map_or("".to_string(), |m| m.as_str().to_string());
+            let value = captures
+                .get(2)
+                .or_else(|| captures.get(4))
+                .map_or("".to_string(), |m| m.as_str().to_string());
+            (key, value)
```
dandavison, line 167: Would you mind reverting this change? It loses the "joint distribution" of the data: i.e. it's not as obvious from the new code that, for example, they will either _both_ be 0 or neither be 0.

dandavison (approved):
Thanks! I like 3/5 of the changes, but would you mind reverting the other two?

src/handlers/blame.rs
```diff
@@ -233,15 +233,15 @@ $
...
+    let commit = caps.get(1)?.as_str();
+    let author = caps.get(2)?.as_str();
+    let timestamp = caps.get(3)?.as_str();
 
     let time = DateTime::parse_from_str(timestamp, timestamp_format).ok()?;
 
-    let line_number = caps.get(4).unwrap().as_str().parse::<usize>().ok()?;
+    let line_number = caps.get(4)?.as_str().parse::<usize>().ok()?;
 
-    let code = caps.get(5).unwrap().as_str();
```
th1000s, line 244: `unwrap()` here is actually okay because the regex is know at compile time. If it is changed but not the corresponding code here then it _should_ crash for the developer and not silently fail and return None.
Similarly for the max line numbers above.

src/git_config/mod.rs
sharma-shray: sure, corrected

src/handlers/blame.rs
sharma-shray: > `unwrap()` here is actually okay because the regex is know at compile time. If it is changed but not the corresponding code here then it _should_ crash for the developer and not silently fail and return None. Similarly for the max line numbers above.

should we reverting this changes?

dandavison: Yes, I think the only bit of the PR that arguably should be retained is not using `else` after the early return in `src/align.rs` -- I do slightly prefer that. But beyond that, there are lots of bugs and feature requests to work on if you've got time! And any performance improvements backed by profiling of course would be welcome.

th1000s: We could create and tag issues as **🟣 good first issue** for that.

sharma-shray: Thanks for the review, corrected

PR #1909, opened 2024-11-14
Allow multiple hyperlinks per line

### LINKED ISSUE #1907: 🚀 Allow multiple hyperlinks per line
I have hyperlinks set up to make commit hashes link to corresponding pages on gitlab; for the most part it works great, but if there are multiple things resembling commit hashes on a line it seems to only make the last one a link. There are a few scenarios in which this behavior is inconvenient for me:

1. I'm using git range-diff, which displays multiple commit hashes per line, all of which I'd like to become links.
2. I'm using git log with a --pretty format that puts the commit hash and the commit title on the same line, and a commit title mentions another commit's hash (say, "revert commit xyz").
3. I'm using git log with a --pretty format that puts the commit hash and the commit title on the same line, and a commit title mentions something that _isn't_ a commit hash, but the parser thinks is one, like an 8 digit number. (The link to a nonexistent commit would be fine, except that it blocks creation of a useful link.)

Links would be more useful for me if in all of these scenarios it turned both into links (or even 3/4/5/etc. links on one line, if multiple of those things happened together, I suppose).

th1000s:
* Allow multiple hyperlinks per line

Previously only the last commit was linked

Do not link numbers (technically also commits), and stop after
finding 12 commits on a line.

Fixes #1907

---

* testability: add pretty_assertions, git mocks

---

I'd want to enhance all the current assertions with pretty_assertions, that is no work compared to using insta.

### DIFF
--- a/src/features/hyperlinks.rs
+++ b/src/features/hyperlinks.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 use std::path::Path;
 
 use lazy_static::lazy_static;
-use regex::{Captures, Regex};
+use regex::{Match, Matches, Regex};
 
 use crate::config::Config;
 use crate::features::OptionValueFunction;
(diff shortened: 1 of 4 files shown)


src/features/hyperlinks.rs
```diff
@@ -19,26 +19,83 @@ pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
...
+        let mut matches = COMMIT_LINE_REGEX.find_iter(line);
+        if let Some(capture) = matches.next() {
+            let mut pos = 0;
+            let mut result = String::new();
+
+            let mut create_commit_hyperlink = |pre_cap_pos, cap: Match| {
+                result.push_str(&line[pre_cap_pos..cap.start()]);
+                let commit = &line[cap.start()..cap.end()];
+                let formatted_commit =
+                    format_osc8_hyperlink(&commit_link_format.replace("{commit}", commit), commit);
```
th1000s, line 54: What about filtering out pure numbers here again, as mentioned in the issue?

dandavison: Yes, why not. The posterior probability of such a string being a number is going to be considerably higher than that of it being a commit. Multiplied by the probability that someone happened to care about the link to their improbably-digit-rich commit...

src/features/hyperlinks.rs
```diff
@@ -19,26 +19,83 @@ pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
...
+            let mut create_repo_hyperlink = |pre_cap_pos, cap: Match| {
+                result.push_str(&line[pre_cap_pos..cap.start()]);
+                let commit = &line[cap.start()..cap.end()];
+                let formatted_commit = format!(
+                    "{osc}8;;{url}{st}{commit}{osc}8;;{st}",
+                    url = repo.format_commit_url(commit),
+                    commit = commit,
+                    osc = "\x1b]",
+                    st = "\x1b\\"
+                );
```
dandavison, line 83: Is there a reason that we can't use `format_osc8_hyperlink` here also? And beyond that, The two branches of this function are quite substantial and look very similar -- did you decide it wasn't attractive to unite them?

dandavison (approved):
Great! Made a couple of comments but feel free to merge when you're happy with it.

src/features/hyperlinks.rs
th1000s: Done.

src/features/hyperlinks.rs
th1000s: Right, just didn't want think about how to nicely factor that out :) - What about this `HyperlinkCommits` helper now?

src/features/hyperlinks.rs
```diff
@@ -19,26 +19,83 @@ pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
...
+    } else if let Some(repo) = remote_from_config(&config.git_config()) {
+        let mut matches = COMMIT_HASH_REGEX.find_iter(line);
+        if let Some(first_match) = matches.next() {
+            let result = HyperlinkCommits(|commit| {
+                format!(
+                    "{osc}8;;{url}{st}{commit}{osc}8;;{st}",
+                    url = repo.format_commit_url(commit),
+                    commit = commit,
+                    osc = "\x1b]",
+                    st = "\x1b\\"
```
th1000s, line 91: Can be replaced with `format_osc8_hyperlink()`, right?

dandavison: Yes I believe so. I'm not sure why the previous code wasn't already using that. Hopefully just a historical accident rather than a Chesterton's fence.

src/features/hyperlinks.rs
dandavison: Nice, the helper LGTM.

src/features/hyperlinks.rs
th1000s: Done (tests agree)!

PR #1916, opened 2024-11-26
Styled zero lines fix

th1000s:
With `interactive.diffFilter = delta --color-only` delta is called by e.g. `git add -p`, but in this mode git hides the terminal from  the pager.

Plus/minus lines correctly use ANSI sequences to paint up to to the end of the line, but zero lines always used spaces. This needs the terminal width, but it is not available for diffFilter. So the fallback of 80 is used, and zero styles did not extend to the full terminal width.

Since zero lines are only rarely styled (e.g. via
`--zero-style='syntax "#1d1f21" dim'`), this was never noticed.

This also crashed delta when a zero line was longer than 80.

### DIFF
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -204,7 +204,7 @@ impl<'p> Painter<'p> {
                 self.config,
                 &mut self.line_numbers_data.as_mut(),
                 None,
-                BgShouldFill::With(BgFillMethod::Spaces),
+                BgShouldFill::default(),
             );
         }
     }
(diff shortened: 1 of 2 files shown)


th1000s:
~~This unfortunately breaks a bunch of unrelated tests, any idea where the **#**, e.g. `commit 94907c` to `#commit 94907c`, on the first line comes from?~~ I added a `#` somewhere, obviously :) Fixed now.

PR #1917, opened 2024-11-26
Print a notice when delta panics

th1000s:
* Print a notice when delta panics

Setting `DELTA_DEBUG_LOGFILE=crash.log` and repeating the command
writes all information needed to reproduce the crash into that file.

* Add --test-panic

To test the debug handler

----

The notice reads:

```
delta panicked and crashed, sorry :(
To quickly repeat the previous command without delta, run 'export GIT_PAGER=less' first. If you want
to report the crash and it is easy to repeat, do so after 'export DELTA_DEBUG_LOGFILE=crash.log'
then submit the logfile to github at <https://github.com/dandavison/delta/issues>.

!! Make sure there is NO sensitive information in the log file, it will contain the entire diff !!
```


src/utils/debug.rs
```diff
@@ -0,0 +1,195 @@
...
+                .unwrap_or(512 * 1024);
+
+            if std::env::var_os(RUST_BACKTRACE).is_none() {
+                // SAFETY:
+                // a) we only get here when `DELTA_DEBUG_LOGFILE` is set, which means a user is expecting a crash anyhow
+                // b) the requirement is "no other threads concurrently writing or reading(!) [env vars],
+                // other than the ones in this [std::env] module.", the rust backtrace handler should use std::env.
+                unsafe {
+                    std::env::set_var(RUST_BACKTRACE, "1");
+                }
```
th1000s, line 129: Preparing for the 2024 edition :)

src/delta.rs
```diff
@@ -147,7 +148,11 @@ impl<'a> StateMachine<'a> {
     where
         I: BufRead,
     {
+        let mut debug_helper = RecordDeltaCall::new(self.config);
+
         while let Some(Ok(raw_line_bytes)) = lines.next() {
+            debug_helper.write(raw_line_bytes);
+
```
dandavison, line 155: So IIUC, the pattern we're using is

Create the `debug_helper = RecordDeltaCall::new()`; This causes some default diagnostics to be written.

...
Append additional arbitrary diagnostics via `debug_helper.write`
...
Terminate the file in the `Drop` impl for `RecordDeltaCall`.


But if the `DELTA_DEBUG_LOGFILE` is not set, then, although we do a few function calls, they are all no-ops.

src/cli.rs
```diff
@@ -984,6 +984,13 @@ pub struct Opt {
     /// Second file to be compared when delta is being used to diff two files.
     pub plus_file: Option<PathBuf>,
 
+    /*
+    Panic on purpose. When a `diff -u a b` (without a `git diff` header) is encountered
+    then panic early, otherwise panic after processing all input.
+     */
+    #[arg(hide = true, long = "test-panic")]
+    pub test_panic: bool,
+
```
dandavison, line 993: I wonder whether this commit (`Add --test-panic`) is more a development commit that we don't need to merge?

dandavison (approved):
The main commit LGTM (PR needs rebasing). I'm inclined to think that we don't really need the second commit in `main` -- save a little complexity from our code readers's eyes.

src/delta.rs
th1000s: Indeed, `write()` will usually return after an an easily branch-predicted if.

src/utils/debug.rs
```diff
@@ -0,0 +1,201 @@
...
+                    .into(),
+            )?;
+
+            if let Some(cfg) = config.git_config.as_ref() {
+                write("git config values:\n".into())?;
+                cfg.for_each(".*", |entry, value: Option<&str>| {
+                    if !(entry.starts_with("user.")
+                        || entry.starts_with("remote.")
+                        || entry.starts_with("branch.")
+                        || entry.starts_with("gui."))
```
th1000s, line 81: Note: I inverted the logic, now printing all git configs (see `git config --get-regexp '.*'`) except a those listed here.

th1000s (commented):
Dropped the second commit, can be merged.

src/utils/debug.rs
```diff
@@ -0,0 +1,201 @@
...
+impl RecordDeltaCall {
+    pub fn new(config: &Config) -> Self {
+        fn make(logfile: &OsString, config: &Config) -> Result<Box<DeltaDetailInternal>, String> {
+            let mut file = File::create(logfile).map_err(|e| e.to_string())?;
+
+            let mut write = |s: String| file.write_all(s.as_bytes()).map_err(|e| e.to_string());
+
+            write(
+                "<details>\n\
+                <summary>Input which caused delta to crash</summary>\n```\n\n"
```
dandavison, line 71:
```suggestion
                <summary>Input which caused delta to crash</summary>\n\n```\n\n"
```

In practice I believe the GitHub markdown renderer requires one blank line before the triple-backtick. E.g. compare

<details>
<summary>Input which caused delta to crash</summary>

```
hello
```
</details>

with

<details>
<summary>Input which caused delta to crash</summary>
```
hello
```
</details>

th1000s:
The raw lines of course contain the usual escape characters, pasting those into an issue isn't that helpful:

`[33mcommit d10095bc732a7675295d3389152675ccc41e8240[m[33m ([m[1;31morigin/main[m[33m)[m`

This turns the escape chars into `�` there. Can they be recovered by quoting someone? But they might not even be valid UTF-8.

A `crash.log` and a `crash.log.raw` might be an option.

PR #1922, opened 2024-12-08
refactor: avoid using format! when it is unnecessary

Integral-Tech:
- Although `format!` macro is very powerful, it introduces overhead of memory allocation on the heap compared to `&str`. Therefore, when `format!` macro is unnecessary, we should avoid using it.

### DIFF
--- a/src/handlers/commit_meta.rs
+++ b/src/handlers/commit_meta.rs
@@ -47,15 +47,28 @@ impl StateMachine<'_> {
             (Cow::from(&self.line), Cow::from(&self.raw_line))
         };
 
-        draw_fn(
-            self.painter.writer,
-            &format!("{}{}", formatted_line, if pad { " " } else { "" }),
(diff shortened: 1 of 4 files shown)


Integral-Tech:
@th1000s I have just submitted a commit to fix the strange round-trip :)

th1000s:
Thanks! Can you split the `concat!()` fix into a separate commit?

The `match decoration_style { ... }` can be made shorter by returning `(write_boxed*, style)`. These write functions are just of type `fn(...) -> Result<()>`, so no boxing is needed. Then the padding-if-else and calling can all happen in one place, not in every match arm (and then padding could be done for every case).

Also, avoid double negatives, e.g. `never_pad = false`. In fact, add an enum for that, `PadText::{Yes, No}`, and place it right after the two text arguments which are padded. Then the purpose of this argument is clear, even without inlay hints.

PR #1930, opened 2024-12-22
chore(deps): update git2 to use libgit2 1.9

chenrui333:
bump git2 crate to use libgit2 1.9


Cargo.lock
```diff
@@ -1,6 +1,6 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
+version = 4
```
th1000s, line 3: While delta doesn't have a MSRV, this version bump _may_ break older versions. Or do you happen to know when this was introduced, and if delta is already depending on this or a higher version?

chenrui333: > Lockfile v4 has been stable since Rust 1.78. We are making it the default version as of 1.83 nighties unless your MSRV is below 1.78.

https://github.com/rust-lang/cargo/issues/14655#issuecomment-2400237392

chenrui333: I guess we can just stay with lockfile v4? happy to revert to v3 if you like though.

th1000s: 1.78 was more than a year ago, thanks for looking that up. The PR it's fine as-is.

PR #1940, opened 2025-01-11
Cache the Git remote URL to speed up rendering hyperlinks

### LINKED ISSUE #1939: 🐛 The hyperlinks option makes delta 55x slower when used as the pager for git log
The hyperlinks option makes delta 55x slower when used as the pager for `git log` because it calls [`GitConfig::get_remote_url`](https://github.com/dandavison/delta/blob/ef3e1be569bf076f035327342939bd9d7c8908bd/src/git_config/mod.rs#L123-L131)
for each commit hash which is incredibly slow and essentially unusable for repos with many commits.

If possible, `delta` should cache the Git remote URL and re-use it. I would create a PR that does this, but my Rust skills are extremely limited (though I may still try).

### Performance tests

Both of these commands were ran from the root of a recent version of the Linux [source tree](git://git.kernel.dk/linux.git). The below results are from an M4 Mac Pro, but were roughly equal to the results I got on my AMD Linux server. We limit the amount of git log data processed to 32MiB (`time head -c $((32 * 1024 * 1024))`) since processing the full output of `git log` when hyperlinks are enabled takes an very very long time (when processing the full amount the slow down is 66x, which I would consider more accurate).

**Note:** A recent version of [pv](https://www.ivarch.com/programs/pv.shtml) is required to run the below commands since the `--stats` option is a recent addition. Homebrew will install the latest version, but if you're on a Linux distro that does not have a recent version then it can easily be built from source (you'll need to run `autoreconf --force --install --verbose -Wall` to generate the `configure` file if building from the Git repo).

#### Hyperlinks Enabled

```bash
# cd to a repo with a large commit history then run:
$ DELTA_PAGER='bash -c "time head -c $((32 * 1024 * 1024)) | pv --stats --output=/dev/null"' git log
32.0MiB 0:00:43 [ 755KiB/s]
rate min/avg/max/mdev = 731516.382/773784.852/814642.291/16590.908 B/s

real	0m43.400s
user	0m0.183s
sys	0m1.367s
```

#### Hyperlinks Disabled

```bash
# cd to a repo with a large commit history then run:
$ DELTA_PAGER='bash -c "time head -c $((32 * 1024 * 1024)) | pv --stats --output=/dev/null"' git log
32.0MiB 0:00:00 [40.1MiB/s]
rate min/avg/max/mdev = 42006890.474/42006890.474/42006890.474/nan B/s

real	0m0.802s
user	0m0.051s
sys	0m0.379s
```

#### Call Tree
Screenshot 2025-01-10 at 6 51 55 PM

#### Flamegraph
Screenshot 2025-01-10 at 6 54 15 PM

charlievieth:
This commit speeds up the rendering of hyperlinks by ~47x by caching the Git repos remote URL instead of fetching it each time a hyperlink is rendered.

Fixes #1939

### Note

I am very new to Rust so I am very open to feedback and completely understand if you want to rewrite or abandon this entire PR (especially if for some reason the remote URL should not be cached). Additionally, I think the [`GitConfig`](https://github.com/charlievieth/delta/blob/f4f71c15a110c77853c9afc8d77a6600394c55a5/src/git_config/mod.rs#L12-L20) should be responsible for storing the cached remote, but struggled to make that work due to borrowing, lifetimes, mutability, etc.

### Benchmarks

The below benchmarks were ran in the root of [Linux v6.12](https://github.com/torvalds/linux/tree/v6.12). We use `head -c $((128 * 1024 * 1024))` to limit the amount of Git log data since otherwise benchmarking `delta` before this change takes a very long time.

**TLDR:** This change makes processing 128Mib of Git log output take 3.7 seconds instead of 176.1 seconds.

#### Without git remote caching

```bash
$ DELTA_PAGER='bash -euo pipefail -c "time head -c $((128 * 1024 * 1024)) | pv --stats --output=/dev/null"' git -c 'pager.log=delta --hyperlinks' log
 128MiB 0:02:56 [ 744KiB/s]
rate min/avg/max/mdev = 654979.457/764202.554/1108112.678/35667.893 B/s

real	2m56.063s
user	0m0.736s
sys	0m5.530s
```

#### With git remote caching (this change)

```sh
$ PATH="{PATH_TO_NEW_DELTA}:${PATH}"
$ DELTA_PAGER='bash -c "time head -c $((128 * 1024 * 1024)) | pv --stats --output=/dev/null"' \
    git -c 'pager.log=delta --hyperlinks' log
 128MiB 0:00:03 [34.6MiB/s]
rate min/avg/max/mdev = 35979500.033/36287504.626/36646720.334/263641.477 B/s

real	0m3.700s
user	0m0.205s
sys	0m1.690s
```

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -9,6 +9,7 @@ use crate::config::delta_unreachable;
 use crate::config::Config;
 use crate::config::GrepType;
 use crate::features;
+use crate::git_config::GitRemoteRepo;
 use crate::handlers::grep;
 use crate::handlers::hunk_header::{AmbiguousDiffMinusCounter, ParsedHunkHeader};
 use crate::handlers::{self, merge_conflict};
@@ -112,6 +113,7 @@ pub struct StateMachine<'a> {
     pub handled_diff_header_header_line_file_pair: Option<(String, String)>,
     pub blame_key_colors: HashMap<String, String>,
     pub minus_line_counter: AmbiguousDiffMinusCounter,
+    pub git_remote_repo: Option<GitRemoteRepo>,
 }
 
 pub fn delta<I>(lines: ByteLines<I>, writer: &mut dyn Write, config: &Config) -> std::io::Result<()>
@@ -140,6 +142,7 @@ impl<'a> StateMachine<'a> {
             config,
             blame_key_colors: HashMap::new(),
             minus_line_counter: AmbiguousDiffMinusCounter::not_needed(),
+            git_remote_repo: features::hyperlinks::remote_from_config(&config.git_config()),
         }
     }
 
@@ -248,7 +251,7 @@ impl<'a> StateMachine<'a> {
         writeln!(
             self.painter.writer,
             "{}",
-            format_raw_line(&self.raw_line, self.config)
+            format_raw_line(&self.raw_line, self.config, &self.git_remote_repo)
         )?;
         let handled_line = true;
         Ok(handled_line)
@@ -265,9 +268,13 @@ impl<'a> StateMachine<'a> {
 
 /// If output is going to a tty, emit hyperlinks if requested.
 // Although raw output should basically be emitted unaltered, we do this.
-pub fn format_raw_line<'a>(line: &'a str, config: &Config) -> Cow<'a, str> {
+pub fn format_raw_line<'a>(
+    line: &'a str,
+    config: &Config,
+    repo: &Option<GitRemoteRepo>,
+) -> Cow<'a, str> {
     if config.hyperlinks && io::stdout().is_terminal() {
-        features::hyperlinks::format_commit_line_with_osc8_commit_hyperlink(line, config)
+        features::hyperlinks::format_commit_line_with_osc8_commit_hyperlink(line, config, repo)
     } else {
         Cow::from(line)
     }
--- a/src/features/hyperlinks.rs
+++ b/src/features/hyperlinks.rs
@@ -21,9 +21,7 @@ pub fn make_feature() -> Vec<(String, OptionValueFunction)> {
 
 #[cfg(test)]
 pub fn remote_from_config(_: &Option<&GitConfig>) -> Option<GitRemoteRepo> {
-    Some(GitRemoteRepo::GitHub {
-        slug: "dandavison/delta".to_string(),
-    })
+    GitRemoteRepo::for_testing()
 }
 
 #[cfg(not(test))]
@@ -41,6 +39,7 @@ lazy_static! {
 pub fn format_commit_line_with_osc8_commit_hyperlink<'a>(
     line: &'a str,
     config: &Config,
+    cached_repo: &Option<GitRemoteRepo>,
 ) -> Cow<'a, str> {
     // Given matches in a line, m = matches[0] and pos = 0: store line[pos..m.start()] first, then
     // store the T(line[m.start()..m.end()]) match transformation, then set pos = m.end().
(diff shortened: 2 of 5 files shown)


th1000s:
Thanks for reporting this!

Caching is fine, the slowdown is from libgit2 being overly correct and assuming that the remote url could change while the program is running.
The PR can indeed be made simpler by using a `RefCell` to add interior mutability to the `GitConfig`, i.e. `cached_remote: RefCell<Option<Option<String>>>` with an not-yet-cached option layer, and another for the result of `.find_remote(..).ok()` - or maybe a custum enum is clearer.

charlievieth:
@th1000s Took a while until I had some spare cycles to look at this again, but following your guidance I was able to use a `OnceCell` to embed the `GitRemoteRepo` struct into the `GitConfig` struct. Thank you again for your suggestions and would love another review if you have the time.

charlievieth:
@dandavison 1) Thank you for delta - it's incredible 2) Any chance you could enable the tests?

dandavison:
Hi @charlievieth -- tests are running! I'm sorry that I personally am finding so little time to work on delta at the moment, but thank you very much for your work here.

charlievieth:
@dandavison Thank you and I completely understand (I've only recently had the cycles to address my own OSS commits and projects).

The `clippy` [failure](https://github.com/dandavison/delta/actions/runs/16953832997/job/49654943580?pr=1940) that I get on this branch/PR are identical to the one I get when running `make lint` (aka `cargo clippy`) against the main branch (d5e0565cbfa47acde98d41d8777ace5d1bc4d690). I'm fairly new to Rust so happy to look into the failure myself and include it into this branch or create a preceding PR (to merge before this one) that addresses the issue.

For context my rust toolchain is:
* rustc `1.89.0` (29483883e 2025-08-04) (Homebrew)
* clippy `0.1.89`

## Script used to check the delta of `cargo clippy`

```bash
#!/usr/bin/env bash

set -euo pipefail

TMP=$(mktemp -d)
TMP_MAIN="${TMP}/clippy.main.txt"
TMP_HYPER="${TMP}/clippy.hyperlink.txt"

if ! git remote -v | grep -q '^charlievieth'; then
    git remote add charlievieth https://github.com/charlievieth/delta.git
fi
git fetch --all

git checkout main
cargo clippy  &> "${TMP_MAIN}" || true

git checkout cev/hyperlink
cargo clippy  &> "${TMP_HYPER}" || true

(
    cd "${TMP}"
    git diff --no-index "$(basename "${TMP_MAIN}")" "$(basename "${TMP_HYPER}")"
)
````

### Output (nothing applicable AFAIK):
```diff
diff --git a/clippy.main.txt b/clippy.hyperlink.txt
index 65e5421..71ee65e 100644
--- a/clippy.main.txt
+++ b/clippy.hyperlink.txt
@@ -71,4 +71,4 @@ help: use `'_` for type paths
     |                                                                              ++++

 warning: `git-delta` (bin "delta") generated 5 warnings
-    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.68s
+    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.70s
```

dandavison:
If you'd like to deal with the clippy complaints then that would be great! I think the best way would be to open a PR for the clippy-related fixes against `main` and then rebase this on that.

charlievieth:
> If you'd like to deal with the clippy complaints then that would be great! I think the best way would be to open a PR for the clippy-related fixes against `main` and then rebase this on that.

Created https://github.com/dandavison/delta/pull/2038 to address the clippy complaints.

charlievieth:
> > If you'd like to deal with the clippy complaints then that would be great! I think the best way would be to open a PR for the clippy-related fixes against `main` and then rebase this on that.
> 
> Created #2038 to address the clippy complaints.

@dandavison now that https://github.com/dandavison/delta/pull/2038 is merged any chance you can take a look at this PR?

charlievieth:
The clippy error reported here also occurs on the main branch.

th1000s:
Thank you for this improvement, and apologies for the delays. I also fixed the new v1.91 (which was released in the meantime) clippy warning.

charlievieth:
@th1000s thank you for the review and for merging this PR!

Issue #1712, opened 2024-05-19
🐛 Memory allocation error for non root user


rljohnsn:
After install a non root user gets a memory error. Running as root works fine.

```bash
❯ sudo delta
[sudo] password for lance:
The main way to use delta is to configure it as the pager for git: see https://github.com/dandavison/delta#get-started. You can also use delta to diff two files: `delta file_A file_B`.
❯ ls -laF /usr/bin/delta
-rwxr-xr-x 1 root root 7059472 Mar 16 07:28 /usr/bin/delta*
❯ ldd /usr/bin/delta
        linux-vdso.so.1 (0x00007ffc8659b000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f9701de0000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f9701dc0000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9701cd9000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9701ab0000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f97024c2000)
❯ delta
memory allocation of 1970324836974592 bytes failed
[1]    11340 IOT instruction  delta
❯ sudo delta
The main way to use delta is to configure it as the pager for git: see https://github.com/dandavison/delta#get-started. You can also use delta to diff two files: `delta file_A file_B`.
❯ uname -a
Linux DESKTOP-UE7J8VS 5.15.146.1-microsoft-standard-WSL2 #1 SMP Thu Jan 11 04:09:03 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
❯ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.4 LTS
Release:        22.04
Codename:       jammy

```

Windows 11, WSL2, Ubuntu 22.04

dandavison:
Hi @rljohnsn, this memory allocation error has come up several times before in delta issues: https://github.com/dandavison/delta/issues?q=is%3Aissue+%22memory+allocation%22

It seems that it's always something to do with stale syntect syntax definitions. Can you try some of the remedies in those threads, such as installing bat and `bat cache --build`?

ernstki:
In my situation, re-running `bat cache --build` causes the core dump to reoccur consistently.

[...]

PR #1941, opened 2025-01-14
Suggest minimum bat version in manual

ernstki:
- make explicit mention of which bat version should be installed in the "Installation" and "Supported languages and themes" sections of the manual

- add missing "Tips & tricks" page

See dandavison/delta#1712 for context.

### DIFF
--- a/manual/src/installation.md
+++ b/manual/src/installation.md
@@ -97,3 +97,5 @@ Users of older MacOS versions (e.g. 10.11 El Capitan) should install using Homeb
 Behind the scenes, delta uses [`less`](https://www.greenwoodsoftware.com/less/) for paging.
 It's important to have a reasonably recent version of less installed.
 On MacOS, install `less` from Homebrew. For Windows, see [Using Delta on Windows](./tips-and-tricks/using-delta-on-windows.md).
+
+If you use [`bat`](https://github.com/sharkdp/bat) and are running `bat cache --build` to install custom themes or language syntaxes then you should install the same version of bat as specified in the [`Cargo.toml`](https://github.com/dandavison/delta/blob/main/Cargo.toml) for the delta release you're using, otherwise delta will crash with a [memory error](https://github.com/dandavison/delta/issues/1712). The current version of delta does not work with bat v0.18.3 or prior versions.
--- a/manual/src/supported-languages-and-themes.md
+++ b/manual/src/supported-languages-and-themes.md
@@ -7,6 +7,7 @@ To add your own custom color theme, or language, please follow the instructions 
 - [Adding a custom language](https://github.com/sharkdp/bat/#adding-new-syntaxes--language-definitions)
 - [Adding a custom theme](https://github.com/sharkdp/bat/#adding-new-themes)
 
-Delta automatically recognizes custom themes and languages added to bat. You will need to install bat in order to run the `bat cache --build` command.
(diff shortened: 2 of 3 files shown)


ernstki:
Bat at v0.18.3 seems to be where the problems arise.

Screenshot of the output of the below Makefile, demonstrating a failure at the point where it tests bat v0.18.3

[Here's the crude `Makefile`](https://github.com/user-attachments/files/18404982/Makefile.txt) that I used to make this determination.

manual/src/supported-languages-and-themes.md
```diff
@@ -9,4 +9,7 @@ To add your own custom color theme, or language, please follow the instructions
 
 Delta automatically recognizes custom themes and languages added to bat. You will need to install bat in order to run the `bat cache --build` command.
 
+Ideally, the version of bat you install should match the version specified in delta's [`Cargo.toml`](https://github.com/dandavison/delta/blob/main/Cargo.toml) for the delta release you're using. There are [known problems](https://github.com/dandavison/delta/issues/1712) with bat v0.18.3 and below.
```
dandavison, line 12:
```suggestion
Delta automatically recognizes custom themes and languages added to bat. You will need to install bat in order to run the `bat cache --build` command. Ideally, the version of bat you install should match the version specified in delta's [`Cargo.toml`](https://github.com/dandavison/delta/blob/main/Cargo.toml) for the delta release you're using. There are [known problems](https://github.com/dandavison/delta/issues/1712) with bat v0.18.3 and below.
```

manual/src/installation.md
```diff
@@ -97,3 +97,5 @@ Users of older MacOS versions (e.g. 10.11 El Capitan) should install using Homeb
 Behind the scenes, delta uses [`less`](https://www.greenwoodsoftware.com/less/) for paging.
 It's important to have a reasonably recent version of less installed.
 On MacOS, install `less` from Homebrew. For Windows, see [Using Delta on Windows](./tips-and-tricks/using-delta-on-windows.md).
+
+Delta also relies on [`bat`](https://github.com/sharkdp/bat) for [custom themes and syntaxes](./supported-languages-and-themes.md). Ideally, you should install the same version of bat as specified in the [`Cargo.toml`](https://github.com/dandavison/delta/blob/main/Cargo.toml) for the delta release you're using. There are [known problems](https://github.com/dandavison/delta/issues/1712) with v0.18.3 and prior versions of bat.
```
dandavison, line 101: For a long time there's been quite common confusion that delta uses the bat executable somehow. How about this wording:

```suggestion
If you use [`bat`](https://github.com/sharkdp/bat) and are running `bat cache --build` to install custom themes or language syntaxes then you should install the same version of bat as specified in the [`Cargo.toml`](https://github.com/dandavison/delta/blob/main/Cargo.toml) for the delta release you're using, otherwise delta will crash with a [memory error](https://github.com/dandavison/delta/issues/1712). The current version of delta does not work with bat v0.18.3 or prior versions.
```

ernstki: That's good. I considered adding "relies on bat _as a library_" but for some reason thought twice about that.

I get that anything but "the current version" is ancient history for a project that hasn't reached 1.0 yet, but is it significant or worth mentioning that the same problem occurs with every version of delta back to 0.16.5 (the last one to include a musl binary in the release assets that I could easily download and run)?

dandavison: I think what you have is good. Thanks for this PR!

ernstki:
@dandavison Cheers, sir!

PR #1948, opened 2025-02-01
Update bat to version 0.25.0

blinxen:
We also need to update unicode-width to version 0.1.14 This version contains does not contain the breaking change which was republished as 0.2.0.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,9 +100,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.79"
+version = "1.0.97"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


charego:
Do any source files need to be changed to support the addition of dark/light themes in bat 0.25.0? I'm not familiar with the code base, but grepping through the repo shows that `BAT_THEME` is used. The new version adds `BAT_THEME_DARK` and `BAT_THEME_LIGHT`, although there is a bug (https://github.com/sharkdp/bat/pull/3168) with those environment variables in 0.25.0.

blinxen:
To be honest, I am not too sure myself ;D. I just ran the tests and did a couple of smoke tests. Maybe this update needs to be reconsidered.

Issue #1678, opened 2024-04-26
Specify light and dark themes simultaneously, and auto-switch based on system appearance


yam-liu:
I want to specify light and dark themes simultaneously, and auto-switch based on the system's appearance like the default behavior.

bash:
Since delta supports bat's `BAT_THEME` environment variable it might be a good idea to wait for bat to implement this first. I have opened a PR for this a while back: https://github.com/sharkdp/bat/pull/2896.

The mentioned PR for bat adds two new options / environment variables:
* `--theme-dark` / `BAT_THEME_DARK`
* `--theme-light` / `BAT_THEME_LIGHT`

simono:
This is now available in bat. And bat has `--theme auto:system` which chooses the theme based on the system appearance.
I think this would be a nice addition to delta, too 😄

hugovandevliert:
`bat` now listens to `BAT_THEME_DARK` and `BAT_THEME_LIGHT` (see: https://github.com/sharkdp/bat/pull/2896). It would be awesome if this was possible in `delta` too!

ZeroKnight:
Proper support for this would be great, but in the meantime I came up with a **Fish-specific** workaround that abuses `BAT_THEME` and Fish's **handler functions**.

Since Delta will respect `BAT_THEME` if it is set, you can create the following function:

```fish
function _update_bat_theme --description 'Set BAT_THEME based on $fish_terminal_color_theme' --on-variable fish_terminal_color_theme
    switch $fish_terminal_color_theme
        case dark
            set -gx BAT_THEME tokyonight_storm
        case light
            set -gx BAT_THEME tokyonight_day
    end
end
```

Notice the `--on-variable fish_terminal_color_theme`: this makes the function a **handler** that will automatically be run whenever `$fish_terminal_color_theme` changes. Fish sets this variable to either `dark` or `light` based on what your terminal emulator reports (assuming it supports `OSC 10` and `OSC 11` sequences).

You can take this a step further and dynamically set `DELTA_FEATURES` to enable further light/dark-specific settings! Just define a feature for both your light and dark themes and set any Delta settings accordingly.

I have the following in my configuration to set appropriate `diff.colorMoved` colors:

```gitconfig
[delta "tokyonight_storm"]
    map-styles = \
        bold purple => "syntax #574e7b", \
        bold cyan => "syntax #42527b"

[delta "tokyonight_day"]
    map-styles = \
        bold purple => "syntax #c8b4e9", \
[...]

PR #2013, opened 2025-07-13
Update docs to make 3 basic configs match

injust:
~~Defaulting to auto light/dark detection is the smarter choice here. Besides, the commands below omit the `delta.dark = true` setting.~~

The basic config appears in 3 places, so make them match, to consistently de-emphasize auto-detection of colors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Or run:
 git config --global core.pager delta
 git config --global interactive.diffFilter 'delta --color-only'
 git config --global delta.navigate true
+git config --global delta.dark true  # or delta.light true, or omit for auto-detection
 git config --global merge.conflictStyle zdiff3
 ```
 


dandavison:
Hi @injust, I believe we initially encouraged auto-detection in the docs, and I subsequently back-pedalled a bit: https://github.com/dandavison/delta/commit/1fb6f99ec498150f552c1c8b5440152c0c2aade7 I didn't do a good job of linking that commit to an issue but I suspect it's one of the ones where the latency associated with the terminal I/O is huge on Windows.

injust:
I see, that makes sense. In that case, since the basic config appears in 3 places, I'll change the PR to make those match.

(Also turns out the color auto-detection doesn't apply to me because I use Catppuccin, and what I actually need is https://github.com/dandavison/delta/issues/1678.)

injust:
@dandavison Gentle ping, these docs changes should be good to merge.

injust:
Closing due to inactivity

dandavison:
Sorry about that @injust! The changes look great and are much appreciated. If you'd like to reopen the PR I'll merge; but if I don't hear back in a few days I'll commit your changes anyway.

It's certainly true that I haven't found time for delta PR and issue maintenance recently. (On the plus side, delta is stable and used by lots of people). But sorry that it's resulting in contributions being ignored.

injust:
Probably easiest for you to commit the changes. I deleted my fork when I was cleaning up old repos, and I can't restore it without paying for GitHub Pro (wtf GitHub?)

Thanks for the great tool; delta has become an invaluable part of my workflow.

Issue #1517, opened 2023-08-22
🚀Option to remove line at top of hunk


nkh:
Hi, thanks for the software.

I'm considering moving from diff-so-fancy to delta and I have played around to get the colors I want to have

I notice that an extra separation line is displayed before the hunk, is there a way to remove it?

![delta](https://github.com/dandavison/delta/assets/45409/7d003c38-e0af-47a3-b1e2-f84ee2f35ce0)

PR #2028, opened 2025-08-25
feat: add `newlines` option to handle when blank lines are added

### LINKED ISSUE #1944: 🐛 Unable to remove extra blank line preceding file and hunk headers
It seems that delta always inserts a blank line before a file or hunk header.

### Plain git diff

Image

### delta

Image

Note the two blank lines marked with a **red arrow** before hunk headers and one blank line marked with a **blue arrow** before file header.

Is there a configuration to remove the empty lines? Especially when not using line numbers, the output can be very confusing (and unnecessarily verbose). There is no way to tell if the blank line is actually part of the source code, or inserted later on for formatting only purposes.

I tried using `file-decoration-style = omit` and `hunk-header-decoration-style = omit` but that also won't remove it.

LunarLambda:
Takes any combination of `file`, `hunk-header`, and `diff` to insert a blank line after the filename line, hunk header line, or after each diff hunk respectively.

Handles edge cases like omitting both file and hunk-header but requesting newlines after diffs.

Ensures output never starts with a blank line (unless git somehow causes it to).

Commit boundaries are not handled due to the fact that git will insert a blank line between commits by itself unless --oneline is requested.

Closes #1944, #1517

Note: There are ~70 test failures I have not yet looked at. This PR took ~6 hours of non-stop work. If the maintainers could be so kind, please go through the failing tests and let me know if any of them are legitimate regressions or if they're harmless (e.g., tests expecting extraneous newlines that I have removed). Thank you ^^

From what I can tell, `git log -p`, `git show -p`, `git diff`, and `git add -p` / `git restore -p` all work. So I *believe* there are no major functionality regressions.


dandavison:
Hi @LunarLambda, thanks for working on this. I'm aware that there have been several requests for control over blank lines in output. Can we make your new option opt-in (i.e. non-default) so that no tests are affected?

LunarLambda:
I'm afraid the code touches too many spots to really make it entirely opt-in. I tried to make the default value be as close to the original behaviour as possible but it does change some things like not having the extra blank line at the start of output.

Especially because I had to move some code to make the logic look reasonable, e.g. printing the newlines after elements not before them (which is incidentally the only thing that allowed file diffs to have a separating blank line between them)

With some guidance I could try to fix the tests but if I try to make the changes completely opt-in the code quality will be much worse I think.

Though that said I also took quite some time to understand the logical structure of the state machine and when and how things are printed so maybe you also have a better idea for how to approach this functionality than I

PR #2056, opened 2025-11-15
fix(git_config): Delta now respects env vars (fixes #1971)

### LINKED ISSUE #1971: 🐛 $XDG_CONFIG_HOME/git/config is not checked
Hiya 👋

I was trying out delta today and had a confusing debug session as I attempted to figure out why none of the `[delta]` settings in my git config seemed to be working. Lo and behold, the issue was that delta only reads from `~/.gitconfig`, whereas I use `$XDG_CONFIG_HOME/git/config`:

```
$ grep -A1 -F '[delta]' $XDG_CONFIG_HOME/git/config
[delta]
        line-numbers = true

$ git config list | grep delta.line-numbers
delta.line-numbers=true

$ delta --show-config | grep line-numbers
    line-numbers                  = false
```

From reading more through the docs I understand there are several ways to overcome this limitation (i.e. passing `--config /path/to/file` to delta). However, given that "you can configure it the same way you configure git" is part of the elevator pitch for delta, I think supporting `$XDG_CONFIG_HOME/git/config` would be worthwhile, as this is [officially supported by git itself](https://git-scm.com/docs/git-config#Documentation/git-config.txt-XDGCONFIGHOMEgitconfig).

That's all. Thanks for creating cool things in your spare time and giving them away for free :)

Daniel-De-Dev:
This change replaces [`git2::Repository::discover()`](https://docs.rs/git2/latest/git2/struct.Repository.html#method.discover) with [`git2::Repository::open_from_env()`](https://docs.rs/git2/latest/git2/struct.Repository.html#method.open_from_env), which _seems_ to correctly respect the `GIT_CONFIG_SYSTEM` and `GIT_CONFIG_GLOBAL` environment variables. (and by extenstion OP's problem? _havent checked_)

I haven’t done extensive testing, but this change resolves the issue I was seeing. As far as I can tell, the core behavior is preserved, but I trust that a reviewer with more experience would be able to confirm.

### DIFF
--- a/src/git_config/mod.rs
+++ b/src/git_config/mod.rs
@@ -42,10 +42,8 @@ impl GitConfig {
     pub fn try_create(env: &DeltaEnv) -> Option<Self> {
         use crate::fatal;
 
-        let repo = match &env.current_dir {
-            Some(dir) => git2::Repository::discover(dir).ok(),
-            _ => None,
-        };
+        let repo = git2::Repository::open_from_env().ok();
+
         let config = match &repo {
             Some(repo) => repo.config().ok(),
             None => git2::Config::open_default().ok(),


Daniel-De-Dev:
I went and attempted in testing OP's original problem and it seems to work on the unpatched version.
I only have `$XDG_CONFIG_DIRS` variable defined on my system, but options added to `~/.config/git/config` were picked up by delta. So the issue might've been resolved earlier or the variable they were specifying `$XDG_CONFIG_HOME` was the problem? Or git2-rs checks `$HOME/.config/git/config` by default? I'm honestly not certain.

Issue #1965, opened 2025-02-26
🚀 Render line numbers as 0 when there is no line number


zx8:
Using v0.18.2

My current setup is:

```
--hyperlinks-file-link-format="vscode://file{path}:{line}"
```

However, when clicking on file links at the top of the diff output (rather than than hunk links), there is no line number, so the link is rendered as:

```
vscode://file/Users/zx8/example/foo.tf:
```

And when I click on it, I get a `No file or directory` error, because the trailing `:` is treated as part of the path.

If this were rendered as `0` when there is no line number, I think the links would work as expected.

timvermeulen:
For me this isn't causing any issues with VSCode (which seems to strip the trailing `:`?) but it is in Zed, which creates a new empty file where the name includes the trailing `:`. Ideally the format to be used when there's no line number could be customized with a separate flag, but defaulting to 0 would do the job as well 🙂

dandavison:
> However, when clicking on file links at the top of the diff output (rather than than hunk links), there is no line number

Thanks for opening this @zx8. Can you (or @schpet?) tell me how to repro this myself? Something like this maybe?

```
git diff | delta --no-gitconfig --hyperlinks --hyperlinks-file-link-format="vscode://file{path}:{line}"
```


But how do I cause it to show a link with no line number?

schpet:
@dandavison i was able to repro this with that command. screenshot from ghostty on mac os below highlighting the trailing colon as a hover over the link, while holding cmd down. 

<img width="2560" height="1540" alt="Image" src="https://github.com/user-attachments/assets/10030d5b-1fbf-4ce7-89c5-d9892a82b783" />

appreciate you checking this out, let me know if there's any more info i can provide.

dandavison:
Great, thanks.

FWIW, I hadn't noticed as I had been displaying the file path in the hunk header via
```gitconfig
file-style = omit
hunk-header-style = file line-number syntax
```
and there it always has a line number.

PR #2060, opened 2025-11-20
Support zed:// hyperlinks with a new colon-omitting {:line} placeholder

schpet:
this adds a new hyperlink placeholder: `{:line}`

similar to `{line}` but includes a colon prefix when a line is provided, otherwise prints nothing. intended as a fix for https://github.com/dandavison/delta/issues/1965 – a problem i ran into with zed.

i'm open to other solutions, such as defaulting line to 1, like what ripgrep does[^1] – but i feel like it's undesirable to take the user back to line 1 if they have the file open and a cursor on a different line. happy to go with line 1 though if that's preferred (that change now up at #2061)

[^1]: https://manpages.debian.org/testing/ripgrep/rg.1.en.html#:~:text=number%20was%20given)%2C-,then%20it%20is%20automatically%20replaced%20with%20the%20value%201.,-%7Bcolumn%7D


schpet:
closing in favour of https://github.com/dandavison/delta/pull/2061

PR #2063, opened 2025-11-23
Update dirs dependency to version 6

musicinmybrain:
Update the `dirs` dependency to the latest version, 6.0.0.

The SemVer-breaking change in `dirs` 6 is just that `dirs-sys` is updated to 0.5, which, in turn, is just due to [updating some platform-specific dependencies](https://github.com/dirs-dev/dirs-sys-rs/compare/f369f0904bec9833572f24c988c7e48454173983...8bcd4aa2c35990d57a2cff2953793525fc42709c). There are therefore no API changes we need to worry about here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -153,7 +153,7 @@ dependencies = [
  "serde_yaml",
  "shell-words",
  "syntect",
- "thiserror",
+ "thiserror 1.0.56",
  "unicode-width",
  "wild",
 ]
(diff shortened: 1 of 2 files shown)


dandavison:
Thanks. Would you mind saying what motivated you to update this? We were using dependabot for a while with this project but are not currently.

musicinmybrain:
A reasonable question! I’m working downstream in Fedora to update the [`rust-dirs`](https://src.fedoraproject.org/rpms/rust-dirs) and [`rust-directories`](https://src.fedoraproject.org/rpms/rust-directories) packages to 6.0.0 and [`rust-dirs-sys`](https://src.fedoraproject.org/rpms/rust-dirs-sys) to 0.5.0 in Fedora without introducing `rust-dirs5`, `rust-directories5`, and `rust-dirs-sys0.4` compat packages, so I’m working through their reverse dependencies, e.g. [`rust-git-delta`](https://src.fedoraproject.org/rpms/rust-git-delta/pull-request/11), and patching them to allow the current versions of these dependencies. It’s our policy to [offer patches upstream](https://docs.fedoraproject.org/en-US/packaging-guidelines/PatchUpstreamStatus/), hence this PR.

I’m not aware of any reason that it would be urgent to actually merge and release this. There aren’t any documented major bugfixes, and we can carry the downstream patch as long as necessary, especially since it’s just a version bound adjustment with no source-code changes.

dandavison:
Right, thanks a lot for explaining and for the work you're doing. Merged; let us know at any point if we are being slow on releases in  a way that is blocking something.

PR #2067, opened 2025-12-03
Switch from ansi-term to nu-ansi-term

jbicha:
ansi-term appears to be unmaintained. Let's switch to the popular maintained fork, nu-ansi-term.

It explicitly adds the "bright" colors and a "default" color. I didn't know how to handle the default color at all and I took a guess at the bright colors. The bright colors are named as Light variants of the existing colors except that DarkGrey is the bright version of black.

My goal was to help one of the last things packaged in Debian using ansi-term to use nu-ansi-term instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,15 +42,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
- "winapi",
-]
-
-[[package]]
 name = "anstream"
 version = "0.6.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


jbicha:
I think some of the line breaks that `cargo fmt` wants aren't as useful for readability. Maybe we need to tweak the [configuration](https://rust-lang.github.io/rustfmt/) or maybe you're happy with what it already recommends.

Feel free to fix this up however you want.

th1000s:
Thank you.

> I didn't know how to handle the default color at all and I took a guess at the bright colors. 

Where did you have to handle those extra `nu_ansi_term` features? Most of the commit looks like straightforward search-and-replace.

Also, I'm thinking of internally re-exporting `nu_ansi_term` as `crate::ansi::term` in `src/ansi/mod.rs` to make the name a bit more brief, if you're ok with force-pushing that in your name. Maybe that also fixes the format issue - which file/line do mean?

jbicha:
I force-pushed an alternate way of keeping the shorter ansi_term name.

ofek:
Can this be rebased? We would like to use Delta in our developer environments but need to satisfy certain supply-chain checks for unmaintained (transitive) dependencies.

PR #2071, opened 2025-12-09
Set computed.color_mode before line numbers feature

th1000s:
This feature inspects on opt.computed.color_mode, which is set by this function. Otherwise --light or --dark have no effect.


th1000s:
And that's it for x86 macOS CI:

"The macOS-13 based runner images are now retired. For more details, see https://github.com/actions/runner-images/issues/13046."

PR #2084, opened 2026-01-04
feat: Adds min-width parameter to enable toggling of side-by-side based on terminal width

### LINKED ISSUE #2083: 🚀 feat: side-by-side support toggle based on terminal width
I've been working around a script that allows me to use delta on `side-by-side: true` only when a new the column width of the terminal is large enough to support it, which emulates a similar experience to github PR review. 

An example script is as follows:

```bash
#!/usr/bin/env bash
set -euo pipefail

# Allow overrides
threshold="${DELTA_SBS_MIN_COLS:-140}"     # change default width here
force="${DELTA_FORCE_SBS:-}"               # set to 1 or 0 to force

# Determine terminal columns
cols="${COLUMNS:-}"
if [[ -z "${cols}" ]]; then
  cols="$(tput cols 2>/dev/null || echo 80)"
fi

# If we're not writing to a TTY (e.g., redirected), don't page and don't sbs
if ! [[ -t 1 ]]; then
  exec delta --paging=never
fi

# Decide side-by-side
use_sbs=0
if [[ "${force}" == "1" ]]; then
  use_sbs=1
elif [[ "${force}" == "0" ]]; then
  use_sbs=0
elif (( cols >= threshold )); then
  use_sbs=1
fi

if (( use_sbs )); then
  exec delta --side-by-side --paging=auto
else
  exec delta --paging=auto
fi
```

It would be useful to have this as a native capability such as `delta --side-by-side-max-width` and configurable with env vars as required.

axsaucedo:
# feat: Adds min-width parameter to enable toggling of side-by-side based on terminal width

This PR includes:

* Extends `--side-by-side` to accept an optional integer argument specifying the minimum
terminal width
* When the terminal is narrower than the specified width, delta falls back to unified diff
* Adding a few tests that validate the min_size applies accordingly (incl git config)
* Extended terminal width param to support TERMINAL_WIDTH_IN_TESTS

Values overview

* `--side-by-side` or `= true` -> Always enabled (backward compatible) 
* `--side-by-side=N` (iff N > 0) -> Enabled only when terminal width >= N columns
* `--side-by-side=0` -> Disabled
* `= false` / `= no` / `= off` -> Disabled
* (absent) -> Disabled (default)

Usage tested in CLI:

```
git --no-pager diff | /delta/target/debug/delta --side-by-side=140
```

Usage tested in git config:

```toml
[core]
    pager = /delta/target/debug/delta

[delta]
    side-by-side = 140
```

As well as:

```toml
[core]
    pager = /delta/target/debug/delta --side-by-side = 140
```

Functionality is as follows with term_width < min_width:

image

And with min_width < term_width:

image

Fixes #2083

### DIFF
--- a/src/options/set.rs
+++ b/src/options/set.rs
@@ -214,7 +214,6 @@ pub fn set_options(
             show_colors,
             show_themes,
             side_by_side,
-            side_by_side_min_width,
             wrap_max_lines,
             wrap_right_prefix_symbol,
             wrap_right_percent,
@@ -243,18 +242,16 @@ pub fn set_options(
         cli::InspectRawLines::from_str(&opt.inspect_raw_lines).unwrap();
     opt.computed.paging_mode = parse_paging_mode(&opt.paging_mode);
 
-    // Apply side-by-side-min-width: disable side-by-side if terminal is too narrow.
-    if opt.side_by_side_min_width > 0
-        && opt.computed.available_terminal_width < opt.side_by_side_min_width
-    {
-        opt.side_by_side = false;
-    }
+    // Resolve side-by-side: parse the optional min-width value and apply terminal width check.
+    let (sbs_enabled, sbs_min_width) = parse_side_by_side(&opt.side_by_side);
+    opt.computed.side_by_side = sbs_enabled
+        && (sbs_min_width == 0 || opt.computed.available_terminal_width >= sbs_min_width);
 
     // --color-only is used for interactive.diffFilter (git add -p). side-by-side, and
     // **-decoration-style cannot be used there (does not emit lines in 1-1 correspondence with raw git output).
     // See #274.
     if opt.color_only {
-        opt.side_by_side = false;
+        opt.computed.side_by_side = false;
         opt.file_decoration_style = "none".to_string();
         opt.commit_decoration_style = "none".to_string();
         opt.hunk_header_decoration_style = "none".to_string();
@@ -402,7 +399,7 @@ fn gather_features(
     if opt.navigate {
         gather_builtin_features_recursively("navigate", &mut features, builtin_features, opt);
     }
-    if opt.side_by_side {
+    if side_by_side_enabled(&opt.side_by_side) {
         gather_builtin_features_recursively("side-by-side", &mut features, builtin_features, opt);
     }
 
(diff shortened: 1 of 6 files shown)


axsaucedo:
I've been using this as my daily driver for a few days now, and working really well. Would be great to explore if this could make it to the release to switch from custom local build - let me know if there are any further changes required, happy to iterate

axsaucedo:
Following up with this one - would it be possible / relevant to get a review? If not no problem (been working with this as a daily driver as using the build has been relatively easy, but would be great if this is contributed upstream)

src/options/set.rs
```diff
@@ -863,4 +895,145 @@ pub mod tests {
...
+    #[test]
+    fn test_side_by_side_min_width_disables_sbs_when_terminal_narrow() {
+        // Use a min-width much larger than any reasonable terminal to guarantee it's disabled.
+        let opt = integration_test_utils::make_options_from_args(&["--side-by-side=99999"]);
+        assert!(!opt.computed.side_by_side);
+    }
+
+    #[test]
+    fn test_side_by_side_always_enables_without_value() {
+        // When no value is given, min-width defaults to 0 (always enabled)
```
pablospe: does `0` mean disabled or enabled? Here it says "always enabled".

pablospe: and in the .md: `Setting `side-by-side = 0` disables side-by-side mode.`. This is what I would except. Probably this comment is incorrect.

pablospe:
Nice! Thanks for this, hopefully it can be merged soon. Btw, please update the PR description too.

src/options/set.rs
axsaucedo: So this is one of the conceptual corner cases that I was talking about. So what this means is that it's when `--side-by-side` is passed without a value, clap internally converts it to `"true"` (via `default_missing_value`), which `parse_side_by_side("true")` maps to `(enabled=true, min_width=0)`. 

So the internal `0` here means "no minimum width threshold". But the user-facing `--side-by-side=0` correctly maps to `(enabled=false, min_width=0)` i.e. disabled, as documented.

Here I can make it such that the explanation is more thorough; for now I will update to: `--side-by-side without a value always sets: enabled=true, min_width=0` - but let me know if you believe that we should still change it.

axsaucedo:
Ok I have now added more details and updated the comment to be more explicit. I have also added the updated PR description with more details, and can also verify that it works in the following:

Tested in cli:

```
git --no-pager diff | /delta/target/debug/delta --side-by-side=140
```

tested in git config:

```toml
[core]
    pager = /delta/target/debug/delta

[delta]
    side-by-side = 140
```

As well as:

```toml
[core]
    pager = /delta/target/debug/delta --side-by-side = 140
```

pablospe (approved):

axsaucedo:
Ran `make format` as it seems lint test failed.

axsaucedo:
@pablospe it should now be ready to go - let me know if anything else required before approving! Thank you

pablospe:
@axsaucedo I can't merge this since I am not a maintainer of this project. I was just trying to help @dandavison with code review.

pablospe (approved):

axsaucedo:
Ah sorry, good to know - thank you for the review!

axsaucedo:
(Sorry seems I forgot to explicitly re-request the review)

pablospe (approved):

cespare:
@axsaucedo thanks for this! I've been using your patch and it's great. I hope it gets merged soon.

axsaucedo:
Great to hear! It should be good to go, @dandavison let me know if there are still any design changes, or further test coverage, happy to follow up accordingly!

PR #2095, opened 2026-02-01
Add --no-* negation flags for boolean CLI options

pablospe:
### Summary                                                                                                                                                                                    
  - Add hidden `--no-*` counterparts for 10 boolean flags: `--no-side-by-side`, `--no-line-numbers`, `--no-navigate`, `--no-hyperlinks`, `--no-color-only`, `--no-diff-highlight`,              
  `--no-diff-so-fancy`, `--no-keep-plus-minus-markers`, `--no-raw`, `--no-relative-paths`                                                                                                       
  - Allows disabling boolean options from the CLI when they are enabled in gitconfig                                                                                                            
  - Uses clap's `overrides_with` for last-flag-wins semantics (e.g. `--side-by-side --no-side-by-side` → disabled)                                                                              

### Motivation                                 
  When boolean flags like `--side-by-side` or `--line-numbers` are set in gitconfig, there is currently no CLI way to disable them. This is needed for tools that invoke delta programmatically and need to toggle features on and off.

**Note**: this is a workaround until https://github.com/clap-rs/clap/issues/815 is resolved, which would allow clap to natively support `--no-*` flag generation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,8 @@ src/junk.rs
 # insta: unreviewed inline data, and separate snapshot files
 *.pending-snap
 **/snapshots/*.snap.new
+
+# task automation
+AGENTS.md
+CLAUDE.md
+.task/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,10 @@
 version = 4
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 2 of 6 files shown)


pablospe:
@dandavison Ready for review. Thanks

dandavison:
Hi @pablospe, how confident are you that the command names here will match exactly what clap introduces, if they do introduce a solution?

pablospe:
That's a fair concern. I'm not fully confident the names would match exactly. However, the upstream issue (https://github.com/clap-rs/clap/issues/815) has been open since 2017 and there doesn't seem to be movement toward a built-in solution anytime soon.

Given how straightforward this implementation is and that it addresses a real usability need (being able to override config file settings from the command line), I think it's worth having now rather than waiting indefinitely. If clap ever does add native support, it should be easy to migrate at that point.

pablospe:
@dandavison. Following up on your question about clap compatibility. I went with the `--no-*` convention which is the most common pattern for flag negation.

Would you prefer a different approach here? For example:
- A different naming convention (e.g. `--disable-*`)
- Restricting this to just a few high-demand flags like `--no-side-by-side`
- Waiting for clap's native solution

Let me know what you'd like and I'll update the PR accordingly.

PR #2117, opened 2026-03-08
Update docs to make 3 basic configs match

injust:
Re-landing https://github.com/dandavison/delta/pull/2013


dandavison:
Thanks @injust!

PR #2118, opened 2026-03-08
Document support for `box ul` style

injust:


injust:
Abandoning due to inactivity

PR #2161, opened 2026-05-17
Add `--wrap-word-lookback` for word-aware line wrapping

igrmk:
Opt-in setting that backs the wrap point up to N columns in search of a more natural break (whitespace if present, else the nearest non-word character) instead of slicing through the middle of a word. Defaults to `0` (disabled), so existing output is unchanged.

## Before (`--wrap-max-lines=3`):

before

## After (`--wrap-max-lines=3 --wrap-word-lookback=30`):

after


igrmk:
I found a bug in this implementation: it only searches within the current line segment instead of the whole line. I'll reopen this if/when I get it fixed

PR #2181, opened 2026-07-02
Support OSC-1717 metadata

### LINKED ISSUE #2180: OSC-1717 support for better integration in lazygit
Hi Dan, I'm the current maintainer of lazygit, and I have come up with a custom OSC protocol that improves the user experience of lazygit when custom pagers are used; see https://github.com/jesseduffield/lazygit/pull/5732 for information about that, including a video that explains what this is about. Your feedback would be very welcome!

In case you are short on time: I want to make it very clear that this doesn't need any work from your side, apart from reviewing the PR that I'm going to post. If you want to give feedback on the approach, that would be very welcome, but if not, that's fine as well.

The main reason why I'm filing this issue (rather than simply open that PR) is that the work on the lazygit side is massive, and I don't want to start it before I have buy-in from the most important pager authors. I _am_ eager to start it though, so if you could give your general ok, that would be great.

stefanhaller:
This adds support for the OSC-1717 spec described in https://github.com/stefanhaller/diff-line-metadata-spec; see also https://github.com/jesseduffield/lazygit/pull/5732 for more context.

The spec is still in draft mode, so we may want to hold off merging this until we got more feedback on it. On the other hand, we want pagers to be ready by the time lazygit releases the feature, that's why I'm opening this for review already.

Closes #2180.

### DIFF
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -149,7 +149,7 @@ impl<'a> StateMachine<'a> {
         }
     }
 
-    fn consume<I>(&mut self, mut lines: ByteLines<I>) -> std::io::Result<()>
+    pub(crate) fn consume<I>(&mut self, mut lines: ByteLines<I>) -> std::io::Result<()>
     where
         I: BufRead,
     {
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -118,6 +118,12 @@ pub fn delta<I>(lines: ByteLines<I>, writer: &mut dyn Write, config: &Config) ->
 where
     I: BufRead,
 {
+    // Announce protocol support as the first output, so a host can probe delta even on
+    // an empty diff (which produces no per-line records). See features::diff_line_metadata.
+    let handshake = features::diff_line_metadata::handshake();
+    if !handshake.is_empty() {
+        writer.write_all(handshake.as_bytes())?;
+    }
     StateMachine::new(writer, config).consume(lines)
 }
 
--- a/src/features/diff_line_metadata.rs
+++ b/src/features/diff_line_metadata.rs
@@ -85,6 +85,20 @@ impl DiffLineMetadata {
         })
     }
 
+    /// A version-1 emitter regardless of environment. Tests cannot go through
+    /// `from_env`: the negotiated version is cached in a process-wide OnceLock,
+    /// so an env var set by one test would leak into all others.
+    #[cfg(test)]
+    pub fn v1_for_tests() -> Self {
+        Self {
+            version: 1,
+            old_line: 0,
+            new_line: 0,
+            file: String::new(),
+            last_record: None,
+        }
+    }
+
     /// Seed the line-number counters and file path at a hunk header. Mirrors
     /// `LineNumbersData::initialize_hunk`: the first entry is the old-file start,
     /// the last is the new-file start.
@@ -226,13 +240,7 @@ mod tests {
     use crate::delta::{DiffType, State};
 
     fn emitter() -> DiffLineMetadata {
-        let mut md = DiffLineMetadata {
-            version: 1,
-            old_line: 0,
-            new_line: 0,
-            file: String::new(),
-            last_record: None,
-        };
+        let mut md = DiffLineMetadata::v1_for_tests();
         // @@ -1,3 +1,4 @@ : old start 1, new start 1.
         md.initialize_hunk(&[(1, 3), (1, 4)], "f.txt".to_owned());
         md
--- a/src/features/diff_line_metadata.rs
+++ b/src/features/diff_line_metadata.rs
@@ -16,6 +16,7 @@
 // Ghostty, VS Code, ConEmu, urxvt); none of them interpret it, so a terminal that
 // is not a participating host skips it harmlessly.
 
+use std::io::{self, Write};
 use std::sync::OnceLock;
 
 use crate::delta::{DiffType, State};
@@ -142,6 +143,81 @@ impl DiffLineMetadata {
             file = self.file,
         )
     }
+
+    /// The `h` (hunk-header) record for the current hunk. `initialize_hunk` has
+    /// just seeded `new_line` with the hunk's new-file start -- which is the
+    /// hunk's first line (spec §5.2) -- so it is in hand exactly when the hunk
+    /// header is rendered. `old-line` is empty on a header.
+    pub fn osc_for_hunk_header(&self) -> String {
+        self.header_osc('h', Some(self.new_line), &self.file)
+    }
+
+    /// The `f` (file-header) record. It carries no line numbers (spec §5.5):
+    /// delta renders the file header (the boxed file name) as soon as it parses
+    /// the `+++` line, *before* it has seen the first `@@`, so the file's first
+    /// hunk line is not known here. The path is supplied by the caller because
+    /// the emitter only learns it at the first hunk header.
+    pub fn osc_for_file_header(&self, file: &str) -> String {
+        self.header_osc('f', None, file)
+    }
+
+    fn header_osc(&self, type_char: char, new_line: Option<usize>, file: &str) -> String {
+        let new_field = new_line.map_or(String::new(), |n| n.to_string());
+        format!(
+            "{OSC};{version};{type_char};{new_field};;{file}{ST}",
+            version = self.version,
+        )
+    }
+}
+
+/// A `Write` adapter that injects `prefix` at the start of the stream and after
+/// every newline. delta draws a file/hunk header as a *multi-row* decoration --
+/// an underline under the file name, a box around the hunk header -- written
+/// straight to the output by the `draw` functions. To tag the whole block we
+/// wrap that writer so every row the decoration emits is preceded by the
+/// header's record, the same "every output row carries the record" rule that
+/// wrapped content rows follow (spec §6.3/§6.4).
+struct OscLinePrefixer<'a> {
+    inner: &'a mut dyn Write,
+    prefix: &'a str,
+    at_line_start: bool,
+}
+
+impl Write for OscLinePrefixer<'_> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        for chunk in buf.split_inclusive(|&b| b == b'\n') {
+            if self.at_line_start {
+                self.inner.write_all(self.prefix.as_bytes())?;
+            }
+            self.inner.write_all(chunk)?;
+            self.at_line_start = chunk.ends_with(b"\n");
+        }
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.inner.flush()
+    }
+}
+
+/// Run `draw`, prefixing every output row it writes with `osc` (a header
+/// record). With `osc` empty the closure writes to `writer` untouched, so the
+/// output stays byte-for-byte identical to stock delta.
+pub fn write_with_header_osc(
+    writer: &mut dyn Write,
+    osc: &str,
+    draw: impl FnOnce(&mut dyn Write) -> io::Result<()>,
+) -> io::Result<()> {
+    if osc.is_empty() {
+        draw(writer)
+    } else {
+        let mut prefixer = OscLinePrefixer {
+            inner: writer,
+            prefix: osc,
+            at_line_start: true,
+        };
+        draw(&mut prefixer)
+    }
 }
 
 #[cfg(test)]
@@ -167,6 +243,53 @@ mod tests {
         // The handshake carries only the version (no further fields), so a host tells
         // it apart from a per-line record by field count.
         assert_eq!(handshake_for_version(1), "\x1b]1717;1\x1b\\");
+    }
+
+    #[test]
+    fn test_hunk_header_record_carries_the_hunks_first_line() {
+        // `h` carries the hunk's first new-file line (just seeded by
+        // initialize_hunk) and no old-line. Emitting it does not advance the
+        // counters: the hunk's first content line reports the same line.
+        let mut md = emitter();
+        assert_eq!(md.osc_for_hunk_header(), "\x1b]1717;1;h;1;;f.txt\x1b\\");
+        assert_eq!(
+            md.osc_for_line(&State::HunkZero(DiffType::Unified, None)),
+            "\x1b]1717;1;c;1;;f.txt\x1b\\"
+        );
+    }
+
+    #[test]
+    fn test_file_header_record_has_no_line_numbers() {
+        // `f` never carries line numbers (spec §5.5): delta draws the file
+        // header before it has seen the first `@@`. The path comes from the
+        // caller, because the emitter's own state is only seeded at the first
+        // hunk header.
+        let md = emitter();
+        assert_eq!(
+            md.osc_for_file_header("src/foo.rs"),
+            "\x1b]1717;1;f;;;src/foo.rs\x1b\\"
+        );
+    }
+
+    #[test]
+    fn test_header_record_prefixes_every_row_of_a_multi_row_block() {
+        // A file/hunk header is a multi-row decoration (an underlined name, a
+        // box); every row of the block carries the record (spec §6.4). With no
+        // record the drawn bytes pass through untouched.
+        let draw = |w: &mut dyn io::Write| -> io::Result<()> {
+            write!(w, "──┐\n1:│\n──┘\n")
+        };
+
+        let mut tagged = Vec::new();
+        write_with_header_osc(&mut tagged, "[osc]", draw).unwrap();
+        assert_eq!(
+            String::from_utf8(tagged).unwrap(),
+            "[osc]──┐\n[osc]1:│\n[osc]──┘\n"
+        );
+
+        let mut untouched = Vec::new();
+        write_with_header_osc(&mut untouched, "", draw).unwrap();
+        assert_eq!(String::from_utf8(untouched).unwrap(), "──┐\n1:│\n──┘\n");
     }
 
     #[test]
--- a/src/features/diff_line_metadata.rs
+++ b/src/features/diff_line_metadata.rs
@@ -1,0 +1,113 @@
+// Emit per-line diff metadata as an OSC sequence, so that a host application
+// (e.g. lazygit) can map a rendered diff row back to its patch-space identity
+// (file, type, new-file line, old-file line) without re-parsing the rendered
+// output. delta restructures the diff (drops the +/- markers, conveys the side
+// by color, adds gutters), so that identity cannot be recovered from the
+// painted text alone -- the pager, which still has it at render time, states it.
+//
+// This is gated on the OSC1717_METADATA environment variable: the host
+// advertises the protocol versions it understands and delta emits the highest
+// mutually-understood one. When the variable is unset (i.e. delta is not running
+// under such a host) nothing is emitted, so a raw terminal / less / tmux see a
+// normal diff.
+//
+// The OSC number 1717 was chosen after auditing the OSC allocations of the major
+// terminal emulators (xterm, VTE, kitty, foot, WezTerm, iTerm2, Windows Terminal,
+// Ghostty, VS Code, ConEmu, urxvt); none of them interpret it, so a terminal that
+// is not a participating host skips it harmlessly.
+
+use std::sync::OnceLock;
+
+use crate::delta::{DiffType, State};
+
+/// Highest protocol version this build of delta knows how to emit.
+const SUPPORTED_VERSION: u32 = 1;
+
+const OSC: &str = "\x1b]1717";
+const ST: &str = "\x1b\\";
+
+/// The highest version in the host's advertised list (e.g. "V1" or "V1,V2")
+/// that this build also understands, or `None` if the lists are disjoint.
+fn pick_version(advertised: &str) -> Option<u32> {
+    advertised
+        .split(',')
+        .filter_map(|v| v.trim().strip_prefix('V')?.parse::<u32>().ok())
+        .filter(|v| *v <= SUPPORTED_VERSION)
+        .max()
+}
+
+/// The protocol version to emit, negotiated against the host's advertised list
+/// in `OSC1717_METADATA` (e.g. "V1" or "V1,V2"), or `None` when no host is
+/// asking (the variable is unset) so delta stays silent.
+pub fn negotiated_version() -> Option<u32> {
+    static VERSION: OnceLock<Option<u32>> = OnceLock::new();
+    *VERSION.get_or_init(|| pick_version(&std::env::var("OSC1717_METADATA").ok()?))
+}
+
+/// Tracks the current old-/new-file line numbers within a hunk and formats the
+/// per-line OSC sequence. One instance is held by the `Painter` (when emission
+/// is negotiated) and re-seeded at every hunk header.
+pub struct DiffLineMetadata {
+    version: u32,
+    old_line: usize,
+    new_line: usize,
+    file: String,
+}
+
+impl DiffLineMetadata {
+    /// Returns an emitter iff the host negotiated a protocol version.
+    pub fn from_env() -> Option<Self> {
+        negotiated_version().map(|version| Self {
+            version,
+            old_line: 0,
+            new_line: 0,
+            file: String::new(),
+        })
+    }
+
+    /// Seed the line-number counters and file path at a hunk header. Mirrors
+    /// `LineNumbersData::initialize_hunk`: the first entry is the old-file start,
+    /// the last is the new-file start.
+    pub fn initialize_hunk(&mut self, line_numbers: &[(usize, usize)], file: String) {
+        self.old_line = line_numbers[0].0;
+        self.new_line = line_numbers[line_numbers.len() - 1].0;
+        self.file = file;
+    }
+
+    /// Advance the counters for `state` and return the OSC sequence to prepend
+    /// to that content line, or an empty string for states that carry no
+    /// metadata (headers, wrapped continuations, and -- in this prototype --
+    /// combined/merge diffs). The counter arithmetic mirrors
+    /// `line_numbers::linenumbers_and_styles`.
+    pub fn osc_for_line(&mut self, state: &State) -> String {
+        use State::*;
+        let (type_char, new_line, old_line) = match state {
+            HunkZero(DiffType::Unified, _) => {
+                let record = ('c', self.new_line, None);
+                self.old_line += 1;
+                self.new_line += 1;
+                record
+            }
+            HunkPlus(DiffType::Unified, _) => {
+                let record = ('a', self.new_line, None);
+                self.new_line += 1;
+                record
+            }
+            HunkMinus(DiffType::Unified, _) => {
+                // A deletion sits at the new-file position the new-file counter
+                // currently holds (it has not advanced past the preceding
+                // context/added lines); it carries both numbers.
+                let record = ('d', self.new_line, Some(self.old_line));
+                self.old_line += 1;
+                record
+            }
+            _ => return String::new(),
+        };
+        let old_field = old_line.map_or(String::new(), |n| n.to_string());
+        format!(
+            "{OSC};{version};{type_char};{new_line};{old_field};{file}{ST}",
+            version = self.version,
+            file = self.file,
+        )
+    }
+}
--- a/src/features/diff_line_metadata.rs
+++ b/src/features/diff_line_metadata.rs
@@ -1,0 +1,113 @@
+// Emit per-line diff metadata as an OSC sequence, so that a host application
+// (e.g. lazygit) can map a rendered diff row back to its patch-space identity
+// (file, type, new-file line, old-file line) without re-parsing the rendered
+// output. delta restructures the diff (drops the +/- markers, conveys the side
+// by color, adds gutters), so that identity cannot be recovered from the
+// painted text alone -- the pager, which still has it at render time, states it.
+//
+// This is gated on the OSC1717 environment variable: the host advertises the
+// protocol versions it understands and delta emits the highest mutually-
+// understood one. When the variable is unset (i.e. delta is not running under
+// such a host) nothing is emitted, so a raw terminal / less / tmux see a normal
+// diff.
+//
+// The OSC number 1717 was chosen after auditing the OSC allocations of the major
+// terminal emulators (xterm, VTE, kitty, foot, WezTerm, iTerm2, Windows Terminal,
+// Ghostty, VS Code, ConEmu, urxvt); none of them interpret it, so a terminal that
+// is not a participating host skips it harmlessly.
+
+use std::sync::OnceLock;
+
+use crate::delta::{DiffType, State};
+
+/// Highest protocol version this build of delta knows how to emit.
+const SUPPORTED_VERSION: u32 = 1;
+
+const OSC: &str = "\x1b]1717";
+const ST: &str = "\x1b\\";
+
+/// The highest version in the host's advertised list (e.g. "V1" or "V1,V2")
+/// that this build also understands, or `None` if the lists are disjoint.
+fn pick_version(advertised: &str) -> Option<u32> {
+    advertised
+        .split(',')
+        .filter_map(|v| v.trim().strip_prefix('V')?.parse::<u32>().ok())
+        .filter(|v| *v <= SUPPORTED_VERSION)
+        .max()
+}
+
+/// The protocol version to emit, negotiated against the host's advertised list
+/// in `OSC1717` (e.g. "V1" or "V1,V2"), or `None` when no host is asking (the
+/// variable is unset) so delta stays silent.
+pub fn negotiated_version() -> Option<u32> {
+    static VERSION: OnceLock<Option<u32>> = OnceLock::new();
+    *VERSION.get_or_init(|| pick_version(&std::env::var("OSC1717").ok()?))
+}
+
+/// Tracks the current old-/new-file line numbers within a hunk and formats the
+/// per-line OSC sequence. One instance is held by the `Painter` (when emission
+/// is negotiated) and re-seeded at every hunk header.
+pub struct DiffLineMetadata {
+    version: u32,
+    old_line: usize,
+    new_line: usize,
+    file: String,
+}
+
+impl DiffLineMetadata {
+    /// Returns an emitter iff the host negotiated a protocol version.
+    pub fn from_env() -> Option<Self> {
+        negotiated_version().map(|version| Self {
+            version,
+            old_line: 0,
+            new_line: 0,
+            file: String::new(),
+        })
+    }
+
+    /// Seed the line-number counters and file path at a hunk header. Mirrors
+    /// `LineNumbersData::initialize_hunk`: the first entry is the old-file start,
+    /// the last is the new-file start.
+    pub fn initialize_hunk(&mut self, line_numbers: &[(usize, usize)], file: String) {
+        self.old_line = line_numbers[0].0;
+        self.new_line = line_numbers[line_numbers.len() - 1].0;
+        self.file = file;
+    }
+
+    /// Advance the counters for `state` and return the OSC sequence to prepend
+    /// to that content line, or an empty string for states that carry no
+    /// metadata (headers, wrapped continuations, and -- in this prototype --
+    /// combined/merge diffs). The counter arithmetic mirrors
+    /// `line_numbers::linenumbers_and_styles`.
+    pub fn osc_for_line(&mut self, state: &State) -> String {
+        use State::*;
+        let (type_char, new_line, old_line) = match state {
+            HunkZero(DiffType::Unified, _) => {
+                let record = ('c', self.new_line, None);
+                self.old_line += 1;
+                self.new_line += 1;
+                record
+            }
+            HunkPlus(DiffType::Unified, _) => {
+                let record = ('a', self.new_line, None);
+                self.new_line += 1;
+                record
+            }
+            HunkMinus(DiffType::Unified, _) => {
+                // A deletion sits at the new-file position the new-file counter
+                // currently holds (it has not advanced past the preceding
+                // context/added lines); it carries both numbers.
+                let record = ('d', self.new_line, Some(self.old_line));
+                self.old_line += 1;
+                record
+            }
+            _ => return String::new(),
+        };
+        let old_field = old_line.map_or(String::new(), |n| n.to_string());
+        format!(
+            "{OSC};{version};{type_char};{new_line};{old_field};{file}{ST}",
+            version = self.version,
+            file = self.file,
+        )
+    }
+}
--- a/src/features/diff_line_metadata.rs
+++ b/src/features/diff_line_metadata.rs
@@ -52,6 +52,11 @@ pub struct DiffLineMetadata {
     old_line: usize,
     new_line: usize,
     file: String,
+    /// The record (`type_char`, `new_line`, `old_line`) most recently emitted
+    /// for a primary content line, so a wrapped continuation row can re-emit it
+    /// without advancing the counters. Set by every primary line; read by the
+    /// wrapped row(s) that immediately follow it.
+    last_record: Option<(char, usize, Option<usize>)>,
 }
 
 impl DiffLineMetadata {
@@ -62,6 +67,7 @@ impl DiffLineMetadata {
             old_line: 0,
             new_line: 0,
             file: String::new(),
+            last_record: None,
         })
     }
 
@@ -76,12 +82,19 @@ impl DiffLineMetadata {
 
     /// Advance the counters for `state` and return the OSC sequence to prepend
     /// to that content line, or an empty string for states that carry no
-    /// metadata (headers, wrapped continuations, and -- in this prototype --
-    /// combined/merge diffs). The counter arithmetic mirrors
-    /// `line_numbers::linenumbers_and_styles`.
+    /// metadata (headers and -- in this prototype -- combined/merge diffs). The
+    /// counter arithmetic mirrors `line_numbers::linenumbers_and_styles`.
+    ///
+    /// A wrapped continuation row (`Hunk*Wrapped`, produced when side-by-side
+    /// wraps a long line into several output rows) re-emits the record of the
+    /// primary line it continues, *without* advancing the counters. delta emits
+    /// each wrapped row as a distinct output line, so a host sees them as
+    /// distinct lines and needs the identity on each -- otherwise acting on a
+    /// continuation row, or treating the wrapped line as one block when
+    /// navigating, breaks.
     pub fn osc_for_line(&mut self, state: &State) -> String {
         use State::*;
-        let (type_char, new_line, old_line) = match state {
+        let record = match state {
             HunkZero(DiffType::Unified, _) => {
                 let record = ('c', self.new_line, None);
                 self.old_line += 1;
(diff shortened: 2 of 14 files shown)


dandavison:
Kimi k3 believes it has found scenarios where this is behaving incorrectly. See the shell script below. I haven't investigated, but maybe this is helpful output to give Claude just as a challenge regarding this PR.

## Summary

| Claim | Status |
|---|---|
| Bug 1a: `f` record on `---` rows in `--color-only` mode (plain `diff -u`) has empty/previous-file path | REPRODUCED |
| Bug 1b: `f` record wrong at `handle_pending_line_with_diff_name` (mode-change-only file) | NOT reproduced (correct: prefilled from diff line) |
| Bug 1c: `f` record on `Only in` rows names previous file | REPRODUCED |
| Bug 1d: `f` record on `rename from` row in `--color-only` mode is empty | REPRODUCED |
| Bug 2: side-by-side wrapped continuation rows lack the OSC record | NOT reproduced (all rows carry the record) |



<details>
  <summary>
kimi's `repro.sh`
  </summary>

```bash
#!/usr/bin/env bash
# Repro script for two suspected correctness bugs in delta's OSC 1717
# "diff line metadata" feature (branch osc-1717-metadata).
#
# Run it with:
#
#     ./repro.sh
#
# It builds delta if needed, runs the scenarios below, and prints a
# self-contained markdown report to stdout. Idempotent: all intermediate
# files are regenerated under /tmp/osc1717-repro/work on every run.
set -euo pipefail

REPO="${DELTA_REPO:-.}"
BIN="$REPO/target/debug/delta"
WORK=/tmp/osc1717-repro/work
mkdir -p "$WORK"

if [ ! -x "$BIN" ]; then
    echo "building delta..." >&2
    cargo build --manifest-path "$REPO/Cargo.toml" >&2
fi

# ---------------------------------------------------------------------------
# Inputs
# ---------------------------------------------------------------------------

cat > "$WORK/git-two-files.diff" <<'EOF'
diff --git a/one.txt b/one.txt
index 257cc56..5716ca5 100644
--- a/one.txt
+++ b/one.txt
@@ -1 +1 @@
-foo
+bar
diff --git a/two.txt b/two.txt
index 257cc56..5716ca5 100644
--- a/two.txt
+++ b/two.txt
@@ -1 +1 @@
-foo
+bar
EOF

cat > "$WORK/mode-change.diff" <<'EOF'
diff --git a/a.txt b/a.txt
index 44371ed..e69de29 100644
--- a/a.txt
+++ b/a.txt
@@ -1 +0,0 @@
-A-content
diff --git a/b.txt b/b.txt
old mode 100644
new mode 100755
EOF

cat > "$WORK/diff-u-two-files.diff" <<'EOF'
--- old1.txt	2024-01-01 00:00:00 +0000
+++ new1.txt	2024-01-01 00:00:00 +0000
@@ -1 +1 @@
-foo
+bar
--- old2.txt	2024-01-01 00:00:00 +0000
+++ new2.txt	2024-01-01 00:00:00 +0000
@@ -1 +1 @@
-foo
+bar
EOF

cat > "$WORK/only-in.diff" <<'EOF'
--- dir1/f.txt	2024-01-01 00:00:00 +0000
+++ dir2/f.txt	2024-01-01 00:00:00 +0000
@@ -1 +1 @@
-foo
+bar
Only in dir1: g.txt
EOF

cat > "$WORK/rename.diff" <<'EOF'
diff --git a/old.txt b/new.txt
similarity index 90%
rename from old.txt
rename to new.txt
EOF

LONG_X=$(printf 'x%.0s' $(seq 1 200))
LONG_Y=$(printf 'y%.0s' $(seq 1 200))
LONG_Z=$(printf 'z%.0s' $(seq 1 200))

cat > "$WORK/wrap-plus.diff" <<EOF
diff --git a/f.txt b/f.txt
index 257cc56..5716ca5 100644
--- a/f.txt
+++ b/f.txt
@@ -1,2 +1,2 @@
 ctx
-old
+${LONG_X}
EOF

cat > "$WORK/wrap-minus.diff" <<EOF
diff --git a/f.txt b/f.txt
index 257cc56..5716ca5 100644
--- a/f.txt
+++ b/f.txt
@@ -1,2 +1,2 @@
 ctx
-${LONG_Z}
+new
EOF

cat > "$WORK/wrap-ctx.diff" <<EOF
diff --git a/f.txt b/f.txt
index 257cc56..5716ca5 100644
--- a/f.txt
+++ b/f.txt
@@ -1,3 +1,3 @@
 ${LONG_Y}
-old
+new
 tail
EOF

# ---------------------------------------------------------------------------
# Helpers
# ---------------------------------------------------------------------------

# The environment this session runs in exports DELTA_FEATURES=+side-by-side
# and the user gitconfig enables hyperlinks etc; sanitize both so the runs
# are controlled and reproducible.
DELTA=(env -u DELTA_FEATURES "$BIN" --no-gitconfig)

# run <name> <command...>: capture stdout, stderr, exit code.
run() {
    local name=$1; shift
    local rc=0
    "$@" > "$WORK/$name.stdout" 2> "$WORK/$name.stderr" || rc=$?
    echo "$rc" > "$WORK/$name.rc"
}

# visible <file>: strip ANSI SGR/color sequences, then render control chars
# with cat -v so the OSC records show as ^[]1717;...^[\ .
visible() {
    perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g' "$1" | cat -v
}

# rows_missing_record <file> <content-regex> <record-regex>:
# count output rows matching <content-regex> that do NOT also match
# <record-regex> (i.e. visual rows of a wrapped line lacking the OSC record).
rows_missing_record() {
    perl -ne 'print if /'"$2"'/ && !/'"$3"'/' "$1" | wc -l | tr -d ' '
}

# ---------------------------------------------------------------------------
# Scenarios
# ---------------------------------------------------------------------------

# Sanity: with OSC1717 unset, output is byte-identical to stock delta; with
# OSC1717=V1, stripping the OSC records restores the stock bytes.
run b0_stock   "${DELTA[@]}" < "$WORK/git-two-files.diff"
run b0_tagged  env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig \
    < "$WORK/git-two-files.diff"
perl -pe 's/\x1b\]1717[^\x1b]*\x1b\\//g' "$WORK/b0_tagged.stdout" \
    > "$WORK/b0_stripped.out"
if cmp -s "$WORK/b0_stock.stdout" "$WORK/b0_stripped.out"; then
    B0_VERDICT="PASS: with OSC1717=V1, stripping the emitted OSC records yields bytes identical to stock delta (OSC1717 unset)."
else
    B0_VERDICT="FAIL: output differs from stock delta even after stripping OSC records."
fi

# Bug 1a: --color-only, plain `diff -u` input (no `diff --git` lines), two
# files. The f record is emitted when the `---` line is handled, before
# plus_file for the current file has been parsed.
run b1a env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig --color-only \
    < "$WORK/diff-u-two-files.diff"
visible "$WORK/b1a.stdout" > "$WORK/b1a.visible"
B1A_EMPTY=$(grep -c '1717;1;f;;;\^\[' "$WORK/b1a.visible" || true)
B1A_STALE=$(grep -c '1717;1;f;;;new1.txt\^\[\\--- old2.txt' "$WORK/b1a.visible" || true)

# Bug 1b: mode-change-only file rendered via handle_pending_line_with_diff_name
# (the call site named in the suspicion). Default (non-color-only) mode.
run b1b env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig \
    < "$WORK/mode-change.diff"
visible "$WORK/b1b.stdout" > "$WORK/b1b.visible"
B1B_OK=$(grep -c '1717;1;f;;;b.txt\^\[\\b.txt (mode +x)' "$WORK/b1b.visible" || true)

# Bug 1c: `Only in dir: file` line rendered via handle_additional_cases.
# Default (non-color-only) mode.
run b1c env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig \
    < "$WORK/only-in.diff"
visible "$WORK/b1c.stdout" > "$WORK/b1c.visible"
B1C_STALE=$(grep -c '1717;1;f;;;dir2/f.txt\^\[\\Only in dir1: g.txt' "$WORK/b1c.visible" || true)

# Bug 1d: --color-only git rename: `rename from` row is emitted before
# plus_file is set (and the a/old b/new diff line prefill yields "").
run b1d env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig --color-only \
    < "$WORK/rename.diff"
visible "$WORK/b1d.stdout" > "$WORK/b1d.visible"
B1D_EMPTY=$(grep -c '1717;1;f;;;\^\[\\rename from old.txt' "$WORK/b1d.visible" || true)

# Bug 2: side-by-side wrapping. A 200-char line at --width 40 wraps into
# several visual rows; each row must carry the line's OSC record.
run b2a env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig \
    --side-by-side --width 40 < "$WORK/wrap-plus.diff"
visible "$WORK/b2a.stdout" > "$WORK/b2a.visible"
B2A_ROWS=$(grep -c 'xxxxx' "$WORK/b2a.visible" || true)
B2A_MISSING=$(rows_missing_record "$WORK/b2a.visible" 'xxxxx' '1717;1;a;2;;f\.txt')

run b2b env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig \
    --side-by-side --width 40 < "$WORK/wrap-ctx.diff"
visible "$WORK/b2b.stdout" > "$WORK/b2b.visible"
B2B_ROWS=$(grep -c 'yyyyy' "$WORK/b2b.visible" || true)
B2B_MISSING=$(rows_missing_record "$WORK/b2b.visible" 'yyyyy' '1717;1;c;1;;f\.txt')

run b2c env -u DELTA_FEATURES OSC1717=V1 "$BIN" --no-gitconfig \
    --side-by-side --width 40 < "$WORK/wrap-minus.diff"
visible "$WORK/b2c.stdout" > "$WORK/b2c.visible"
B2C_ROWS=$(grep -c 'zzzzz' "$WORK/b2c.visible" || true)
B2C_MISSING=$(rows_missing_record "$WORK/b2c.visible" 'zzzzz' '1717;1;d;2;2;f\.txt')

# ---------------------------------------------------------------------------
# Report
# ---------------------------------------------------------------------------

show() { # show <name> <max-lines>
    head -n "$2" "$WORK/$1.visible" | cut -c1-100
}

cmd_line() { echo "$*"; }

cat <<EOF
# OSC 1717 diff-line-metadata: repro report for two suspected bugs

Repo: \`$REPO\` (binary \`target/debug/delta\`).
All runs use \`--no-gitconfig\` and unset \`DELTA_FEATURES\` (the session
environment exports \`DELTA_FEATURES=+side-by-side\`). Output below is rendered
with ANSI color sequences stripped and control chars shown via \`cat -v\`, so an
OSC 1717 record appears as \`^[]1717;1;<type>;<new>;<old>;<file>^[\\\\\`.

## Suspected bug 1: \`f\` (file-header) record names the wrong file

\`diff_header_osc()\` reads \`self.plus_file\`/\`self.minus_file\`, which are only
populated when the \`---\`/\`+++\` (or \`rename from\`/\`to\`) lines are handled.
At call sites that emit the header row *before* those lines are processed, the
record can carry an empty path or the previous file's path while the visible row
names the correct file.

### Scenario B1a: \`--color-only\`, plain \`diff -u\` input, two files

Input (\`diff-u-two-files.diff\`): two \`---\`/\`+++\`/\`@@\` file sections
(\`old1.txt\`→\`new1.txt\`, \`old2.txt\`→\`new2.txt\`), no \`diff --git\` lines.

Command:

\`\`\`bash
OSC1717=V1 delta --no-gitconfig --color-only < diff-u-two-files.diff
\`\`\`

(exit code $(cat "$WORK/b1a.rc"); full raw stdout: \`$WORK/b1a.stdout\`)

Output:

\`\`\`
$(show b1a 12)
\`\`\`

Expected: the \`f\` record on each row names the file that row is about.
Actual:

- The \`--- old1.txt\` row is tagged \`^[]1717;1;f;;;^[\\\\\` — **empty path**
  (matches: $B1A_EMPTY).
- The \`--- old2.txt\` row is tagged \`^[]1717;1;f;;;new1.txt^[\\\\\` — the
  **previous file's** path (matches: $B1A_STALE), while the row itself says
  \`old2.txt\` and the following \`h\` record correctly says \`new2.txt\`.

EOF

if [ "$B1A_EMPTY" -gt 0 ] || [ "$B1A_STALE" -gt 0 ]; then
cat <<EOF
**Verdict: BUG REPRODUCED.** In \`--color-only\` mode the \`f\` record is
emitted from \`handle_diff_header_minus_line\` (via
\`should_write_generic_diff_header_header_line\`) at the \`---\` line, when
\`plus_file\` still holds the previous file's path (or \`""\` for the first
file). The record disagrees with the row it tags and with the subsequent \`h\`
record for the same file.

EOF
else
cat <<EOF
**Verdict: not reproduced.**

EOF
fi

cat <<EOF
### Scenario B1b: mode-change-only file (\`handle_pending_line_with_diff_name\`)

This is the call site named in the original suspicion: the header text comes
from \`self.diff_line\` while the OSC path comes from \`plus_file\`.

Input (\`mode-change.diff\`): a normal change to \`a.txt\`, then a
mode-change-only section for \`b.txt\` (\`old mode\`/\`new mode\`, no
\`---\`/\`+++\` lines).

Command:

\`\`\`bash
OSC1717=V1 delta --no-gitconfig < mode-change.diff
\`\`\`

(exit code $(cat "$WORK/b1b.rc"); full raw stdout: \`$WORK/b1b.stdout\`)

Output (grep of the relevant row):

\`\`\`
$(grep 'mode +x' "$WORK/b1b.visible" | cut -c1-100)
\`\`\`

EOF

if [ "$B1B_OK" -gt 0 ]; then
cat <<EOF
**Verdict: NOT reproduced — correct behavior.** The \`f\` record names
\`b.txt\`, matching the visible \`b.txt (mode +x)\` header (match count:
$B1B_OK). Reason: \`handle_diff_header_diff_line\` pre-fills
\`minus_file\`/\`plus_file\` from the \`diff --git\` line
(\`src/handlers/diff_header_diff.rs\`, "Pre-fill header fields from the diff
line"), so by the time the pending mode-change header is flushed, \`plus_file\`
already names the right file.

EOF
else
cat <<EOF
**Verdict: BUG REPRODUCED — the \`f\` record on the mode-change header does not
name \`b.txt\`.** See the output above.

EOF
fi

cat <<EOF
### Scenario B1c: \`Only in dir: file\` line (\`handle_additional_cases\`)

Input (\`only-in.diff\`): one \`diff -u\` file section
(\`dir1/f.txt\`→\`dir2/f.txt\`) followed by \`Only in dir1: g.txt\`.

Command:

\`\`\`bash
OSC1717=V1 delta --no-gitconfig < only-in.diff
\`\`\`

(exit code $(cat "$WORK/b1c.rc"); full raw stdout: \`$WORK/b1c.stdout\`)

Output (the relevant row):

\`\`\`
$(grep 'Only in' "$WORK/b1c.visible" | cut -c1-100)
\`\`\`

EOF

if [ "$B1C_STALE" -gt 0 ]; then
cat <<EOF
**Verdict: BUG REPRODUCED.** The \`Only in dir1: g.txt\` row — which is about
\`dir1/g.txt\` — is tagged \`^[]1717;1;f;;;dir2/f.txt^[\\\\\`, the *previous*
file's path, because \`handle_additional_cases\`
(\`src/handlers/mod.rs\`) calls \`diff_header_osc()\` with whatever
\`plus_file\` was left over from the preceding file section.

EOF
else
cat <<EOF
**Verdict: not reproduced.**

EOF
fi

cat <<EOF
### Scenario B1d: \`--color-only\` git rename

Input (\`rename.diff\`): \`diff --git a/old.txt b/new.txt\` + \`rename from\` /
\`rename to\` lines.

Command:

\`\`\`bash
OSC1717=V1 delta --no-gitconfig --color-only < rename.diff
\`\`\`

(exit code $(cat "$WORK/b1d.rc"); full raw stdout: \`$WORK/b1d.stdout\`)

Output:

\`\`\`
$(show b1d 6)
\`\`\`

EOF

if [ "$B1D_EMPTY" -gt 0 ]; then
cat <<EOF
**Verdict: BUG REPRODUCED.** The \`rename from old.txt\` row is tagged
\`^[]1717;1;f;;;^[\\\\\` — empty path. The \`diff --git a/old b/new\` pre-fill
yields \`""\` for a rename (the two paths differ, so
\`get_repeated_file_path_from_diff_line\` returns \`None\`), and the \`f\`
record for the \`rename from\` row is emitted before \`rename to\` sets
\`plus_file\`.

EOF
else
cat <<EOF
**Verdict: not reproduced.**

EOF
fi

cat <<EOF
## Suspected bug 2: side-by-side wrapped continuation rows lack the OSC record

Claim: in \`--side-by-side\` mode, when a long line wraps into multiple visual
rows, the continuation rows are not preceded by the line's OSC record, because
the record is pushed once per \`line_alignment\` entry.

### Scenario B2a: long added line

Input (\`wrap-plus.diff\`): one context line, \`-old\`, and a 200-char \`+xxx…\`
line. Command:

\`\`\`bash
OSC1717=V1 delta --no-gitconfig --side-by-side --width 40 < wrap-plus.diff
\`\`\`

(exit code $(cat "$WORK/b2a.rc"); full raw stdout: \`$WORK/b2a.stdout\`)

Output:

\`\`\`
$(show b2a 13)
\`\`\`

Visual rows containing the wrapped content: $B2A_ROWS; of those, rows missing
the line's \`a;2;;f.txt\` record: $B2A_MISSING.

### Scenario B2b: long context line

\`\`\`bash
OSC1717=V1 delta --no-gitconfig --side-by-side --width 40 < wrap-ctx.diff
\`\`\`

(exit code $(cat "$WORK/b2b.rc"); full raw stdout: \`$WORK/b2b.stdout\`)

\`\`\`
$(show b2b 14)
\`\`\`

Visual rows containing the wrapped content: $B2B_ROWS; rows missing the
\`c;1;;f.txt\` record: $B2B_MISSING.

### Scenario B2c: long deleted line

\`\`\`bash
OSC1717=V1 delta --no-gitconfig --side-by-side --width 40 < wrap-minus.diff
\`\`\`

(exit code $(cat "$WORK/b2c.rc"); full raw stdout: \`$WORK/b2c.stdout\`)

\`\`\`
$(show b2c 13)
\`\`\`

Visual rows containing the wrapped content: $B2C_ROWS; rows missing the
\`d;2;2;f.txt\` record: $B2C_MISSING.

EOF

if [ "$B2A_MISSING" -eq 0 ] && [ "$B2B_MISSING" -eq 0 ] && [ "$B2C_MISSING" -eq 0 ]; then
cat <<EOF
**Verdict: NOT reproduced — correct behavior.** Every visual row of a wrapped
added, context, and deleted line carries the line's OSC record. The claim that
the record is pushed only once per *unwrapped* alignment entry does not hold:
when wrapping is active, \`wrap_minusplus_block\`/\`wrap_zero_block\`
(\`src/wrapping.rs\`) expand the alignment *and* the per-side state vector,
inserting \`HunkMinusWrapped\`/\`HunkPlusWrapped\`/\`HunkZeroWrapped\` entries,
and \`DiffLineMetadata::osc_for_line\` re-emits the primary line's record for
those states without advancing the counters — so the OSC strings vec is also
expanded and every emitted visual row is preceded by a record, exactly as the
code comment says.

Caveat worth knowing (not the suspected bug): on a side-by-side row whose left
panel is empty, no record appears at the very start of the row — the right
panel's record appears mid-row, immediately before the right panel's cell (see
rows of B2a/B2c). A host must therefore scan the whole row for records, not
only the row prefix; and context rows carry the record twice (once per panel).

EOF
else
cat <<EOF
**Verdict: BUG REPRODUCED — some wrapped continuation rows lack the record**
(B2a missing: $B2A_MISSING, B2b missing: $B2B_MISSING, B2c missing:
$B2C_MISSING). See outputs above.

EOF
fi

cat <<EOF
## Sanity check: OSC1717 unset ⇒ stock bytes

\`\`\`bash
delta --no-gitconfig < git-two-files.diff > stock.out
OSC1717=V1 delta --no-gitconfig < git-two-files.diff | sed 's/\x1b]1717[^\x1b]*\x1b\\\\//g' > stripped.out
cmp stock.out stripped.out
\`\`\`

$B0_VERDICT

## Summary

| Claim | Status |
|---|---|
| Bug 1a: \`f\` record on \`---\` rows in \`--color-only\` mode (plain \`diff -u\`) has empty/previous-file path | $( [ "$B1A_EMPTY" -gt 0 ] || [ "$B1A_STALE" -gt 0 ] && echo "REPRODUCED" || echo "not reproduced" ) |
| Bug 1b: \`f\` record wrong at \`handle_pending_line_with_diff_name\` (mode-change-only file) | $( [ "$B1B_OK" -gt 0 ] && echo "NOT reproduced (correct: prefilled from diff line)" || echo "REPRODUCED" ) |
| Bug 1c: \`f\` record on \`Only in\` rows names previous file | $( [ "$B1C_STALE" -gt 0 ] && echo "REPRODUCED" || echo "not reproduced" ) |
| Bug 1d: \`f\` record on \`rename from\` row in \`--color-only\` mode is empty | $( [ "$B1D_EMPTY" -gt 0 ] && echo "REPRODUCED" || echo "not reproduced" ) |
| Bug 2: side-by-side wrapped continuation rows lack the OSC record | $( [ "$B2A_MISSING" -eq 0 ] && [ "$B2B_MISSING" -eq 0 ] && [ "$B2C_MISSING" -eq 0 ] && echo "NOT reproduced (all rows carry the record)" || echo "REPRODUCED" ) |
EOF
```

</details>

dandavison:
Just FYI I'm comparing this branch to some vibecoded `f` and `h` commits that I made and am seeing some incorrect behavior on this branch. My `f` and `h` commits are in branch `osc-1717-metadata-extensions`: https://github.com/dandavison/delta/compare/main...osc-1717-metadata-extensions?body=&expand=1

E.g.

```
\commit bdd5c62c062771ab174c09b0e33098ae33bbc633 (HEAD -> osc-1717-metadata)
```

```
;;;src/delta.rs• src/delta.rs
;;;src/delta.rs──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────────────

;118;;src/delta.rs────────────────────────────────────────────────────────────────────────────────────────┐
;118;;src/delta.rs• 118: pub fn delta<I>(lines: ByteLines<I>, writer: &mut dyn Write, config: &Config) -> │
;118;;src/delta.rs────────────────────────────────────────────────────────────────────────────────────────┘
```

<details>
  <summary>input
  </summary>
  
```diff
commit bdd5c62c062771ab174c09b0e33098ae33bbc633 (HEAD -> osc-1717-metadata)
Author: Stefan Haller <stefan@haller-berlin.de>
Date:   4 weeks ago

    XXXXXXXXX
    
    Emit a version-only metadata handshake as the first output
    
    A host can already learn delta's per-line diff metadata, but only from
    the records emitted before content lines — so a diff with no content (a
    binary file, or the empty diff a host would use to probe) emits nothing,
    and "speaks the protocol" is indistinguishable from "unsupported pager".
    
    Emit a version-only OSC 1717 record (no further fields) once, before
    consuming the diff, whenever a version is negotiated. It's
    content-independent, so a host can probe delta on an empty diff and get
    a conclusive answer.
    
    See the diff-line-metadata OSC spec, §4.4.
    
    Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
    
    Emit `f` and `h` records on file and hunk header rows
    
    Annotate the structural rows too: every row of a file-header block
    carries an `f` record, and every row of a hunk-header block an `h`
    record (spec §5.1/§6.4). This lets a host anchor file and hunk
    navigation on the headers themselves, act on a header row, and see
    files that emit no content records at all (pure renames, mode-only
    changes, binaries).
    
    The two types deliberately differ in payload (spec §5.5): `h` always
    carries the hunk's first new-file line, which is in hand exactly when
    the header is drawn (initialize_hunk has just seeded it), while `f`
    never carries one -- delta draws the file header when it parses the
    `+++` line, before it has seen the first `@@`, so a line number there
    would require buffering the whole header. The emitter does not know
    the path at that point either, so diff_header_osc supplies it, with
    the same plus/minus selection the hunk header uses.
    
    Both headers are multi-row decorations written row by row by the draw
    functions, so write_with_header_osc wraps the writer in a line-aware
    prefixer that repeats the record at the start of every row of the
    block -- the same rule wrapped content rows follow (spec §6.3). The
    grep code shares the hunk-header draw helper and passes no record.
    
    With OSC1717 unset, output remains byte-for-byte identical to stock
    delta.

diff --git a/src/delta.rs b/src/delta.rs
index 5464df9..728489b 100644
--- a/src/delta.rs
+++ b/src/delta.rs
@@ -118,6 +118,12 @@ pub fn delta<I>(lines: ByteLines<I>, writer: &mut dyn Write, config: &Config) ->
 where
     I: BufRead,
 {
+    // Announce protocol support as the first output, so a host can probe delta even on
+    // an empty diff (which produces no per-line records). See features::diff_line_metadata.
+    let handshake = features::diff_line_metadata::handshake();
+    if !handshake.is_empty() {
+        writer.write_all(handshake.as_bytes())?;
+    }
     StateMachine::new(writer, config).consume(lines)
 }
 
diff --git a/src/features/diff_line_metadata.rs b/src/features/diff_line_metadata.rs
index 1ab655b..a10630b 100644
--- a/src/features/diff_line_metadata.rs
+++ b/src/features/diff_line_metadata.rs
@@ -16,6 +16,7 @@
 // Ghostty, VS Code, ConEmu, urxvt); none of them interpret it, so a terminal that
 // is not a participating host skips it harmlessly.
 
+use std::io::{self, Write};
 use std::sync::OnceLock;
 
 use crate::delta::{DiffType, State};
@@ -44,6 +45,19 @@ pub fn negotiated_version() -> Option<u32> {
     *VERSION.get_or_init(|| pick_version(&std::env::var("OSC1717").ok()?))
 }
 
+/// The handshake record: a version-only OSC 1717 (no further fields) that a
+/// conforming pager emits once, as its first output, to announce it speaks the
+/// protocol (see the spec, §4.4). It lets a host probe delta on an empty diff --
+/// which emits no per-line records -- and tell "speaks the protocol" apart from
+/// "unsupported pager". Empty when no host negotiated a version.
+pub fn handshake() -> String {
+    negotiated_version().map_or(String::new(), handshake_for_version)
+}
+
+fn handshake_for_version(version: u32) -> String {
+    format!("{OSC};{version}{ST}")
+}
+
 /// Tracks the current old-/new-file line numbers within a hunk and formats the
 /// per-line OSC sequence. One instance is held by the `Painter` (when emission
 /// is negotiated) and re-seeded at every hunk header.
@@ -129,6 +143,81 @@ pub fn osc_for_line(&mut self, state: &State) -> String {
             file = self.file,
         )
     }
+
+    /// The `h` (hunk-header) record for the current hunk. `initialize_hunk` has
+    /// just seeded `new_line` with the hunk's new-file start -- which is the
+    /// hunk's first line (spec §5.2) -- so it is in hand exactly when the hunk
+    /// header is rendered. `old-line` is empty on a header.
+    pub fn osc_for_hunk_header(&self) -> String {
+        self.header_osc('h', Some(self.new_line), &self.file)
+    }
+
+    /// The `f` (file-header) record. It carries no line numbers (spec §5.5):
+    /// delta renders the file header (the boxed file name) as soon as it parses
+    /// the `+++` line, *before* it has seen the first `@@`, so the file's first
+    /// hunk line is not known here. The path is supplied by the caller because
+    /// the emitter only learns it at the first hunk header.
+    pub fn osc_for_file_header(&self, file: &str) -> String {
+        self.header_osc('f', None, file)
+    }
+
+    fn header_osc(&self, type_char: char, new_line: Option<usize>, file: &str) -> String {
+        let new_field = new_line.map_or(String::new(), |n| n.to_string());
+        format!(
+            "{OSC};{version};{type_char};{new_field};;{file}{ST}",
+            version = self.version,
+        )
+    }
+}
+
+/// A `Write` adapter that injects `prefix` at the start of the stream and after
+/// every newline. delta draws a file/hunk header as a *multi-row* decoration --
+/// an underline under the file name, a box around the hunk header -- written
+/// straight to the output by the `draw` functions. To tag the whole block we
+/// wrap that writer so every row the decoration emits is preceded by the
+/// header's record, the same "every output row carries the record" rule that
+/// wrapped content rows follow (spec §6.3/§6.4).
+struct OscLinePrefixer<'a> {
+    inner: &'a mut dyn Write,
+    prefix: &'a str,
+    at_line_start: bool,
+}
+
+impl Write for OscLinePrefixer<'_> {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        for chunk in buf.split_inclusive(|&b| b == b'\n') {
+            if self.at_line_start {
+                self.inner.write_all(self.prefix.as_bytes())?;
+            }
+            self.inner.write_all(chunk)?;
+            self.at_line_start = chunk.ends_with(b"\n");
+        }
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        self.inner.flush()
+    }
+}
+
+/// Run `draw`, prefixing every output row it writes with `osc` (a header
+/// record). With `osc` empty the closure writes to `writer` untouched, so the
+/// output stays byte-for-byte identical to stock delta.
+pub fn write_with_header_osc(
+    writer: &mut dyn Write,
+    osc: &str,
+    draw: impl FnOnce(&mut dyn Write) -> io::Result<()>,
+) -> io::Result<()> {
+    if osc.is_empty() {
+        draw(writer)
+    } else {
+        let mut prefixer = OscLinePrefixer {
+            inner: writer,
+            prefix: osc,
+            at_line_start: true,
+        };
+        draw(&mut prefixer)
+    }
 }
 
 #[cfg(test)]
@@ -149,6 +238,60 @@ fn emitter() -> DiffLineMetadata {
         md
     }
 
+    #[test]
+    fn test_handshake_is_version_only() {
+        // The handshake carries only the version (no further fields), so a host tells
+        // it apart from a per-line record by field count.
+        assert_eq!(handshake_for_version(1), "\x1b]1717;1\x1b\\");
+    }
+
+    #[test]
+    fn test_hunk_header_record_carries_the_hunks_first_line() {
+        // `h` carries the hunk's first new-file line (just seeded by
+        // initialize_hunk) and no old-line. Emitting it does not advance the
+        // counters: the hunk's first content line reports the same line.
+        let mut md = emitter();
+        assert_eq!(md.osc_for_hunk_header(), "\x1b]1717;1;h;1;;f.txt\x1b\\");
+        assert_eq!(
+            md.osc_for_line(&State::HunkZero(DiffType::Unified, None)),
+            "\x1b]1717;1;c;1;;f.txt\x1b\\"
+        );
+    }
+
+    #[test]
+    fn test_file_header_record_has_no_line_numbers() {
+        // `f` never carries line numbers (spec §5.5): delta draws the file
+        // header before it has seen the first `@@`. The path comes from the
+        // caller, because the emitter's own state is only seeded at the first
+        // hunk header.
+        let md = emitter();
+        assert_eq!(
+            md.osc_for_file_header("src/foo.rs"),
+            "\x1b]1717;1;f;;;src/foo.rs\x1b\\"
+        );
+    }
+
+    #[test]
+    fn test_header_record_prefixes_every_row_of_a_multi_row_block() {
+        // A file/hunk header is a multi-row decoration (an underlined name, a
+        // box); every row of the block carries the record (spec §6.4). With no
+        // record the drawn bytes pass through untouched.
+        let draw = |w: &mut dyn io::Write| -> io::Result<()> {
+            write!(w, "──┐\n1:│\n──┘\n")
+        };
+
+        let mut tagged = Vec::new();
+        write_with_header_osc(&mut tagged, "[osc]", draw).unwrap();
+        assert_eq!(
+            String::from_utf8(tagged).unwrap(),
+            "[osc]──┐\n[osc]1:│\n[osc]──┘\n"
+        );
+
+        let mut untouched = Vec::new();
+        write_with_header_osc(&mut untouched, "", draw).unwrap();
+        assert_eq!(String::from_utf8(untouched).unwrap(), "──┐\n1:│\n──┘\n");
+    }
+
     #[test]
     fn test_wrapped_rows_reemit_the_primary_record_without_advancing() {
         let mut md = emitter();
diff --git a/src/handlers/diff_header.rs b/src/handlers/diff_header.rs
index b5fe114..168b83c 100644
--- a/src/handlers/diff_header.rs
+++ b/src/handlers/diff_header.rs
@@ -57,12 +57,14 @@ fn should_write_generic_diff_header_header_line(&mut self) -> std::io::Result<bo
         // (it connects the plus_file and minus_file),
         // and to call fn handle_generic_diff_header_header_line directly.
         if self.config.color_only {
+            let osc = self.diff_header_osc();
             write_generic_diff_header_header_line(
                 &self.line,
                 &self.raw_line,
                 &mut self.painter,
                 &mut self.mode_info,
                 self.config,
+                &osc,
             )?;
             Ok(true)
         } else {
@@ -189,6 +191,25 @@ pub fn handle_diff_header_file_operation_line(&mut self) -> std::io::Result<bool
         Ok(handled_line)
     }
 
+    /// The `f` (file-header) record for the current file, or an empty string
+    /// when no host negotiated emission. The path is selected exactly as the
+    /// hunk header selects it (`plus_file`, or `minus_file` for a deletion).
+    /// delta renders the file header before it has parsed the first `@@`, so
+    /// the record's new-line is left empty (`osc_for_file_header`).
+    pub(crate) fn diff_header_osc(&self) -> String {
+        match self.painter.diff_line_metadata.as_ref() {
+            Some(md) => {
+                let file = if self.plus_file == "/dev/null" {
+                    &self.minus_file
+                } else {
+                    &self.plus_file
+                };
+                md.osc_for_file_header(file)
+            }
+            None => String::new(),
+        }
+    }
+
     /// Construct file change line from minus and plus file and write with DiffHeader styling.
     fn _handle_diff_header_header_line(&mut self, comparing: bool) -> std::io::Result<()> {
         let line = get_file_change_description_from_file_paths(
@@ -200,12 +221,14 @@ fn _handle_diff_header_header_line(&mut self, comparing: bool) -> std::io::Resul
             self.config,
         );
         // FIXME: no support for 'raw'
+        let osc = self.diff_header_osc();
         write_generic_diff_header_header_line(
             &line,
             &line,
             &mut self.painter,
             &mut self.mode_info,
             self.config,
+            &osc,
         )
     }
 
@@ -242,12 +265,14 @@ pub fn handle_pending_line_with_diff_name(&mut self) -> std::io::Result<()> {
             let label = format_label(&self.config.file_modified_label);
             let name = get_repeated_file_path_from_diff_line(&self.diff_line).unwrap_or_default();
             let line = format!("{}{}", label, format_file(&name));
+            let osc = self.diff_header_osc();
             write_generic_diff_header_header_line(
                 &line,
                 &line,
                 &mut self.painter,
                 &mut self.mode_info,
                 self.config,
+                &osc,
             )
         } else if !self.config.color_only
             && self.should_handle()
@@ -270,6 +295,7 @@ pub fn write_generic_diff_header_header_line(
     painter: &mut Painter,
     mode_info: &mut String,
     config: &Config,
+    metadata_osc: &str,
 ) -> std::io::Result<()> {
     // If file_style is "omit", we'll skip the process and print nothing.
     // However in the case of color_only mode,
@@ -283,14 +309,20 @@ pub fn write_generic_diff_header_header_line(
         // Maintain 1-1 correspondence between input and output lines.
         writeln!(painter.writer)?;
     }
-    draw_fn(
+    crate::features::diff_line_metadata::write_with_header_osc(
         painter.writer,
-        &format!("{}{}", line, if pad { " " } else { "" }),
-        &format!("{}{}", raw_line, if pad { " " } else { "" }),
-        mode_info,
-        &config.decorations_width,
-        config.file_style,
-        decoration_ansi_term_style,
+        metadata_osc,
+        |w| {
+            draw_fn(
+                w,
+                &format!("{}{}", line, if pad { " " } else { "" }),
+                &format!("{}{}", raw_line, if pad { " " } else { "" }),
+                mode_info,
+                &config.decorations_width,
+                config.file_style,
+                decoration_ansi_term_style,
+            )
+        },
     )?;
     if !mode_info.is_empty() {
         mode_info.truncate(0);
diff --git a/src/handlers/grep.rs b/src/handlers/grep.rs
index 696c3cc..b119ea5 100644
--- a/src/handlers/grep.rs
+++ b/src/handlers/grep.rs
@@ -177,6 +177,7 @@ fn emit_ripgrep_format_grep_line(
                 &HunkHeaderIncludeCodeFragment::YesNoSpace,
                 "",
                 self.config,
+                "",
             )?
         }
         if new_section {
@@ -240,6 +241,7 @@ fn emit_ripgrep_format_grep_line(
             &HunkHeaderIncludeCodeFragment::YesNoSpace,
             grep_line.line_type.file_path_separator(),
             self.config,
+            "",
         )
     }
 
@@ -266,6 +268,7 @@ fn emit_classic_format_grep_line(&mut self, grep_line: GrepLine) -> std::io::Res
                     &HunkHeaderIncludeCodeFragment::Yes,
                     grep_line.line_type.file_path_separator(),
                     self.config,
+                    "",
                 )?
             }
             _ => {
diff --git a/src/handlers/hunk_header.rs b/src/handlers/hunk_header.rs
index 0679b64..f7789ea 100644
--- a/src/handlers/hunk_header.rs
+++ b/src/handlers/hunk_header.rs
@@ -186,9 +186,19 @@ pub fn emit_hunk_header_line(
                 .initialize_hunk(line_numbers_and_hunk_lengths, file);
         }
 
+        // The `h` record for every row of the hunk-header decoration. Empty
+        // when no host negotiated emission, in which case the decoration is
+        // drawn unchanged.
+        let h_osc = self
+            .painter
+            .diff_line_metadata
+            .as_ref()
+            .map_or(String::new(), |md| md.osc_for_hunk_header());
+
         if self.config.hunk_header_style.is_raw {
-            write_hunk_header_raw(&mut self.painter, line, raw_line, self.config)?;
+            write_hunk_header_raw(&mut self.painter, line, raw_line, &h_osc, self.config)?;
         } else if self.config.hunk_header_style.is_omitted {
+            write!(self.painter.writer, "{h_osc}")?;
             writeln!(self.painter.writer)?;
         } else {
             // Add a blank line below the hunk-header-line for readability, unless
@@ -217,6 +227,7 @@ pub fn emit_hunk_header_line(
                 &self.config.hunk_header_style_include_code_fragment,
                 ":",
                 self.config,
+                &h_osc,
             )?;
         };
         self.painter.set_highlighter();
@@ -282,6 +293,7 @@ fn write_hunk_header_raw(
     painter: &mut Painter,
     line: &str,
     raw_line: &str,
+    metadata_osc: &str,
     config: &Config,
 ) -> std::io::Result<()> {
     let (mut draw_fn, pad, decoration_ansi_term_style) =
@@ -289,14 +301,20 @@ fn write_hunk_header_raw(
     if config.hunk_header_style.decoration_style != DecorationStyle::NoDecoration {
         writeln!(painter.writer)?;
     }
-    draw_fn(
+    crate::features::diff_line_metadata::write_with_header_osc(
         painter.writer,
-        &format!("{}{}", line, if pad { " " } else { "" }),
-        &format!("{}{}", raw_line, if pad { " " } else { "" }),
-        "",
-        &config.decorations_width,
-        config.hunk_header_style,
-        decoration_ansi_term_style,
+        metadata_osc,
+        |w| {
+            draw_fn(
+                w,
+                &format!("{}{}", line, if pad { " " } else { "" }),
+                &format!("{}{}", raw_line, if pad { " " } else { "" }),
+                "",
+                &config.decorations_width,
+                config.hunk_header_style,
+                decoration_ansi_term_style,
+            )
+        },
     )?;
     Ok(())
 }
@@ -318,6 +336,7 @@ pub fn write_line_of_code_with_optional_path_and_line_number(
     include_code_fragment: &HunkHeaderIncludeCodeFragment,
     file_path_separator: &str,
     config: &Config,
+    metadata_osc: &str,
 ) -> std::io::Result<()> {
     let (mut draw_fn, _, decoration_ansi_term_style) = draw::get_draw_function(decoration_style);
     let line = match (config.color_only, include_code_fragment) {
@@ -353,14 +372,21 @@ pub fn write_line_of_code_with_optional_path_and_line_number(
             painter,
             config,
         );
-        draw_fn(
+        let output_buffer = &painter.output_buffer;
+        crate::features::diff_line_metadata::write_with_header_osc(
             painter.writer,
-            &painter.output_buffer,
-            &painter.output_buffer,
-            "",
-            &config.decorations_width,
-            config.null_style,
-            decoration_ansi_term_style,
+            metadata_osc,
+            |w| {
+                draw_fn(
+                    w,
+                    output_buffer,
+                    output_buffer,
+                    "",
+                    &config.decorations_width,
+                    config.null_style,
+                    decoration_ansi_term_style,
+                )
+            },
         )?;
         painter.output_buffer.clear();
     }
diff --git a/src/handlers/mod.rs b/src/handlers/mod.rs
index 40bdf68..b8cffd6 100644
--- a/src/handlers/mod.rs
+++ b/src/handlers/mod.rs
@@ -40,12 +40,14 @@ pub fn handle_additional_cases(&mut self, to_state: State) -> std::io::Result<bo
         self.state = to_state;
         if self.should_handle() {
             self.painter.emit()?;
+            let osc = self.diff_header_osc();
             diff_header::write_generic_diff_header_header_line(
                 &self.line,
                 &self.raw_line,
                 &mut self.painter,
                 &mut self.mode_info,
                 self.config,
+                &osc,
             )?;
             handled_line = true;
         }

```  
  
</details>

stefanhaller:
Sorry for the delay in reacting to this. I handed Kimi's review feedback to claude, and it came up with a few fixes that are commited as fixups here. It also made an addition to the draft spec to add a "Common pitfalls" section mentioning these; the difftastic and diff-so-fancy PRs had similar issues, and I pushed fixup commits there, too.

Could you check if this also addresses your second comment? To be honest I'm not really sure what that one was about exactly.

<details>
  <summary>
Claude's response, for reference.
  </summary>

The report was valid — I confirmed and fixed all three reproduced bugs, plus one more of the same class that the report missed.

## What the report got right

Kimi's analysis was accurate on all five counts: bugs 1a, 1c, and 1d reproduced exactly as described, and 1b and 2 are indeed not bugs. The shared root cause is that `diff_header_osc()` reads `self.plus_file`/`self.minus_file` at rows where those fields don't yet (or no longer) describe the row being rendered.

## Fixes

1. **Empty path → no record** ([diff_header.rs:199-233](src/handlers/diff_header.rs#L199-L233)): `diff_header_osc()` now emits nothing when the file path is unknown. This fixes the `rename from` row in `--color-only` (bug 1d) and the first `---` row of plain `diff -u` input (half of bug 1a). Rationale per the spec: tagging every header row is a recommendation, not a requirement, while a wrong or empty path hands the host a phantom file identity — so an unknown-path row is better left untagged (the `+++`/`rename to` row right below carries the correct `f`).

2. **Stale path cleared at section start** ([diff_header.rs:96-105](src/handlers/diff_header.rs#L96-L105)): plain `diff -u` has no `diff --git` line to reset the fields, so at a `---` row `plus_file` still held the *previous* file's path. It's now cleared there (the other half of bug 1a). The full snapshot suite confirms this changes no rendered output.

3. **"Only in" / "Submodule" rows carry no record** ([mod.rs:21-28](src/handlers/mod.rs#L21-L28), [diff_header_misc.rs:49-63](src/handlers/diff_header_misc.rs#L49-L63), [submodule.rs:16-18](src/handlers/submodule.rs#L16-L18)): `handle_additional_cases` now takes the record from its call sites, because only they know whether the row is about the current file. "Only in" rows (bug 1c) and Submodule rows (same latent bug, untested by the report) get none; "Binary files" rows in a git diff keep their correct record.

4. **Bonus find**: in default mode, a binary file's record was `f;;;img.png (binary file)` — the display suffix is appended to `plus_file` in place, so it leaked into the record. The suffix is now a shared constant and gets stripped in `diff_header_osc()`.

## Verification

- Kimi's exact `repro.sh` now reports **all five claims "not reproduced"**, and its sanity check still passes (stripping records yields bytes identical to stock delta).
- Full suite: 454 passed, 0 failed; clippy clean for the changed code.
- Added 5 snapshot regression tests in [diff_header.rs](src/handlers/diff_header.rs#L955), with a small harness (`run_delta_with_diff_line_metadata` + `DiffLineMetadata::v1_for_tests`) that injects the emitter directly, since the env-var handshake is cached in a process-wide `OnceLock` and can't be toggled per test.

One thing worth knowing but not touched: the odd row ordering visible in `--color-only` with multi-file `diff -u` input (file 1's content rows flushed after file 2's header rows) is a pre-existing delta bug — released delta does the same with no OSC involved — so I left it alone.
</details>

ccoVeille:
Hi @stefanhaller 

This feedback is valuable

Did you consider creating a repository about osc-1717 and provide a clear readme

If osc-1717 starts to be adopted widely, it could be interesting to gather everything in one place instead of posting in all repositories: lazydiff, delta, diff-so-fancy, diffstatic ...

Thanks for your work on this 🎉

stefanhaller:
> Did you consider creating a repository about osc-1717 and provide a clear readme

That is an excellent idea; I created the repo and opened https://github.com/stefanhaller/diff-line-metadata-spec/pull/1 for discussion of the v1 draft. I'll close https://github.com/jesseduffield/lazygit/pull/5731.

stefanhaller:
I set this to draft for the time being. Like in the diff-so-fancy PR, we should only merge this after testing it against the actual, non-prototype lazygit host implementation, which I'm currently in the middle of building. I hope to have something for testing by the end of next weekend.

stefanhaller:
@dandavison I added one more fixup (4b0ba542f2b7) that fixes the behavior for submodules, which didn't work quite right.

Did you get a chance to review the fixes from [this comment above](https://github.com/dandavison/delta/pull/2181#issuecomment-5231001942)? For reference, they are eeed4c6de844 (there's a hunk about submodules in this one which is changed again by the submodule fix) and 51b4c757cc28.

Let me know when you'd like me to auto-squash the fixups; or feel free to do that yourself and force-push.

I'm making very good progress on the lazygit side of this; I'll post a series of PRs very soon, hope to merge them to master some time next week, and release the whole thing early October.

dandavison:
@stefanhaller thanks, please feel free to squash your fixups. For the delta gitu TUI I'm using the latest tip of this branch plus one commit, and will let you know if I encounter any problems.

This is the one commit I have added: https://github.com/dandavison/delta/commit/48c3b84  With that commit, delta emits a `C` record on rows in `git log` output, containing the commit hash. This allows gitu to add behavior (keybindings) to a git log view that has been decorated by delta.

stefanhaller:
Interesting, I wasn't aware that delta decorates log output too. I have to say that I'm reluctant to add this to the spec; the spec is about diff output only, and I'd like to keep it focused on that.

Your experiment shows how future extensions can affect current functionality though; this actually broke a feature in lazygit. Not serious, easy to work around, I just found this interesting: lazygit shows diffs of a commit using `--stat`, so there's a "table of content" below the commit message, and I recently added the feature to turn the filenames in the diff stat into clickable links that scroll to the beginning of the clicked file. Lazygit actually parses the diff stat to do that, and it stops parsing it as soon as it sees the first OSC 1717 record (besides the initial handshake), on the assumption that metadata records are only emitted after the message and the diff stat. I can easily fix that by only stopping to parse when I see one of the known records, of course.

But I also wonder: does delta actually alter the output of a git log, similar to how it alters a diff? Because if not, I'd question the usefulness of the `C` record; it seems easy enough for the host to search for commit lines itself, and that way it works with other diff renderers too.

stefanhaller:
@dandavison Do you have any more thoughts on the above, specifically on the usefulness of a `C` record? I find the topic of ad-hoc protocol extensions pretty important, and would like some more clarity on it before I finalize the spec.

PR #2193, opened 2026-07-16
fix: bump plist to 1.10.0, pulling patched quick-xml (RUSTSEC-2026-0194, -0195)

### LINKED ISSUE #2183: RUSTSEC-2026-0194: Quadratic run time when checking a start tag for duplicate attribute names
> Quadratic run time when checking a start tag for duplicate attribute names

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `quick-xml`                      |
| Version             | `0.39.4`                   |
| URL                 | [https://github.com/tafia/quick-xml/issues/969](https://github.com/tafia/quick-xml/issues/969) |
| Date                | 2026-06-29                         |
| Patched versions    | `>=0.41.0`                  |

`BytesStart::attributes()` returns an `Attributes` iterator which, by default
(`with_checks(true)`), rejects a start tag that repeats an attribute name. For
each attribute yielded, the iterator compared the new name against every name
seen so far in the same tag using a linear scan, so a start tag with `N`
distinct attribute names cost `O(N²)` byte comparisons. There was no bound on
`N` other than the size of the buffered start tag.

## Impact

Any code that parses untrusted XML and iterates a start tag&#39;s attributes with
the default duplicate check enabled can be made to spend CPU time quadratic in
the number of attributes on a single tag. Because the check is pure computation
with no `.await`/I/O, an I/O-based timeout on the consumer (for example a read
or request timeout) cannot interrupt it while it runs.

Measured cost of a single start tag, release build:

| Attributes on one tag | Time |
|---|---|
| 80,000  | ~6 s   |
| 800,000 | ~10 min |

The cost grows with the square of the attribute count, so a start tag of a few
tens of megabytes can stall a parsing thread for hours. No memory is exhausted
and the parser does not crash; the effect is CPU exhaustion on the thread doing
the parsing: a single crafted start tag can pin a CPU core for minutes to hours,
denying service to that worker. A deployment that places a wall-clock bound on
parsing, or confines it to a non-critical thread, may consider the availability
impact lower.

## Affected code paths

* `BytesStart::attributes()` / `Attributes` iterated with checks enabled (the
  default), and `BytesStart::try_get_attribute`.
* `NsReader`, which resolves namespaces by iterating a tag&#39;s attributes and so
  reaches the same check internally.

Consumers that iterate attributes with `.attributes().with_checks(false)` and do
not use `NsReader` are not affected.

This was reported as reachable by a remote, unauthenticated attacker in a
real-world RPKI relying party (NLnet Labs Routinator) via a crafted RRDP
`snapshot.xml`.

## Remediation

Upgrade to `quick-xml &gt;= 0.41.0`, where the duplicate check keeps the linear
scan for start tags with a small number of attributes and switches to an `O(1)`
hash pre-filter above a threshold, making the whole tag `O(N)`. The reported
`AttrError::Duplicated` positions are unchanged.

If upgrading is not possible and duplicate-name detection is not required,
disable it with `.attributes().with_checks(false)` (this does not help
`NsReader` consumers, which have no equivalent opt-out before 0.41.0).

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2026-0194.html) for additional details.

### LINKED ISSUE #2184: RUSTSEC-2026-0195: Unbounded namespace-declaration allocation in `NsReader` enables memory-exhaustion denial of service
> Unbounded namespace-declaration allocation in `NsReader` enables memory-exhaustion denial of service

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `quick-xml`                      |
| Version             | `0.39.4`                   |
| URL                 | [https://github.com/tafia/quick-xml/issues/970](https://github.com/tafia/quick-xml/issues/970) |
| Date                | 2026-06-29                         |
| Patched versions    | `>=0.41.0`                  |

`NsReader` resolves namespaces by calling `NamespaceResolver::push` for every
`Start`/`Empty` event *before* the event is returned to the caller. `push`
iterated all `xmlns` / `xmlns:*` attributes on the start tag and, for each one,
appended the prefix bytes to an internal buffer and pushed a `NamespaceBinding`
(32 bytes on 64-bit) to an internal `Vec`, with no upper bound on the number of
declarations.

## Impact

A start tag with `N` namespace declarations drove roughly `3×` the tag&#39;s byte
size in `NamespaceResolver` heap, allocated *inside* `quick-xml` before the
`NsReader` consumer ever received the event and could inspect or reject it. A
consumer that bounds its *input* size therefore still cannot bound this
allocation: an `M`-byte start tag yields on the order of `3 × M` bytes of
resolver heap the caller never sees.

On untrusted XML this lets a remote, unauthenticated attacker force large heap
allocations with a single start tag. With several `NsReader`s running
concurrently on independent inputs (a common server pattern), the allocations
stack and can exhaust process memory, causing the operating system to kill the
process (OOM). This was confirmed against a real-world RPKI relying party (NLnet
Labs Routinator), where concurrent RRDP validation workers parsing a crafted
`snapshot.xml` exceeded the memory limit and the process was OOM-killed.

## Affected code paths

Consumers using `NsReader` (which always calls `NamespaceResolver::push` before
yielding `Start`/`Empty`), or calling `NamespaceResolver::push` directly. A plain
`Reader` that does not perform namespace resolution is not affected.

## Remediation

Upgrade to `quick-xml &gt;= 0.41.0`. `NamespaceResolver::push` now rejects a start
tag that declares more than `DEFAULT_MAX_DECLARATIONS_PER_ELEMENT` (256)
namespace bindings, returning the new `NamespaceError::TooManyDeclarations`
instead of allocating without limit. The limit is configurable via
`NamespaceResolver::set_max_declarations_per_element` (use `usize::MAX` to
restore the previous unbounded behavior), and `NsReader::resolver_mut()` is
provided to reach it.

There is no clean workaround for `NsReader` consumers before 0.41.0, as the
allocation happens inside the reader with no configuration knob to cap it.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2026-0195.html) for additional details.

albatrossflyon-coder:
Fixes #2183
Fixes #2184

`quick-xml` was locked at 0.32.0, pulled in transitively via `syntect` (delta's syntax-highlighting dependency, used to parse `.plist`-format theme files) → `plist` 1.7.0 → `quick-xml` 0.32.0.

Two RUSTSEC advisories affect `quick-xml` before 0.41.0:
- **RUSTSEC-2026-0194** — quadratic run time when checking a start tag for duplicate attribute names (CPU-exhaustion DoS)
- **RUSTSEC-2026-0195** — unbounded namespace-declaration allocation in `NsReader` (memory-exhaustion DoS)

Since `syntect` uses `plist` to parse theme files, this is a real attack surface if delta is ever pointed at an untrusted theme file — not just advisory-scanner noise.

`plist` 1.7.0's own dependency constraint caps `quick-xml` at 0.32.0, so a plain `cargo update -p quick-xml` can't reach the patched version. Bumping `plist` to 1.10.0 pulls in the patched `quick-xml` 0.41.0.

No source changes needed — `Cargo.lock` only. Verified:
- `cargo check --workspace` — clean
- `cargo test --workspace` — 414 passed, 8 ignored

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -474,12 +474,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
-version = "0.3.11"
+version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
-dependencies = [
- "powerfmt",
-]
+checksum = "7cd812cc2bc1d69d4764bd80df88b4317eaef9e773c75226407d9bc0876b211c"
 
 [[package]]
 name = "diff"
@@ -1073,9 +1070,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "521739c6d2bac4aa25192232afe6841231376b2b26d4d9fae5ecf8ca5772e441"
 
 [[package]]
 name = "num-traits"
@@ -1258,9 +1255,9 @@ checksum = "69d3587f8a9e599cc7ec2c00e331f71c4e69a5f9a4b8a6efd5b07466b9736f9a"
 
 [[package]]
 name = "plist"
-version = "1.7.0"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "42cf17e9a1800f5f396bc67d193dc9411b59012a5876445ef450d449881e1016"
+checksum = "7da1d65da6dd5d1e44199ac0f58712d241c0f439f80adea8924d832384087f85"
 dependencies = [
  "base64",
  "indexmap",
@@ -1315,9 +1312,9 @@ dependencies = [
 
 [[package]]
 name = "quick-xml"
-version = "0.32.0"
+version = "0.41.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1d3a6e5838b60e0e8fa7a43f22ade549a37d61f8bdbe636d0d7816191de969c2"
+checksum = "e660451e55124f798a69a5af3f49ccfbefbd41910eefd25caf2393e1f3473ec1"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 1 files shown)


albatrossflyon-coder:
Hi — just checking in on this one, it's been open about 7 days with no review yet. Happy to make any changes if there's feedback, or let me know if anything's blocking it. Thanks for maintaining delta!

PR #2194, opened 2026-07-16
Don't misparse git grep -h output as path=code

### LINKED ISSUE #2192: 🐛 `git grep -h` drops content starting with filename + "="
With `git grep -h`, delta drops the beginning of matching lines if they begin with the filename, followed by an equal sign. To create a minimal repo for reproducing the behavior:

```bash
cd $(mktemp -d)
git init
echo foo=equals >foo
git add foo
git commit -m 'Initial commit'
```

Then to reproduce:

```console
$ git grep -h . -- ./foo
equals
$ git --no-pager grep -h . -- ./foo
foo=equals
```

albatrossflyon-coder:
Fixes #2192

## Root cause

Delta's `parse_grep_line` tries several heuristic regexes to infer a `path` + separator + `code` structure from raw grep output, since the format is inherently ambiguous (this is documented directly in the code: "The format is ambiguous, but we attempt to parse it.").

With `git grep -h` (or `--no-filename`), git suppresses the filename entirely — every output line is just raw file content, with no path/separator structure at all. So a content line that happens to contain a run like `word=word` gets misdetected as a `path=code` grep line purely by coincidence.

Minimal repro (from the issue):
```console
$ echo foo=equals >foo && git add foo && git commit -m x
$ git grep -h . -- ./foo
equals          # delta strips the leading "foo=" -- wrong
$ git --no-pager grep -h . -- ./foo
foo=equals       # correct raw output
```

## Fix

Detect `-h`/`--no-filename` via the existing `CallingProcess::GitGrep` short/long-option tracking — the same pattern already used elsewhere in this file for `-W`/`--function-context` and `-p`/`--show-function`. When either is present, skip heuristic parsing entirely and let the line pass through unmodified, since there's no reliable filename signal to parse against in the first place.

## Testing

Added a regression test (`test_parse_grep_h_no_filename_not_misparsed_as_path`) using the exact repro from the issue. Full workspace test suite passes: 415 passed, 8 ignored (up from 414 — the new test).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -474,12 +474,9 @@ dependencies = [
 
 [[package]]
 name = "deranged"
-version = "0.3.11"
+version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b42b6fa04a440b495c8b04d0e71b707c585f83cb9cb28cf8cd0d976c315e31b4"
-dependencies = [
- "powerfmt",
-]
+checksum = "7cd812cc2bc1d69d4764bd80df88b4317eaef9e773c75226407d9bc0876b211c"
 
 [[package]]
 name = "diff"
@@ -1073,9 +1070,9 @@ dependencies = [
 
 [[package]]
 name = "num-conv"
-version = "0.1.0"
+version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51d515d32fb182ee37cda2ccdcb92950d6a3c2893aa280e540671c2cd0f3b1d9"
+checksum = "521739c6d2bac4aa25192232afe6841231376b2b26d4d9fae5ecf8ca5772e441"
 
 [[package]]
 name = "num-traits"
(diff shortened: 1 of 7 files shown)


albatrossflyon-coder:
Hi — just checking in on this one, it's been open about 7 days with no review yet. Happy to make any changes if there's feedback, or let me know if anything's blocking it. Thanks for maintaining delta!

dandavison:
Hi @albatrossflyon-coder, thanks for this. I've been struggling to get to PRs promptly recently due to my day job. But I will try -- I can see this is carefully done and looks simple and like a good improvement!

dandavison:
Are the changes to Cargo.lock intentional? Can we do them in a separate PR if so?

albatrossflyon-coder:
Good catch, and thanks for reviewing! Not intentional — turned out to be a side effect of running `cargo build`/`cargo test` locally, which silently pulled in newer transitive dependency versions (deranged, time/time-core/time-macros, plist) into the lockfile. The grep.rs fix itself doesn't touch dependencies at all. One of those (plist 1.7.0 -> 1.10.0) is actually already its own dedicated fix in #2193, so no need to duplicate it here.

Just pushed a commit reverting Cargo.lock to match main — this PR now only touches src/handlers/grep.rs. Thanks for taking the time to look, especially with everything else going on!

albatrossflyon-coder:
Fixed the Clippy failure in `65c2bc9`. Turned out unrelated to this PR's actual change (`src/handlers/grep.rs`) — 5 lints newly firing because CI runs on latest stable clippy (1.97) and `main`'s last green run predates these becoming stricter defaults. All 5 are clippy's own suggested one-line fixes (drop a redundant `.into_iter()`, `truncate(0)` → `clear()`, 3x drop a redundant `&` in format/writeln args), no behavior change. `cargo clippy -- -D warnings` (your CI's exact command) is clean now, all 423 tests still pass.

dandavison:
Merged, thanks a lot!

PR #2209, opened 2026-08-07
Indicate dirty state in short-format submodule diffs

### LINKED ISSUE #1838: 🚀 Indicate submodule dirty state
It might be nice to indicate that submodules are "dirty" rather than showing red and green with the same commit hash.

image

MsfPablo:
## Summary

When `diff.submodule = short` (the default), a dirty submodule shows:

```
-Subproject commit ca030fd1a02225a6fc1a834c480276d9c97a8c6f
+Subproject commit 803be42ca46af0fbc65b54a9abfb499389516939-dirty
```

`get_submodule_short_commit` matched the optional `-dirty` suffix but only
ever returned the hash capture group, discarding the flag. So delta
rendered a dirty submodule identically to a clean one — just
`<red hash>..<green hash>` — which is exactly the ambiguity described in #1838.

This propagates the dirty flag through and appends `-dirty` to the
printed commit range when the working-tree side is dirty:

```
ca030fd1a022..803be42ca46a-dirty
```

## Test plan

- [x] `cargo test` — 438 passed, 0 failed
- [x] Updated `test_simple_dirty_submodule_diff` to assert the `-dirty` suffix is rendered
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --all-targets` — no new warnings introduced (pre-existing warnings are in unrelated `utils/process.rs`)

Fixes #1838


MsfPablo:
Following up on this one. It surfaces the dirty marker in short-format submodule diffs, matching what git itself shows. Glad to add a test case if that'd help review.

PR #2240, opened 2026-09-19
Take the terminal width from $COLUMNS when it is set

### LINKED ISSUE #2239: Use the COLUMNS env var to learn the terminal width
Currently, delta queries only the terminal using the TIOCGWINSZ ioctl to know the window width, and falls back to a default of 80 when output is not a tty. This is a problem for hosts running delta in a pipe in order to render its output in a view; they have to jump through hoops to let users include a `--width` argument in the delta command.

Lazygit is such a host: it used to run the git command with delta as a GIT_PAGER in a pty, but we recently had to change this on Windows because our pty implementation there won't work well with the upcoming OSC 1717 feature. We now run git and delta in a pipe without a pty, and setting COLUMNS would allow us to not make this a breaking change for users.

Here's a PR that implements this: #2240. Disclaimer: it was created by Claude, but I reviewed it thoroughly and made the important decisions together with Claude; in particular the one that COLUMNS takes precedence over ioctl. I think this makes sense because git does the same, and because it gives hosts a way to override the terminal width should they have a reason to do that (lazygit doesn't).

stefanhaller:
Delta learns how wide to render by asking the terminal on its stdout. A caller that runs delta with no terminal there cannot state a width, so delta falls back to 80 columns. Editors and TUIs that render delta's output into one of their own views hit this, and have to pass --width on every invocation.

Read $COLUMNS first, and ask the terminal only if $COLUMNS is unset or does not hold a positive number. Git resolves the width the same way in term_columns(), and sets $COLUMNS for the pager it spawns, because its own stdout is a pipe to that pager by then and the ioctl no longer reaches a terminal. Following git keeps delta and the git process feeding it on the same width. An explicit --width still wins over both.

Fixes #2239.


dandavison:
Thanks @stefanhaller, merged. I was also experimenting with TUI integration and have confirmed that this avoids passing `--width`.

dandavison:
If you have an approximate sense of when you'll need this released by please just let me know and I'll make sure that happens.

